PR #1, opened 2014-08-11
8.0 add website doc jsonrpc

OSguard:
add module to display models and fields inside website

### DIFF
--- a/website_doc_jsonrpc/__openerp__.py
+++ b/website_doc_jsonrpc/__openerp__.py
@@ -1,0 +1,52 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#    OpenERP, Open Source Management Solution
+#    Copyright (C) 2010-2013 OpenERP s.a. (<http://openerp.com>).
+#    Copyright (C) 2014 initOS GmbH & Co. KG (<http://www.initos.com>).
+#    Author Markus Schneider <markus.schneider at initos.com>
+#
+#    This program is free software: you can redistribute it and/or modify
+#    it under the terms of the GNU Affero General Public License as
+#    published by the Free Software Foundation, either version 3 of the
+#    License, or (at your option) any later version.
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'Website Odoo JSON RPC doc',
+    'version': '0.1',
+    'author': 'initOS GmbH & Co. KG',
+    'category': '',
+    'description': """Generates a documentation for JSON RPC
+    for the current installation on the fly""",
+    'website': 'http://www.initos.com',
+    'license': 'AGPL-3',
+    'images': [
+    ],
+    'depends': [
+        'website'
+    ],
+    'data': [
+        'views/model_view.xml'
+    ],
+    'js': [
+    ],
+    'qweb': [
+    ],
+    'css': [
+    ],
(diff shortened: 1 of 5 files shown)


OSguard:
On 10.08.2014 19:49, Nhomar Hernandez wrote (on mailing list):

> And the name of your module say "jsonrpc" but it actually returns "http"
> (obviously using the json cookie management api on controller).
> 
> IMHO It is a good feature but a l think you open some backdoors that can
> not be expose "public".
> 
> If you fix:
> 
> 1.- Not depend of website instead of web.
> 2.- auth="user" and not "public".

OSguard:
We can discuss the name, the idea was to have a documentation what you can have access via jsonrpc and the models are only one endpoint. It depends if it should be list of all JSON data or is just a module to explore models and fields.

I see no diffrents between user and public. Your source is also public available (as it is AGPL) so you don't share any new information. It would be better if you only see objects and fields you have access depends on your current user. But this is not included in the first version.

max3903:
Can you rebase so we can test on runbot ?

rafaelbn:
Interesting for testing when available

rafaelbn:
cc @yajo @pedrobaeza

yajo:
What's the use case for this?

OSguard:
the aim was to have a module to easy explore all models and fields of your current installation.
You have a similar feature in the technical feature but then you can not easy click on relation.

This is morge for develop purpose then for production system.
We try to make it independent from "website" and only use "web" module but then we need to implement much more feature to support a nice template. So we skip this.

yajo:
> but then you can not easy click on relation

What do you mean with this?

OSguard:
Compare:
http://3148243-1-414c31.runbot2.odoo-community.org/doc/jsonrpc/invoice-line-274

with:
http://3148243-1-414c31.runbot2.odoo-community.org/web#id=274&view_type=form&model=ir.model&menu_id=29&action=13

Then you can see you can just click on "account.invoice"

The idea behind was to give a NON-Odoo-Developer a nice view for a XML/JSON-RPC API doku

yajo:
Why not just add this link field in the backend and a portal group with read access to those fields? This seems a little overkill honestly.

OSguard:
This was the first PR for the new website feature almost 2 years ago. It was more a PoC to figure out want we can do with the new Core feature and gives us more opportunities then hacking JS widget for backend.
So i fine if we close this. If someone needs this code. He can find it via odoo-code-search and use it. But we don't need then to maintain this in OCA if no one is interested.

yajo:
I'm not going to vote against this just because I don't need it, but neither will vote up because I think core features are enough for this purpose. Let others talk. Thanks for your understanding :wink:

OSguard:
@rafaelbn @pedrobaeza @dreispt 
can you give a vote if we want this or close the PR?

rafaelbn:
I would like to have the opinion of website fighters. Please @antespi @sergio-teruel @carlosdauden could you vote here?
Any case: if you @OSguard are not using it anymore maybe that is enough yo close it.
Let's see this week whay people say

rafaelbn:
HI @OSguard ,

As nobody answer then my vote is to merge o close depending of your needs. If you are going to maintain and migrate to v9 go on, if not maybe is better to close it.

dreispt:
This is not TraviCI checked, and there are a few guidelines I 'm sure need fixing.
Can you rebase on latest website 8.0?

website_doc_jsonrpc/__openerp__.py
```diff
@@ -0,0 +1,52 @@
...
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': 'Website Odoo JSON RPC doc',
+    'version': '0.1',
+    'author': 'initOS GmbH & Co. KG',
+    'category': '',
+    'description': """Generates a documentation for JSON RPC
+    for the current installation on the fly""",
```
dreispt, line 29: I'm sorry, it's not clear to me the purpose of this module.

rafaelbn:
Hi, I'm closing this old PR. If needed please make a new PR.

PR #2, opened 2014-08-17
German website sale additions

mncoon:
Localization for the `website_sale` to make it ready for B2C in Germany.


mncoon:
I close this pull request becaus we have a German community, where the module is better off.

PR #6, opened 2015-01-14
website_crm_mandatory_phone_number.

foutoucour:
Simple module to set the field phone number in the Contact Form as mandatory


pedrobaeza:
Some line length issues. Please solve them.

website_crm_mandatory_phone_number/__openerp__.py
```diff
@@ -0,0 +1,70 @@
...
+    'name': 'Website CRM Contact Mandatory Phone Number',
+    'version': '0.1',
+    'author': 'Savoir-faire Linux',
+    'maintainer': 'Savoir-faire Linux',
+    'website': 'http://www.savoirfairelinux.com',
+    'license': 'AGPL-3',
+    'category': 'Website',
+    'summary': 'This module set as required the phone number.',
+    'description': """
+Website CRM Contact Mandatory Phone Number
```
pedrobaeza: You can extract this to a README.rst file to have compatibility with both GitHub and Odoo, and remove it from here.

pedrobaeza:
You should remove company logo from module.

pedrobaeza:
One last thing, please rename it to _website_crm_phone_required_, that it's more explicit and follow current Odoo naming.

StefanRijnhart:
Do we really want to have modules that only set an attribute on a single field in a single view in the community repositories? Just curious after other people's opinions.

pedrobaeza:
You're right that it's a little weight, but if it's a common need, I vote for its inclusion.

website_crm_mandatory_phone_number/views/website_crm.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
+    <data>
+        <template id="website_crm_mandatory_phone_number_template" name="Phone Number Mandatory"
+                  inherit_id="website_crm.contactus_form" customize_show="True">
+            <xpath expr="//input[@name='phone']" position="attributes">
+                <attribute name="required">True</attribute>
```
hbrunn, line 7: as of http://www.w3.org/TR/2008/WD-html5-20080610/semantics.html 3.2.2, this should be `required` and not `True`

rafaelbn:
I think is more common that the standard required fields becomes not required (as "subject" or "your question" or "your company"). I mean that in the same way we will have at least 4 or 5 modules to manage the standard form of Odoo.
What do you think about the idea of a configuration menu in Setting - Configuration - Website where you can manage all fields of the form?

rafaelbn:
Hi, what do you think about this matter?
cc @yajo @sergio-teruel @carlosdauden @antespi

yajo:
IMHO to solve this problem we should PR Odoo itself with a patch that requires that either the phone or the email fields are filled. If you have a question from somebody and cannot answer (it happened to me), it's definitely a bug in core Odoo.

StefanRijnhart:
@yajo the email address is required by default.

I guess this is a special case so we may as well merge it and be done with it. :+1:

rafaelbn:
I would like to teste this un runbot but runbot is not available in thie PR. @foutoucour could yo review this to test it before merging? Thanks

rafaelbn:
@hbrunn @StefanRijnhart I cannot test it un runbot, I don't want to merge this without testing. @foutoucour doesn't look answering anymore.

rafaelbn:
@StefanRijnhart it looks @foutoucour is not answering anymore. No runbot. Have anyone tested functionally?

rafaelbn:
Hi @OCA/website-maintainers I vote to close this PR as original author doesn't answer. Are you agree?

StefanRijnhart:
Indeed @rafaelbn, let's close this one.

PR #9, opened 2015-01-20
[ADD] website_backend_views

hbrunn:
# Backend views for website

This module allows to embed old style (backend) views into the website frontend. This way, complex forms don't need to be redeveloped but still there's no break in user experience. Caveat is that this does not work for anonymous users.

This is specifically convenient if you need to use on_change functions, user input dependent domains or got used to be able to create linked objects inline.
# Usage

Install this module with demo data enabled and visit `/website_backend_views/demo`. Then read `view/demo.xml`.
# Known issues / Roadmap
- more testing needed
- not all css styles from the backend apply (should be just a matter ob having the right html structure/classes attached to the container
- maybe we can do something for anonymous users?
- view switching?
# Credits
## Contributors
- Holger Brunn hbrunn@therp.nl
## Maintainer

.. image:: http://odoo-community.org/logo.png
:alt: Odoo Community Association
:target: http://odoo-community.org

This module is maintained by the OCA.

OCA, or the Odoo Community Association, is a nonprofit organization whose mission is to support the collaborative development of Odoo features and promote its widespread use.

To contribute to this module, please visit http://odoo-community.org.


website_backend_views/README.rst
```diff
@@ -0,0 +1,40 @@
+Backend views for website
+=========================
+
+This module allows to embed old style (backend) views into the website frontend. This way, complex forms don't need to be redeveloped but still there's no break in user experience. Ceveat is that this does not work for anonymous users.
```
pedrobaeza: s/Ceveat/Caveat/

website_backend_views/README.rst
```diff
@@ -0,0 +1,40 @@
...
+Usage
+=====
+
+Install this module with demo data enabled and visit `/website_backend_views/demo`. Then read `view/demo.xml`.
+
+Known issues / Roadmap
+======================
+
+* more testing needed
+* not all css styles from the backend apply (should be just a matter ob having the right html structure/classes attached to the container
```
pedrobaeza: s/ob/of/ and please close parenthesis

website_backend_views/README.rst
```diff
@@ -0,0 +1,40 @@
...
+* maybe we can do something for anonymous users?
```
pedrobaeza, line 18: With portal permissions I think it's enough. Don't you?

pedrobaeza:
Please remove Therp logo

hbrunn:
Thanks Pedro, you review faster than I can notice that I didn't clean up the logo from my default template ;-)

website_backend_views/README.rst
hbrunn, line 18: Yes, but that's a user. The whole backend framework needs a uid, but I think we could create one without permissions (~portal) and su all calls to that. But it's more tricky than it sounds I think

pedrobaeza, line 18: Yeah, very tricky. A use case that I can imagine for this is to replace a form to fill by a backend view, but this have to be done as a very special mode with user notice I think.

guewen:
Wow
LGTM (code review, no test)

pedrobaeza:
:+1:

PR #14, opened 2015-02-12
[ADD] analytics_piwik

hbrunn:
# Piwik analytics

This module allows to simply configure you odoo website for using Piwik_.

.. _Piwik: http://piwik.org
# Configuration

To configure this module, you need to:
- go to Settings/Configuration/Website Settings
- fill in `Piwik website ID` and `Piwik host`
- Take care that you fill in the correct value for `Piwik host`, it should be
  so that http[s]://[`Piwik host`]/piwik.php resolves to your copy of `piwik.php`


pedrobaeza:
Why not website_analytics_piwik?

hbrunn:
I can do that if you want

pedrobaeza:
I think it's a better name that implies the "area".

hbrunn:
done

website_analytics_piwik/models/website_config_settings.py
```diff
@@ -0,0 +1,30 @@
...
+#
+##############################################################################
+from openerp import models, fields, api
+
+
+class WebsiteConfigSettings(models.Model):
+    _inherit = 'website.config.settings'
+
+    piwik_analytics_id = fields.Integer(
+        related = ['website_id', 'piwik_analytics_id'])
```
pedrobaeza: Curious form of putting a related field. I always put it this way: `related="website_id.piwik_analytics_id"`

Anyways, don't put spaces in a keyword argument.

website_analytics_piwik/views/templates.xml
```diff
@@ -0,0 +1,22 @@
...
+<openerp>
+    <data>
+        <template id="layout" inherit_id="website.layout">
+            <xpath expr="html/body" position="inside">
+                <script t-if="website and website.piwik_analytics_id and website.piwik_analytics_host" type="text/javascript">
+                    var _paq = _paq || [];
+                    _paq.push(['trackPageView']);
+                    _paq.push(['enableLinkTracking']);
+                    (function() {
+                    var u="//<t t-esc="website.piwik_analytics_host" />/";
```
pedrobaeza: t-esc or t-raw?

website_analytics_piwik/models/website_config_settings.py
hbrunn: it's the new way to do it (tm): https://github.com/odoo/odoo/blob/8.0/openerp/fields.py#L451

website_analytics_piwik/views/templates.xml
hbrunn: yes, raw makes more sense here, thanks!

website_analytics_piwik/models/website_config_settings.py
pedrobaeza: I still feels more readable the other way, but I'm saying only as opinion, not to make you to change it.

pedrobaeza:
:+1: 

Thanks

lmignon:
:+1:  Thanks

antespi:
:+1:

PR #16, opened 2015-02-26
[ADD] website_hr_department

lmignon:
New module to display the departments in the website


website_hr_department/__openerp__.py
```diff
@@ -0,0 +1,43 @@
...
+#     You should have received a copy of the
+#     GNU Affero General Public License
+#     along with website_hr_department.
+#     If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+{
+    'name': "Departments Page",
+    'summary': """
+        Display the structure of your departments and their members.
```
pedrobaeza, line 27: Don't put triple quota on summary. It introduces line breaks.

sbidoul:
:+1: (functional test only)

website_hr_department/__openerp__.py
sbidoul, line 27: Note this style comes from the Odoo default module template:
https://github.com/odoo/odoo/blob/8.0/openerp/cli/templates/default/__openerp__.py

pedrobaeza, line 27: Well, have you tested if it appears well?

lmignon, line 27: yes it appears well
image
image

pedrobaeza, line 27: OK then, thanks for the explanation

pedrobaeza:
:+1: (code review)

hbrunn:
:+1:

PR #17, opened 2015-03-03
[ADD] website_event_register_free_with_sale

pedrobaeza:
# Register for free website events - Sale extension

Extension for module _website_event_register_free_ that allows to combine
both kind of registrations: free and paid.
# Installation

This module will be auto-installed when _website_event_register_free_ and
_website_event_sale_ are both installed.
# Usage

You have two options for events:
- Don't put any type of ticket, so the event will be considered as free, and
  the registration screen of _website_event_register_free_ module will appear.
  _Warning:_ This doesn't work for now due to a bug in Odoo that doesn't allow
  to store max seats available.
- Define ticket types, but let one or some of the tickets with price 0. When
  you go to normal registration screen, the tickets with price will be added
  to shopping cart, and the priceless tickets will be automatically registered.
  If there are no free tickets selected, then normal checkout screen will
  appeared.
# Known issues / Roadmap
- Hide on checkout page the fields that are not needed for free events.


hbrunn:
nice, that's what I had in mind! Just one question about the checkout process: To me, it seems free tickets always win and you will have to press 'checkout' after finishing the registration for free tickets. (I'm talking about the case now where you order a free ticket as well as a paid one)

Couldn't this be turned around so that in the mixed case, you go to the normal checkout, fill in your details and they will be used for the free registration?

pedrobaeza:
That's was more difficult to implement, so I opted for the fast approach of doing this way.

hbrunn:
understandable, maybe mention that in the readme?

pedrobaeza:
@hbrunn, done!

hbrunn:
:+1:

antespi:
:+1:

pedrobaeza:
@hbrunn, good news! My customer has financed to change the module to use the same checkout screen, so here it is! This depends on  #18, so please review first that.

rafaelbn:
:+1:

rafaelbn:
Is it ready to be merged?

pedrobaeza:
This can be considered the final version. I have squashed all commits.

PR #23, opened 2015-04-16
[FIX] the website uses its own qweb model by now

hbrunn:


pedrobaeza:
:+1:

eLBati:
LGTM :+1: 

(ref: https://github.com/odoo/odoo/commit/510a4f51d2a5b75f7b4c3b8b345f693eb044b337)

website_backend_views/model/ir_qweb.py
```diff
@@ -24,7 +24,7 @@
 
 
 class IrQweb(models.Model):
```
yvaucher: Don't you want to rename the model as well?

hbrunn: @yvaucher I was tempted to quote Shakespeare on this one, but you're right of course. As a followup, the file name and the import have to change too

yvaucher:
@hbrunn Thanks :+1:

website_backend_views/model/ir_qweb.py
yvaucher: @hbrunn I prefer to quote Einstein about simplicity. Though his sentence has less soul.

PR #26, opened 2015-04-17
[ADD] website_sale_order_company

eLBati:
# e-commerce order company

This module sets online (ecommerce) order's company according to purchased products.
# Usage

Set 'company' field for products that must be sold by that company.
If a product can be sold by any company, just leave company field empty.


pedrobaeza:
Can you explain the reasoning behind this module?

eLBati:
@pedrobaeza updated https://github.com/eLBati/website/blob/website_sale_order_company/website_sale_order_company/README.rst
See if it's clearer

pedrobaeza:
OK, I see. It's a little twist to get a multi-company shop, but it can serve until v9 arrives.

website_sale_order_company/models/sale_order.py
```diff
@@ -0,0 +1,41 @@
...
+        Check products companies. If all products have the same company
+        (or empty), return it
+        """
+        companies = {}
+        company = False
+        for order in self:
+            for line in order.order_line:
+                if line.product_id and line.product_id.company_id:
+                    companies[line.product_id.company_id] = True
+        if len(companies) == 1:
```
pedrobaeza, line 42: Shouldn't you return companies[0]?, because a sale order cannot be without company.

eLBati, line 42: If `get_products_company` doesn't find a unique company, the default one id used (returned by '_get_default_company' of 'sale.order')

Also updated comment

pedrobaeza:
Thanks :+1:

hbrunn:
:+1:

Issue #27, opened 2015-04-20
website_event_register_free_with_sale: Error 500 when buying a Event with price


rafaelbn:
1 - 
![20-04-2015 14-18-56](https://cloud.githubusercontent.com/assets/8736623/7229857/a9ab4dcc-e768-11e4-8351-e25556e5ae7a.png)
2 - 
![20-04-2015 14-20-24](https://cloud.githubusercontent.com/assets/8736623/7229862/b0bf54a0-e768-11e4-939d-9acf53b96e09.png)
3 - 
![20-04-2015 14-20-40](https://cloud.githubusercontent.com/assets/8736623/7229867/b70f9ac2-e768-11e4-94d8-d0bc3ffe3236.png)
4 - Log:
File "/var/openerp/project/website_event_register_free_with_sale/controllers/website_sale.py", line 56, in checkout
return super(WebsiteSale, self).checkout(self, *_post)
File "/var/openerp/common/openerp/v8/openerp/http.py", line 399, in response_wrap
response = f(_args, **kw)
TypeError: checkout() takes exactly 1 argument (2 given)

pedrobaeza:
It seems you don't have last version, because the line miss one `*`:

https://github.com/OCA/website/blob/8.0/website_event_register_free_with_sale/controllers/website_sale.py#L56

rafaelbn:
@JavierIniesta could you check this please?

PR #31, opened 2015-05-25
[FIX] website_event_register_free: Error 500 when an not public user …

sergio-teruel:
…register into an event.

[ADD] website_event_register_free: Spanish translation.


pedrobaeza:
:+1:

hbrunn:
this breaks website_event_register_free_with_sale: https://github.com/OCA/website/blob/8.0/website_event_register_free_with_sale/controllers/website_sale.py#L76 - please also include appropriate changes to this module.

And being at it, rename `partner_id` to `partner` in order to reflect what kind of data we're dealing with.

pedrobaeza:
The change is done. @sergio-incaser, please notify putting a comment when you make a change so that we can be informed.

Can you please sign OCA CLA?

sergio-teruel:
I already have signed and I've emailed it.

pedrobaeza:
@gurneyalex, can you please check this CLA?

gurneyalex:
@pedrobaeza @sergio-incaser CLA received (pending processing)

OSguard:
need to rebase for runbot issue

rafaelbn:
I have tested and it works, please @sergio-incaser test it again please and define exactly the bug. Thanks

rafaelbn:
@sergio-incaser could you review this?. As @pedrobaeza has close this one https://github.com/OCA/website/issues/27 maybe this has been solved too.
I have tested again and it work in runbot.
Try in http://3113813-8-0-4b9709.runbot2.odoo-community.org/ register to Event TEST which is free.
thanks!

sergio-teruel:
@rafaelbn I have tested and the issue continue.
In _prepare_registration:
            'email': post['email'],
            'phone': post['phone'],
            'name': post['name'],
If any of this key not come in post the server give an 500 error.
This occurs when a registered user reserve a ticket

rafaelbn:
Tested in runbot :+1:

website_event_register_free/model/event_registration.py
```diff
@@ -24,15 +24,15 @@
...
             'date_open': fields.Datetime.now(),
-            'email': post['email'],
-            'phone': post['phone'],
-            'name': post['name'],
+            'email': partner and partner.email or post['email'],
+            'phone': partner and partner.phone or post['phone'],
+            'name': partner and partner.name or post['name'],
             'user_id': user_id,
-            'partner_id': partner_id,
+            'partner_id': partner and partner.id,
```
pedrobaeza, line 37: This comparison is useless. Put directly `partner.id`

website_event_register_free/model/event_registration.py
```diff
@@ -24,15 +24,15 @@
...
+    def _prepare_registration(self, event, post, user_id, partner=False):
         return {
             'origin': 'Website',
             'nb_register': int(post['tickets']),
             'event_id': event.id,
             'date_open': fields.Datetime.now(),
-            'email': post['email'],
-            'phone': post['phone'],
-            'name': post['name'],
+            'email': partner and partner.email or post['email'],
```
pedrobaeza: You are giving more priority to the store field instead of the possible data filled on the form. You should invert it.
`post.get('email') or partner.email`

pedrobaeza: Can you answer this?

rafaelbn:
@sergio-incaser after changes I can test again. Thanks

website_event_register_free/model/event_registration.py
sergio-teruel, line 37: @pedrobaeza in this function partner is an optional argument with False as default value so i need check before.

pedrobaeza, line 37: OK, sorry. I was thinking in partner as an empty recordset, but it's not the case.

rafaelbn:
:+1:

rafaelbn:
This PR has 2 +1 and FIX a BUG in a module in OCA. Could anyone make last review to merge and FIX the BUG? Thanks

PR #35, opened 2015-05-29
[ADD] website canonical url

tremlin:
The proposed module adds a HTML tag to the header that contains the canonical URL for the current page.

See also http://en.wikipedia.org/wiki/Canonical_link_element


website_canonical_url/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#############################################################################
+
+import models
```
OSguard: from . import models

rafaelbn:
Hi @tremlin could you add usage in the readme file? I would like to know what I must test. Thanks

rafaelbn:
Hi @tremlin could you please review readme?
@OSguard what do you think about this PR?
cc @Yajo @pedrobaeza

website_canonical_url/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+#
+#    This program is distributed in the hope that it will be useful,
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#############################################################################
```
yajo: Please replace with new style reduced license headers.

website_canonical_url/__openerp__.py
```diff
@@ -0,0 +1,46 @@
...
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#############################################################################
+{
+    'name': "Website Canoncial URL",
+
+    'summary': """
+        Canonical URL in Website Headers""",
```
yajo: Please put in 1 line.

website_canonical_url/__openerp__.py
```diff
@@ -0,0 +1,46 @@
...
+
+    'author': "initOS GmbH & Co. KG",
+    'website': "http://www.initos.com",
+
+    'category': 'Website',
+    'version': '0.1',
```
yajo: 8.0.1.0.0

website_canonical_url/__openerp__.py
```diff
@@ -0,0 +1,46 @@
...
+    ],
+    'data': [
+        'views/templates.xml',
+    ],
+
+    'demo': [
+    ],
+
+    'tests': [
+    ],
```
yajo: Remove empty keys

website_canonical_url/models/website.py
```diff
@@ -0,0 +1,35 @@
...
+#    but WITHOUT ANY WARRANTY; without even the implied warranty of
+#    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#    GNU Affero General Public License for more details.
+#
+#    You should have received a copy of the GNU Affero General Public License
+#    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#############################################################################
+
+from openerp.osv import orm
```
yajo: from openerp import models

website_canonical_url/models/website.py
```diff
@@ -0,0 +1,35 @@
...
+from openerp.http import request
+
+
+class website(orm.Model):
+    _inherit = 'website'
+
+    def get_canonical_urls(self, cr, uid, ids, req=None, **kwargs):
```
yajo: Use new api please.

website_canonical_url/views/templates.xml
```diff
@@ -0,0 +1,16 @@
...
+                <t t-if="request and website">
+                    <t t-foreach="website.get_canonical_urls(request.httprequest)" t-as="canonical_url">
+                        <link rel="canonical" t-att-href="canonical_url" />
+                    </t>
+                </t>
+            </xpath>
+        </template>
+
+    </data>
+</openerp>
```
yajo: newline at end

website_canonical_url/README.rst
```diff
@@ -0,0 +1,14 @@
...
+=====================
+
+This module adds a HTML tag in the header that contains a canoncial URL for the current page.
+
+
+Contributors
+------------
+
+* Thomas Rehn <thomas.rehn@initos.com>
+
```
yajo, line 43: Please use the newest README template.

website_canonical_url/models/website.py
```diff
@@ -0,0 +1,35 @@
...
+        canonical_urls = []
```
yajo: What's the point of using a list? AFAIK there's only 1 canonical URL (that's its purpose after all).

website_canonical_url/models/website.py
```diff
@@ -0,0 +1,35 @@
...
+        if req is None:
+            req = request.httprequest
+        if req and req.base_url:
+            canonical_urls.append(req.base_url)
```
yajo: Bug: not working when using a different language (it should start with the lang part).

yajo:
The lack of this feature should be considered a bug, SEO POV. Could you please open PR to Odoo/master? Let's see if after some years we will not need this module.

rafaelbn:
Hi @tremlin , please take a look in comments. Thanks

tremlin:
Sorry for the delay and thanks for all the feedback! We've tried to address all the issues, the multi-language handling is a bit cumbersome, but should work now, too.
Once this module has passed review, we will make a PR against Odoo.

rafaelbn:
Runbot fails.
@Yajo this PR is for v8, Odoo will not consider so this module will be useful in v8 websites. @tremlin will make a PR to Odoo master so this is enough.
Let's review and get merge

yajo:
Much better now. You forgot about pagers. They add `/page/X` (X = any number from 2).

I think they are supposed to have a canonical url without that part, no?

tremlin:
As far as I understand the concept of canonical URLs, pagination should lead to different canonical URLs per page since the content is different on each page.

yajo:
You're right, [`<link rel="next/prev">` should be used instead](https://support.google.com/webmasters/answer/1663744?hl=en). If there's a `pager` env variable ([the one used by `website.pager` template](https://github.com/odoo/odoo/blob/8.0/addons/website/views/website_templates.xml#L382-L394)) you could add them too.

Well, it's just a suggestion, but it would certainly fit into this module (and your future PR to odoo/master).

If you want to add that, it will be great. Otherwise you still get my :+1: (code & functional).

rafaelbn:
Thanks. Functional :+1:

pedrobaeza:
@tremlin, if you don't add @Yajo 's suggestions, please at least add this to Known issues section.

rafaelbn:
@tremlin you only have to achieve @Yajo suggestions or just add them to known issue section. After this we will merge this PR. Thanks a lot!

tremlin:
@rafaelbn Sorry for my delay. I've just added pagination support as suggested by @Yajo. Thanks for the suggestion!

paolovalier:
:+1:

yajo:
Functional & code :+1:

PR #42, opened 2015-06-17
[FIX] checkout_values not autocomplete the fields partner

Endika:
In website_sale in checkout_values checked "data" is empty
https://github.com/odoo/odoo/blob/8.0/addons/website_sale/controllers/main.py#L382
in this case complete partner data https://github.com/odoo/odoo/blob/8.0/addons/website_sale/controllers/main.py#L392
But in the other case partner is empty and not completed in the form.

### DIFF
--- a/website_event_register_free_with_sale/controllers/website_sale.py
+++ b/website_event_register_free_with_sale/controllers/website_sale.py
@@ -51,7 +51,8 @@ class WebsiteSale(website_sale):
         order = request.website.sale_get_order(force_create=0)
         has_paid_tickets = bool(order.order_line)
         if request.session.get('free_tickets') and not has_paid_tickets:
-            values = self.checkout_values(data={'shipping_id': -1})
+            values = self.checkout_values()
+            values['shipping_id'] = -1
             return request.website.render("website_sale.checkout", values)
         else:
             return super(WebsiteSale, self).checkout(**post)


antespi:
:+1:

antespi:
@pedrobaeza this PR fix an issue when website_event_register_free_with_sale installed, autocomplete form doesn't work when user is logged in

pedrobaeza:
One of the reason of not passing shipping_id was for avoiding the creation of the sale order in free events. Are you sure this patch doesn't modify this?

antespi:
@pedrobaeza we're going to check it right now

Endika:
@pedrobaeza @antespi I tested this case with this Fix and without this, and in both cases created a new sale order in draft state.

Endika:
@pedrobaeza @antespi  Ok sorry, i tested in clean BBDD and both cases not create a new sale order.

pedrobaeza:
It is not a question of a clean DB. You have to test it with a free event. Does it work in this case?

Endika:
Yes, in this case works.

pedrobaeza:
:+1: then

antespi:
@rafaelbn rebase done

rafaelbn:
I have tested in runbot after @antespi 's rebase. @pedrobaeza my steps:
- I created a new res.user with a res.partner associated
- Created a new event, free!
- Logged with this new res.user I have gone to Events and reserved 1 ticket
- Form is autocompleted with all data of Partner --> OK
- Sale Order with this FREE event is not created --> OK

There is only one thing we must fix, in the checkout form the field "shipping" is by default in **"ship to the same address"** and bellow this **doesn't appears anything else**, when you select **"create a new address"** it appears a new **form for the new shipping address**. This happens with non-free events, with free events even though "ship to the same address" is selected appears the form of "create a new address" ant it mustn't. Please check screen-shoot for more detailed:

JavierIniesta:
I have tested in runbot. Sales order is not created. Form is autocompleted with the partner data.

But "Company Name" label shows "street" field and "Street" label shows "street2". Is this correct?

rafaelbn:
We need to review this FIX @pedrobaeza @Yajo @antespi . Thanks @Tecnativa

rafaelbn:
@pedrobaeza could you please make a quick review here please to achieve the solutions. @Endika is not going to continue this PR anymore I guess. This is a FIX. If not we can close it.

rafaelbn:
I think @Yajo you should continue this. @Endika (hello my friend! :smiley: ) is not going to continue this PR I guess.

What do you think @pedrobaeza ?

rafaelbn:
To much time without an answer :cry:

PR #57, opened 2015-07-24
[FIX] website_sale_vat_required: error when checking out with logged user:

eLBati:
Element '<label class="control-label" for="vat" data-oe-id="1725" data-oe-source-id="1506" data-oe-xpath="/data/label" data-oe-model="ir.ui.view" data-oe-field="arch">' cannot be located in parent view

### DIFF
--- a/website_sale_vat_required/website_order_view.xml
+++ b/website_sale_vat_required/website_order_view.xml
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <openerp>
 <data>
-    <template id="saas_users" inherit_id="website_sale.checkout">
-        <label class="control-label" for="vat" position="attributes">
+    <template id="vat_checkout" inherit_id="website_sale.checkout">
+        <xpath expr="//div[@class='container oe_website_sale']/form/div[@class='row']/div[@class='col-md-8 oe_cart']/div[@class='row']/div[4]/label" position="attributes">
             <attribute name="style"></attribute>
-        </label>
+        </xpath>
     </template>
 </data>
 </openerp>


website_sale_vat_required/website_order_view.xml
```diff
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <openerp>
 <data>
-    <template id="saas_users" inherit_id="website_sale.checkout">
-        <label class="control-label" for="vat" position="attributes">
+    <template id="vat_checkout" inherit_id="website_sale.checkout">
+        <xpath expr="//div[@class='container oe_website_sale']/form/div[@class='row']/div[@class='col-md-8 oe_cart']/div[@class='row']/div[4]/label" position="attributes">
```
pedrobaeza, line 5: Can you simplify this xpath expression or is not possible?

nhomar, line 5: Can you even add an option="True" in order to have this template as optional with the module installed.

Why? you can simply add a test on your own activating it but you will not brake other tests that inherit the same view buit do not depend of this feature.

eLBati, line 5: @pedrobaeza I didn't find a way to simplify it. Do you have suggestions?

@nhomar do you have an example about `option="True"`?

pedrobaeza, line 5: No, I was just asking. OK then

rafaelbn:
Hi @eLBati , could you check this. Testing in runbot I get this error in shop, buying an ipad

```
500: Internal Server Error
Error

Error message:

Element '<label class="control-label" for="vat" data-oe-id="910" data-oe-source-id="889" data-oe-xpath="/data/label" data-oe-model="ir.ui.view" data-oe-field="arch">' cannot be located in parent view

Error context:
View `saas_users`
[view_id: 910, xml_id: website_sale_vat_required.saas_users, model: n/a, parent_id: 889]
```

thanks

pedrobaeza:
Superseeded by https://github.com/OCA/website/pull/75

PR #60, opened 2015-07-31
The sass file was lying.

yajo:
Changed to match the CSS contents (and stop my Sass compiler's complains).


pedrobaeza:
This was a verbatim copy of an in progress PR, so maybe this is incorrect, but I didn't use it, so that's why I didn't find out.

:+1:

yajo:
I see some tests fail, but I don't believe it's my fault.

pedrobaeza:
It's due to the change of location of wkhtmltopdf project

yajo:
So nothing to fix here?

pedrobaeza:
I think so

yvaucher:
:+1:

PR #61, opened 2015-08-02
Template extension correction

sysadminmatmoz:
Nothing much, just cleaning my translations infrastructure and correcting the translations templates mistakes I find out in all the OCA repos


pedrobaeza:
:+1: I fast-track this

PR #63, opened 2015-08-04
Website Social Media Icon Extention

Benniphx:
Extension for the social media icons from the odoo core.
For Xing / Dribbble / Tripadvisor


pedrobaeza:
Thanks for your contribution, Benjamin!

I think it's more interesting to split the module in each of the possible social media places you add, calling them _website_social_tripadvisor_, _website_social_dribbble_ and so on, and https://github.com/OCA/social can be a best host, because this repository starts to be very bloated with all kind of website modules (as the topic itself is very wide).

dreispt:
Pedro, it's neater and I'm OK with having a single module:
If someone want to add yet another social network, it would do so extending this module instead of creating new ones.

pedrobaeza:
I don't think it's the correct approach, because we will end up with a bloated module like account_currency_update, and it makes the migration of these modules a lot of more complicated.

qtheuret:
I'm not for a multiplication of modules but I would like to know why it is more complicated to migrate a single module than multiple ones. Maybe it's a good argument.

pedrobaeza:
The question is simple: more code, more difficult to migrate because the amount of work, checks and so on increases. If you have a module per site, a person interested in migrating that site will probably do it and contribute it, but if you have one module with all sites (and increasing), this will discourage the person to do it because all the work to do it, so probably patch in his/her own repository the part of that site, and don't contribute it back.

LeartS:
I agree with @pedrobaeza , one per site would be cleaner. Odoo can handle dozens of modules installed without much problems, especially small ones as these. OCA/Social at the moment is empty and its scope is narrower than OCA/website, so it can welcome these site-specific modules, 3 or 15 that may be.
Also, with one per site they can have the site name and icon in the module name and icon, which greatly improves discoverability.
To add new ones, one would only need to take a random one as template and work from there.

Benniphx:
ok I will split them an when I will a do pull request to OCA/Social

Benniphx:
OCA/social#7
OCA/social#6
OCA/social#5

PR #64, opened 2015-08-05
[9.0] [ADD] added a website switcher

Benniphx:
If you are using different websites this Add-on will let you switch easily through them.


nhomar:
Why the 9.0 label? over which odoo's repository are you working on?

lmignon:
@Benniphx I'm not aware that multi website domains are supported in 8.0. The only way I know to get multi domain in v8 is to use https://github.com/odoo/odoo-extra/blob/8.0/website_multi

Benniphx:
Hi, this is an Add-on for the current master (9.0 alpha) where you have multi website integrated. I know that there is no 9.0 branch in the OCA right now, but I wanted to share this Add-on. Or should I contribute it later?

pedrobaeza:
As current master is still too volatile, I don't recommend you to do anything on top of it, because in some days probably this module won't work. OCA also doesn't support addons to master branch.

Benniphx:
Ok. than pull request can be closed and I will contribute it again when 9.0 is released.

pedrobaeza:
Thanks anyway for the contribution.

PR #24, opened 2015-04-17
[ADD] website_sale_vat_required

eLBati:
This module extends checkout e-commerce form in order to force user to fill 'VAT number' field

PR #75, opened 2015-08-25
[FIX] Greenify Travis

pedrobaeza:


rafaelbn:
Travis fails and maybe not related with my test but I have tested and even though website_sale_vat_required is installed checkout doesn't ask you about VAT and you cannot finish it (you cannot finish any purchase). It must be a VAT field in checkout form.
I have tried to test the original PR https://github.com/OCA/website/pull/24 of this module but runbot it not available.
cc @eLBati

.travis.yml
```diff
@@ -13,13 +16,23 @@ python:
...
+  - VERSION="8.0" TESTS="0" LINT_CHECK="0" TRANSIFEX="0"
+  - TRANSIFEX_USER='transbot@odoo-community.org'
+  - secure: "eNfbQI7Pkp3mjTFQgjrVVldmyE/Yn3+iCMVykhXppC85dJaCFMpjwPUfZa7p38tnnH7jhng2JL2tAXnJ4uOry0B3m9xYR7fNYXg/9uj1988FHvpDhfDFEKcMF8s70mr11HRijj6U8p6WDt/0ygD6aZ0LJzWzxhMFecFavOb6XOw="
+  matrix:
   - LINT_CHECK="1"
-  - VERSION="8.0" ODOO_REPO="odoo/odoo" LINT_CHECK="0"
-  - VERSION="8.0" ODOO_REPO="OCA/OCB" LINT_CHECK="0"
+  - TRANSIFEX="1"
+  - TESTS="1" ODOO_REPO="odoo/odoo" EXCLUDE="website_sale_vat_required"
+  - TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE="website_sale_vat_required"
```
eLBati, line 27: Don't we also need the INCLUDE lines?
@pedrobaeza do you know why EXCLUDE / INCLUDE are not mentioned at
https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.travis.yml
?

website_sale_vat_required/website_order_view.xml
```diff
@@ -1,10 +1,10 @@
...
 <openerp>
 <data>
-    <template id="saas_users" inherit_id="website_sale.checkout">
-        <label class="control-label" for="vat" position="attributes">
-            <attribute name="style"></attribute>
-        </label>
+    <template id="checkout_vat_required" inherit_id="website_sale.checkout">
+        <xpath expr="//label[@for='vat']" position="attributes">
+            <attribute name="style"/>
+        </xpath>
```
eLBati, line 7: Just tested on runbot (both with logged and not logged user) and I can't find VAT field on checkout form

pedrobaeza, line 7: This is still a WIP. I can't even get a green, so I haven't worried about that yet, but in local I got the field shown.

pedrobaeza:
Superseeded by https://github.com/OCA/website/pull/97

PR #62, opened 2015-08-03
[ADD] website_openstreetmap module to replace google maps on website

gfcapalbo:

PR #76, opened 2015-09-01
[ADD] website_dynamic_map

dgzurita:
This module was written to extend the functionality of Google Maps static map on "Contact Us" view (company address) to support markers and allow you to choice between Google Maps static or dynamic map and to write the precise coordinates of a partner address.


rafaelbn:
Absolutely great module! This make sense and useful the default contact us page of Odoo...

Tested:
- In "static map" selected please review what happend when clicking in the map. Yes it redirects yo Google maps but without marker,
- In "dynamic map" the map **doesn't show marker** and show a lot of no-necessary information. It must show only the map with the marker. Check image:

antespi:
@rafaelbn Fix latitud and longitud precision. You can try again, thanks!

rafaelbn:
In "static map" selected please review it show in correct map but when click on marke it goes to correct address in Google Maps.



In "dynamic map" --> Solved!

rafaelbn:
Check module base_geolocalize:

antespi:
@rafabn Squash and rebase

Now static map doesn't include Partner name, like Odoo standard does.

Latitude and longitud parameters are recommended only when company address has no geo localization in GoogleMap, so in these cases `base_geolocalize` is not working too.

antespi:
@rafaelbn I think your issue is fixed, please check it!

janetrp:
tested in runbot :+1:

rafaelbn:
Super :+1:

website_dynamic_map/README.rst
```diff
@@ -0,0 +1,80 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+===================
+Website dynamic map
+===================
+
+This module was written to extend the functionality of Google Maps static map on "Contact Us" view (company address) to support markers
+and allow you to choice between Google Maps static or dynamic map and to write the precise coordinates of a partner address.
```
pedrobaeza, line 10: s/allow/allows

website_dynamic_map/README.rst
```diff
@@ -0,0 +1,80 @@
...
+To configure this module, you need to:
+
+* Go To Settings -> Companies -> "Your Company"
+* Edit
+* Fill the fields in "General information" tab
+
+Usage
+=====
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
```
pedrobaeza, line 24: Put here go to website > Contact us and so on

website_dynamic_map/__openerp__.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
+# See README.rst file on addon root folder for license details
+
+{
+    'name': "Website dynamic map",
+    'category': 'Personalization',
```
pedrobaeza, line 6: Put the same as the one of website (I think is `Website`)

website_dynamic_map/__openerp__.py
```diff
@@ -0,0 +1,29 @@
...
+    'js': [
+    ],
+    'author': 'Antiun Ingeniería S.L., '
+              'Odoo Community Association (OCA)',
+    'website': 'http://www.antiun.com',
+    'license': 'AGPL-3',
+    'demo': [],
+    'test': [],
+    'installable': True,
+    # 'auto_install':False,
```
pedrobaeza, line 27: Remove these keys and empty ones

pedrobaeza:
Remove website_dynamic_map/i18n/.gitkeep

pedrobaeza:
I'm thinking that this one can go to geo-engine project, and maybe make use of some of its functions. What do you think, @lmignon?

@dgzurita, can you make a nicer icon for this one? I know you are able :wink:

lmignon:
@dgzurita Thank you for this addons. IMO you should depend of `base_geolocalize` to avoid redefining new fields to store the latitude and longitude. Otherwise, If your module is installed at the same time of base_geolocalize, we'll have 2 ways to specify the latitude and the longitude. Another concern is the module name. All your devs are specific to googlemap but the same functionality can be provided by OSM. I'm not in favor of silently promoting google while an OS alternative exist. 
IMO, you should split you work into two modules.
- a first one that extends base_geolocalise and adds 2 new fields: map_marker and map_zoom. (These fields are common to map providers) 
- a second one 'website_dynamic_googlemap' that extends website and implement your functionality by using google.
  This way it will be possible to provide the same functionality by using OSM without redefining the same fields.
  @pedrobaeza IMO, all this work can stay out of OCA/geospatial if the localisation information are stored in the fields defined by base_geolocalize. The same fields are also used by geoengine_base_geolocalize to build the GeoPoint (https://github.com/OCA/geospatial/blob/8.0/geoengine_base_geolocalize/models/res_partner.py#L78)

pedrobaeza:
@lmignon, thanks for your valuable opinion. Making a module for only 2 fields can be excessive, although I understand your rationale. I'm also not in favour of promoting Google Maps against OpenStreeMap, but the reality is that the Google one is for now most advanced.

What do you think, @dgzurita and @rafaelbn?

lmignon:
@pedrobaeza The module for 2 fields is also a way to keep a proper split between backend and website.  A nice enhancement should be to provide the view on the map also in the backend. (Nice to Have)

rafaelbn:
cc @antespi @dgzurita please review thanks

rafaelbn:
@antespi please could you review?

rafaelbn:
@lmignon @pedrobaeza Odoo by default set a Google Maps's map which doesn't "work" (is not useful as it is), this module improve the core functionality of Odoo. So improve means that what Odoo made, now it will work much more better.

rafaelbn:
I propose change name for less confusing. This module improve (I can say FIX) the maps Odoo insert in website_crm, so new name could be:

website_crm_dynamic_google_map

What do you think?

@antespi travis fail, could you review please.

rafaelbn:
Hi @antespi , could you change name ot this module to website_dynamic_google_map ?

@lmignon this is another aproach diferent to https://github.com/OCA/website/pull/62

I think we can go ahead a merge this PR. What do you think?

Thanks

rafaelbn:
Are travis errors related? It looks error are the same as in https://github.com/OCA/website/pull/62

simahawk:
3yrs old, v8... no action... closing.
Feel free to re-open, thanks!

PR #79, opened 2015-09-17
[FIX] website_event_register_free_with_sale: Validate checkout independently for free and paid events

pedrobaeza:
Sharing both free and paid tickets validation method has problems because the same information is not passed over them. Code already handles that free events return as soon as it gets an error, so no special problem with this isolation.

### DIFF
--- a/website_event_register_free_with_sale/controllers/website_sale.py
+++ b/website_event_register_free_with_sale/controllers/website_sale.py
@@ -30,7 +30,7 @@ class WebsiteSale(website_sale):
     # TODO: Not used yet
     optional_free_registration_fields = ["street", "city", "country_id", "zip"]
 
-    def checkout_form_validate(self, data):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
cc @rafaelbn @antespi

rafaelbn:
Tested in customer's production environment :+1:

JavierIniesta:
Tested on runbot.
When I try to select 2 tickets for a paid event, it returns an internal server error.

selection_001

selection_002

pedrobaeza:
That's not a problem on this fix, but a recurring problem with a module in this repo. Please uninstall  website_sale_vat_required in the runbot instance to make the tests.

JavierIniesta:
Tested on runbot. :+1:

dgzurita:
Tested in runbot. :+1:

janetrp:
it works :+1:

PR #84, opened 2015-10-13
Add website logo cobranding based on URI

lasley:
This module adds cobranding on the website based on the URI it is deployed on. This seemed like a good repo to drop it in.

It meets my immediate needs at this point, but I will also be adding some more features as detailed in the ReadMe within the next few months.

Suggestions welcome!

### DIFF
--- a/website_cobrand/README.md
+++ b/website_cobrand/README.md
@@ -1,0 +1,15 @@
+Website Cobranding
+===
+
+This allows for Odoo front-end to be be branded based on the URI it is being accessed on.
+
+If no user is logged in, company context will be discovered based on the URI by trying to find a company with a matching URI.
+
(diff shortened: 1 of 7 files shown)


rafaelbn:
Hi @lasley , as I see is this PR a WIP?

lasley:
@rafaelbn - Yeah honestly I forgot I submitted it, thanks for the ping! I have expanded the cobrand logic to allow for routing to different templates based on the the URI as well - much more useful than just the logo IMO. I'll add this to our next sprint.

lasley:
On second thought looking at the code, it's only 9.0 compatible. I'll close this PR and submit a relevant one against v9 - I don't think this module is useful enough as is to warrant a merge

PR #86, opened 2015-10-15
Add module website_crm_quick_answer.

yajo:
It answers automatically when a request comes from the website contact form. No models, just settings, but automated and with instructions.

cc @rafaelbn


yajo:
Oh sorry, targeting the wrong branch.

PR #87, opened 2015-10-16
Cookie notice translate and simpler controller

yajo:
Newly created companies did not have the messages in the right language.

Existing installations probably will have to change this by hand.

Also simplified some redundant code from the controller.

cc @rafaelbn

### DIFF
--- a/website_cookie_notice/controllers/main.py
+++ b/website_cookie_notice/controllers/main.py
@@ -22,22 +22,6 @@
 from openerp.addons.web import http
 from openerp.http import request
 
-CONFIG_FIELDS = [
-    'cookieAnalytics',
-    'cookieMessage',
-    'cookiePolicyLink',
-    'cookieOverlayEnabled',
-    'cookieAnalyticsMessage',
-    'cookieErrorMessage',
-    'cookieDeclineButton',
-    'cookieAcceptButton',
-    'cookieResetButton',
-    'cookieWhatAreTheyLink',
-    'cookieAcceptButtonText',
-    'cookieDeclineButtonText',
-    'cookieResetButtonText',
-    ]
-
 
 class CookieNotice(http.Controller):
     @http.route(
@@ -51,6 +35,7 @@ class CookieNotice(http.Controller):
         company_id = user_model._get_company(cr, uid, context)
         company = company_model.browse(cr, uid, company_id, context)
         res = {}
-        for field in CONFIG_FIELDS:
-            res[field] = getattr(company, field)
+        for field in dir(company):
+            if field.startswith("cookie"):
+                res[field] = getattr(company, field)
         return res
--- a/website_cookie_notice/models/res_company.py
+++ b/website_cookie_notice/models/res_company.py
@@ -19,57 +19,61 @@
 #
 ##############################################################################
 
-from openerp import models, fields
-from openerp.tools.translate import _
+from openerp import _, models, fields
 
 
 class Company(models.Model):
     _inherit = "res.company"
     cookieAnalytics = fields.Boolean(
-        string="cookieAnalytics",
-        help="just using a simple analytics package? change this to true")
+        string="Analytics",
+        help="Just using a simple analytics package? change this to true")
     cookieMessage = fields.Char(
-        string="cookieMessage",
-        default=_('We use cookies on this website, you can '
-                  '<a href="{{cookiePolicyLink}}" title="read about our '
-                  'cookies">read about them here</a>. To use the website as '
-                  'intended please...'),
+        string="Message",
+        default=lambda self: _(
+            'We use cookies on this website, you can '
+            '<a href="{{cookiePolicyLink}}" title="read about our '
+            'cookies">read about them here</a>. To use the website as '
+            'intended please...'),
         translate=True)
     cookiePolicyLink = fields.Char(
-        string="cookiePolicyLink",
-        help="if applicable, enter the link to your privacy policy here...",
+        string="Policy link",
+        help="If applicable, enter the link to your privacy policy here...",
         default='/page/privacy')
     cookieOverlayEnabled = fields.Boolean(
-        string="cookieOverlayEnabled",
-        help="don't want a discreet toolbar? Fine, set this to true")
+        string="Overlay enabled",
+        help="Don't want a discreet toolbar? Fine, set this to true")
     cookieAnalyticsMessage = fields.Char(
-        string="cookieAnalyticsMessage",
-        default=_('We use cookies, just to track visits to our website, we '
-                  'store no personal details.'),
+        string="Analytics message",
+        default=lambda self: _(
+            'We use cookies, just to track visits to our website, we '
+            'store no personal details.'),
         translate=True)
     cookieErrorMessage = fields.Char(
-        string="cookieErrorMessage",
-        default=_("We\'re sorry, this feature places cookies in your browser "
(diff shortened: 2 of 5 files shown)


website_cookie_notice/controllers/main.py
```diff
@@ -51,6 +35,7 @@ def get_cookie_notice_config(self):
         company_id = user_model._get_company(cr, uid, context)
         company = company_model.browse(cr, uid, company_id, context)
         res = {}
-        for field in CONFIG_FIELDS:
-            res[field] = getattr(company, field)
+        for field in dir(company):
```
pedrobaeza, line 38: Don't use `dir`, but `company._fields`

website_cookie_notice/models/res_company.py
```diff
@@ -19,57 +19,61 @@
...
 ##############################################################################
 
-from openerp import models, fields
-from openerp.tools.translate import _
+from openerp import _, models, fields
 
 
 class Company(models.Model):
     _inherit = "res.company"
     cookieAnalytics = fields.Boolean(
```
pedrobaeza, line 27: field names should be snake_case (this is not your fault), but please change them to this convention and use oldname to have automatically migrated the columns (bump also the version y number, please)

rafaelbn:
Tested in runbot, I have installed Spanish as language and when I change to Spanish the website it still shows Cookies in English

rafaelbn:
Hi @eLBati , we have a problem with module https://github.com/OCA/website/tree/9.0/website_cookie_notice it's not translatable as Odoo translate website. Could you give us a tip to fix this. In our solution in this moment is re-coding module. I guess you could tell us something to fix it in a more easy way? Thanks

rafaelbn:
Sorry I mean the module in v8 : https://github.com/OCA/website/tree/8.0/website_cookie_notice

yajo:
After some more insight, it seems like my fix is useless as it gets hit by the same Odoo ~~bug~~ behavior that the current implementation has: **Odoo does not translate default values.**

New companies created by a user that has selected a translated language get the translated strings, but old companies not. Also, those strings remain in the original language regardless of the user's setting when browsing the website.

The only viable solution I imagine now (apart from fixing Odoo, which will be blocked for 8.0/9.0 probably) is to drop all those settings and do it by views, which are translatable by Odoo by default. After all, having a different setting per company does not have much sense. However, it implies a complete refactoring of this module and I'd like to know the original author and OCA's opinion before starting.

eLBati:
@Yajo @rafaelbn I confirm: odoo does not correctly load fields content translations by module.

Indeed, if you go, after having installed Spanish language, to your company -> edit -> cookie notice -> click on the translation button of a field (say cookieMessage)
translate

you see the English value as 'translated value'.
Nevertheless, if you change the translated value, you are able to translate the field's content and the translation will be displayed by the web site.

About the solution: I think odoo should be fixed. Translatable fields exist and they must be translatable :smirk: even by module.
What do you think?

rafaelbn:
Could reproduce this issue in a Odoo runbot instance?

yajo:
I agree in the sense Odoo should be fixed, but the problem is that I'm pretty sure they will not want to merge it in 8.0, or even 9.0. That's why I was thinking about just using views. After all, Odoo translates views correctly.

rafaelbn:
If this is a bug in Odoo Core, they will fix it. That's why I ask: could we reproduce this in Odoo core?
cc @pedrobaeza

yajo:
We'd have to find a core module that adds default values to a translated field. Any clue?

pedrobaeza:
As the module is writing, you will only get the string in the language of the user that installs the module. For this to work, you have to override default_get, and load by hand the translation on each language in ir.translation model. For working with loading the string on each language, you'll need to make a method trick this way:

```
def default_get(...):
    def _load_string(context=None):
        <store_on_ir_translation>(_('string to translate'))

    for lang in <available_langs>:
        load_string(context={'lang': lang.code)
```

eLBati:
@Yajo i can't find core modules doing exactly this.

But I found a similar example: product.category field name is translatable and `All` value is loaded by XML
https://github.com/odoo/odoo/blob/8.0/addons/product/product_data.xml#L7-L9
and correctly translated by .po
https://github.com/odoo/odoo/blob/8.0/addons/product/i18n/es.po#L774-L777

So, if you want to make a less invasive change, you could load company data by XML. Then odoo should be able to translate them

pedrobaeza:
@eLBati, it's not the same case, because you have an static record. Here, you have a default for each company. Please see my explanations on how to do it

yajo:
I tried what you say, @pedrobaeza, but it does not work because `ir.translation` asks for the record id when you are translating a field, and by definition `default_get` does not have a record.

yajo:
I'm thinking 2 alternatives:
1. Save translation in an `@onchange`, but only the first time (I don't know how to guess that though).
2. Drop all fields. After all, why 13 fields for customizing such a short message? Odoo has the ability to alter views, and you can even do it with submodules if needed. Also those will get translated correctly. 1 `t-call` in main layout should do the trick.

What do you think?

pedrobaeza:
Yeah, you're right! I didn't though about the record ID is not yet available. We can do what you say at qweb template level, but we will lost per-company customization.

yajo:
Is there per-company website, after all?

pedrobaeza:
Not in v8 directly, but via an extra module by Odoo In v9 the support is native, so this is also preparing the migration to v9.

yajo:
Anyway cookies law apply equally to any company, no?

pedrobaeza:
Well, I'm not sure about that: different messages, laws in different countries.... @eLBati, what do you think?

eLBati:
@pedrobaeza @Yajo as said at https://github.com/OCA/website/pull/87#issuecomment-149853199 I would have loaded data via XML. If users create new companies, they will have to translate cookie notice fields if needed, not so onerous.

yajo:
Not very useful to provide default values then...

eLBati:
Default values could be useful for new companies created by users

yajo:
Indeed, but this module will never be _feature complete_ with its current status. Having to translate for each installation is tedious for sysadmins having to install this across lots of instances in different languages.

I'll do the view-based version.

rafaelbn:
This module is OCA and in this moment is not translatable, so this module as it is has a BUG as cannot be used for OCA when website is in 2 o more languages.
We will fix it, I just want to know, all of you @eLBati @Yajo @pedrobaeza are agree with the changes that we must do?

pedrobaeza:
Agree

yajo:
Me too.

eLBati:
On 23 October 2015 at 16:09, Yajo notifications@github.com wrote:

> Indeed, but this module will never be _feature complete_ with its current
> status. Having to translate for each installation is tedious for sysadmins
> having to install this across lots of instances in different languages.
> 
> Actually, with my proposal
> https://github.com/OCA/website/pull/87#issuecomment-149853199 (maybe I
> did not explain ti well?), sysadmins will not have to translate the module
> for each installation. They will have to translate it for every additional
> company after the first.

That being said, I am not against your alternative approach.

yajo:
I understand what you say @eLBati, but even with that being true it's just not enough to make it feature-complete, as seen in https://github.com/OCA/website/pull/87#issuecomment-150230090.

After discussion, I'll start migrating this module to view-based message. Closing this for the time being.

BTW, I think I'll rely on website_legal_page for the long message.

PR #93, opened 2015-10-23
[FIX] Remove unpublished departments from the breadcrumb

lmignon:
to avoid AccessDenied for visitors

### DIFF
--- a/website_hr_department/controllers.py
+++ b/website_hr_department/controllers.py
@@ -28,6 +28,27 @@ from openerp.addons.website_hr.controllers.main import website_hr
 
 class WebsiteHr(website_hr):
 
+    def __get_parent_department(self, department):
+        """Return the parent department. The method perform a search
+        on the parent_id to enforce the security rules since on a manytoone
+        field, the id of the related object is always available on the record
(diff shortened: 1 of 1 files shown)


rafaelbn:
In which URL it's the issue?

lmignon:
@rafaelbn 
Hzre it's the steps to reproduce the bug without the patch:
- Create department DEP1
- Create department DEP11 child of DEP1
  In a website page create a link to the page generated for DEP11 (the url starts with /departments/...)
- Publish the page of the department DEP11
- as an anonymous visitor, try to access to the  page by clikcink on the new link.
  Current behaviour: AccessDenied
  Expected behaviour: The page is displayed and the parent department is not displayed in the breadcrumb since it's not published.

StefanRijnhart:
:+1:

lmignon:
rebased after merge of #97

rafaelbn:
Tested in runbot :+1:

lmignon:
@pedrobaeza @hbrunn I think that this one can be merged.

lmignon:
Thank you @StefanRijnhart

PR #94, opened 2015-10-23
Signup email check

yajo:
This module was written to extend the functionality of public sign up to force
users to provide a valid email address.

To reinforce this requirement, the user will not need to provide a password at
sign up, but when logging in later for the first time.

- Log out.
- Sign up `/web/signup` with a valid email.

CC @rafaelbn @sergio-incaser.


rafaelbn:
@Yajo must fix check failing in runbot first

dreispt:
@Yajo did you find troublesome the harcoded 1 day validity for the signup token?

rafaelbn:
@Yajo after signup Odoo says " Check your email to activate your account! " but there is not any message sent. We expect that Odoo sends an email with the link for confirming it.

sergio-teruel:
@Yajo I test into a private server with a valid email setting. when are you send to user the validation email?

yajo:
Thanks everyone, I'll send some commits but please do not review until I say.

yajo:
It should be ready for prime time. Travis should work too. However, runbot will fail because of new dependency on `validate_email` python package.

yajo:
@dreispt That does not matter.

rafaelbn:
Tested :+1: , please @sergio-incaser @carlos-incaser could you review again.

rafaelbn:
@pedrobaeza @dreispt how to solve this

> It should be ready for prime time. Travis should work too. However, runbot will fail because of new dependency on validate_email python package

StefanRijnhart:
Let's ask @gurneyalex to add the validate_email python library to the runbot configuration.

auth_signup_verify_email/__openerp__.py
```diff
@@ -0,0 +1,25 @@
...
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "external_dependencies": {
+        "python": [
+            "validate_email",
+        ],
+    },
+    "depends": [
+        "auth_signup",
```
StefanRijnhart, line 20: If you don't depend on the _website_ module, then what is your reason to add this module here?

sergio-teruel:
@Yajo @rafaelbn Tested in local server and all ok, for me :+1:

auth_signup_verify_email/__openerp__.py
yajo, line 20: Good question... Where should it go?

yajo, line 20: Ah nevermind, found it: https://github.com/OCA/server-tools. Let me close this and resend there.

yajo:
Closing PR to open in server-tools. Sorry for that.

auth_signup_verify_email/__openerp__.py
StefanRijnhart, line 20: Thanks!

PR #90, opened 2015-10-20
[ADD][8.0] Website portal purchase

sergio-teruel:
This module adds the partner's purchase documents in the frontend portal. Suppliers will be able to see the list and state of their request for quotations, purchase orders and supplier invoices.

PR #101, opened 2015-11-02
Add travis dependency

sergio-teruel:
Add server-tools repository to solve dependency error.


pedrobaeza:
You have to include it in oca_dependencies.txt file, and why don't include it in the PR where is needed?

sergio-teruel:
Changes included in #90

PR #107, opened 2015-11-11
[8.0][website_snippet_contact_form] Contact form in a snippet

yajo:
CC @rafaelbn @antespi.


pedrobaeza:
Better to call it _website_snippet_crm_

website_crm_snippet/__openerp__.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# © 2015 Antiun Ingeniería, S.L.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Contact Form Snippet",
+    "summary": "Put a contact form anywhere you want",
+    "version": "8.0.1.0.0",
+    "category": "Uncategorized",
```
pedrobaeza: Put the same category as Website

pedrobaeza:
For the icon, please follow the same style as https://github.com/odoo/odoo/tree/8.0/addons/website/static/src/img/blocks

rafaelbn:
> Better to call it website_snippet_crm

Odoo core module which add contact form is website_crm, this module adds a snippet, don't you think @pedrobaeza website_crm_snippet is more correct an intuitive?

pedrobaeza:
No, but the standard way to name this is website_snippet__, so that you have all the snippets together sorting alphabetically. It's the same as the widgets for backend: web_widget__.

rafaelbn:
Tested, please:
- You don't show checkbox for accepting privacy policy but form expect it


- Improve a little bit design, make fields not full width



Thanks

pedrobaeza:
You should add the possibility to add the form as a floating element, and maybe this can replace what @rafaelbn says about the size, because the width will adapt to the available space in the upper container.

yajo:
The point of _crm_ in name is to agree with _website_crm_ module name. If I prefix this with _website_snippet_, I'll call it _website_snippet_contact_form_. About the other problems, I'm on it.

pedrobaeza:
OK, I agree with your suggested name.

rafaelbn:
I agree

yajo:
I'm trying to make it able to be resized as ssuggested, but no way! I always get this error in JS console: `have not insert action: data-drop-near or data-drop-in`, but no idea on where it comes from or how to fix! Can you help please?

rafaelbn:
@dgzurita could you review and help please?

yajo:
Still lacking an appropiate icon for the snippet, but that does not seem very critical for version 1.0, right?

rafaelbn:
Where is runbot?

antespi:
:+1: tested in runbot

rafaelbn:
Tested and :+1: great job!

bd231872-6845-11e5-8550-1f071c9e5e5c

dgzurita:
:+1: tested in runbot

pedrobaeza:
What about the possibility of configuring the fields to show?

rafaelbn:
It takes the parameters of the standard contact form. We will see to improve it.

pedrobaeza:
You should mention then on the Known issues section. Although the icon part.

PR #109, opened 2015-11-13
[8.0][website_img_bg_style] Allow to set some CSS classes more easily

yajo:
Please read known issues.

I'd have no problem in pushing website_less module to OCA. Source code contains no license info, but since it's based in 8.0 (AGPL), whatever license it has cannot be non-GPL compliant.

However that can pose some legal issues or opinion flamewars, so for now I chose just to make this module have less options to make a more straight forward PR.

PD: I remember to have downloaded a previous version of website_less that had AGPL headers, if anyone has it, it's a good candidate for OCA, before Odoo SA decides to close all code (which should never happen, but who knows...).


dgzurita:
Tested in runbot :+1:

rafaelbn:
Thanks for this contribution :+1:

rafaelbn:
cc @pedrobaeza

PR #113, opened 2015-11-17
[8.0][website_snippet_big_button] Snippet to quickly add "like" and "don't like" buttons

yajo:
This is the result from this snippet with the default Odoo theme:

captura de pantalla de 2015-11-17 12-21-00

User can customize the URL for each button.

A simple js file is used to make the whole button follow the first link found when clicked.

A class is added to the parent element to make it easy to be styled.


pedrobaeza:
Please rename to singular: _website_snippet_big_button_

yajo:
Ready.

website_snippet_big_button/views/snippets.xml
```diff
@@ -0,0 +1,48 @@
...
+            <section class="oe_snippet_body readable big-buttons">
+                <div class="container">
+                    <div class="row mt16 mb16">
+                        <div class="text-center col-md-6 col-xs-6">
+                            <div class="btn btn-success js_big_button">
+                                <span class="fa fa-thumbs-o-up fa-5x"/>
+                                <h3 class="mt8 mb0">
+                                    <a href="/page/contactus">I want this</a>
+                                </h3>
+                                <p>Request budget</p>
```
hbrunn, line 26: shouldn't this be some more neutral example text?

website_snippet_big_button/views/snippets.xml
```diff
@@ -0,0 +1,48 @@
...
+                        </div>
+                        <div class="text-center col-md-6 col-xs-6">
+                            <div class="btn btn-danger js_big_button">
+                                <span class="fa fa-thumbs-o-down fa-5x"/>
+                                <h3 class="mt8 mb0">
+                                    <a href="/">
+                                        I want another thing
+                                    </a>
+                                </h3>
+                                <p>Continue searching</p>
```
hbrunn, line 37: and this too?

rafaelbn:
I rebuild for testing

website_snippet_big_button/views/snippets.xml
yajo, line 26: Seems good to me (both), but I'm open to suggestions.

rafaelbn:
@yajo do you need this module? If so please answer @hbrunn , thanks,.
@pedrobaeza please take a look.

rafaelbn:
I tested module. :+1:

yajo:
@hbrunn is answered: https://github.com/OCA/website/pull/113#discussion_r45580958

And yes, I need this. Recently I did a SEO training, and this is something that surprised gladly the expert teacher, because reduces bouncing: you provide an alternative to your visitor other than just going back to where they came from.

Besides, if any client wants a different default message, as @hbrunn says, you can always extend this with your custom client's module.

rafaelbn:
@hbrunn is the question answered? I think the text could be more creative but is not relevant, you can quickly change in the editor.

rafaelbn:
Please @pedrobaeza @sergio-teruel @carlosdauden @antespi  review.

lmignon:
:+1: (Code review + Functional test)

PR #115, opened 2015-11-19
Create empty sl.po

sysadminmatmoz:
... to be able to start with the transifex translation


sysadminmatmoz:
Since it's an empty language file, would it be possible to skip the review process?

pedrobaeza:
But this is not needed for translating... Just go to Transifex and translate!

sysadminmatmoz:
At the time I made this commit, the website_slides was not present among the tarnsifex oca website project modules :smile:

PR #116, opened 2015-11-19
[8.0][website_event_register_free_with_sale] Ensure event_id is int.

yajo:
Wrong type was raising the error shown in https://github.com/OCA/event/pull/19.


pedrobaeza:
Curious that the browse method doesn't convert this value, and saves it as an str.

:+1: for this quick patch, although maybe it's convenient to put a PR over Odoo to assure that the value is converted before being stored on the ID field.

yajo:
Not now, but at least there is a new issue.

yajo:
cc @rafaelbn @JavierIniesta.

JavierIniesta:
:+1:  as a quick fix.

rafaelbn:
Already answered by Odoo here https://github.com/odoo/odoo/issues/9650#issuecomment-158316415

PR #117, opened 2015-11-24
[FIX] Allowing all users to read filter

antespi:
When creating filter to be associated with email action, then it is created with `user_id = Administrator` and then other users can not read it.

When a user creates a lead or an opportunity then action is evaluated and filter read, then an Access Error is raised because user can not read Administrator filters.

If `user_id = False`, then all users can read it and no Access Error will appear.

### DIFF
--- a/website_crm_quick_answer/data/automation.xml
+++ b/website_crm_quick_answer/data/automation.xml
@@ -4,6 +4,7 @@
 
 <record id="website_form_leads_filter" model="ir.filters">
     <field name="name">Coming from the website contact form</field>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
:+1:

antespi:
Please @Yajo, review it when you can.

rafaelbn:
:+1: this is a solution

hbrunn:
:+1:

rafaelbn:
@Yajo take care that this has been merged for improvements, thanks

yajo:
Great, thanks.

PR #118, opened 2015-11-25
[8.0][website_snippet_anchor] Allow to set anchors to your web pages

yajo:
See README for usage instructions.

CC @rafaelbn @antespi.


website_snippet_anchor/README.rst
```diff
@@ -0,0 +1,80 @@
...
+
+To link to any page's anchor, you need to:
+
+* Edit any web page.
+* Select some text.
+* Press *Link* in the editor toolbar.
+* Choose a page in the opened dialog.
+* Set an anchor there too.
+* Press *Save*.
+
```
antespi, line 36: Please add an example of menu item with anchor

antespi:
After testing in runbot I found these issues:
1. Anchor not saved in block if no content has been changed. This is a normal operation, user first add contents and save, then edit again to set anchors to text blocks without adding new content.
2. When creating a link, if user only set anchor field, he can not save because no page or url defined. Maybe it s better without adding anchor field and user set anchor in URL field like `#myanchorname`. This is exactly what happend at menu item edit dialog.

website_snippet_anchor/__openerp__.py
```diff
@@ -0,0 +1,25 @@
...
+    "name": "Set Snippet's Anchor",
+    "summary": "Allow to reach a concrete section in the page",
+    "version": "8.0.1.0.0",
+    "category": "Website",
+    "website": "http://www.antiun.com",
+    "author": "Antiun Ingeniería S.L., Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "external_dependencies": {
```
elicoidal: Hi @Yajo 
Thanks for your contribution
I will check mostly format and guidelines in your PR
Can you remove the empty keys?

website_snippet_anchor/README.rst
```diff
@@ -0,0 +1,80 @@
...
+to any URL and get directly to the element with that anchor, in the form of
+http://www.antiun.com/#anchor-name.
+
+Usage
+=====
+
+To add an anchor to any element, you need to:
+
+* Edit any web page.
+* Go to *Insert blocks > Structure*.
```
elicoidal, line 23: Does it work only with "structure" blocks?
If so maybe state it somewhere.

rafaelbn:
Please @dgzurita review. This is a great functionality to make ONEPAGE Websites with Odoo

website_snippet_anchor/README.rst
yajo, line 23: First paragraph says "to any section of your web page". I would not consider a section something that is not inside structure menu.

website_snippet_anchor/__openerp__.py
yajo: np

yajo:
@antespi 

> Anchor not saved in block if no content has been changed. This is a normal operation, user first add contents and save, then edit again to set anchors to text blocks without adding new content.

This seems Odoo bug, I'll investigate more on it.

> When creating a link, if user only set anchor field, he can not save because no page or url defined.

This should be fixed now.

dgzurita:
Tested in runbot:
1. When adding an anchor for a "Banner" block with several slides (thats a Bootstrap Carousel) the arrows lose their functionality because they lost their reference.
2. It would be perfect if you can add a bit of jQuery to animate the scroll and make it more user friendly. That way you improves the feel of navigating the same page. Something like this https://css-tricks.com/snippets/jquery/smooth-scrolling/ 
3. For large OnePage views i think it's a good idea to add a "back to top" anchor link also fixed at the right bottom of the page.

Thanks @Yajo

yajo:
@dgzurita:

>   When adding an anchor for a "Banner" block with several slides (thats a Bootstrap Carousel) the arrows lose their functionality because they lost their reference.

Thanks! fixed.

>    It would be perfect if you can add a bit of jQuery to animate the scroll and make it more user friendly. That way you improves the feel of navigating the same page. Something like this https://css-tricks.com/snippets/jquery/smooth-scrolling/

Seems quite nice, and was one of the first things I thought about, but we decided to leave that for future versions of the module.

>    For large OnePage views i think it's a good idea to add a "back to top" anchor link also fixed at the right bottom of the page.

Hmm I think this is more related to the theme itself. Maybe other themes prefer to add a sticky side menu (like in [Odoo docs](https://www.odoo.com/documentation/8.0/reference/orm.html)), or others prefer the link at another place. What do you think @rafaelbn?

@antespi:

> Anchor not saved in block if no content has been changed. This is a normal operation, user first add contents and save, then edit again to set anchors to text blocks without adding new content.

This should be fixed. It works locally, let's see in runbot.

antespi:
:+1: Cool @Yajo works very well. Thanks

dgzurita:
:+1: Thanks @Yajo

rafaelbn:
Tested in runbot! great thanks! :+1:

PR #121, opened 2015-12-02
[8.0][website_marginless_gallery] New snippet for image galleries

yajo:
It allows to add a snippet of photos with floating text like this:

captura de pantalla de 2015-12-02 09-04-56

Mostly unstlyed, to let themes do their job.


yajo:
Do not trust Coveralls.

dgzurita:
:+1: Thanks @Yajo . Works fine.

I suggest to improve it using a Masonry Block kind like in https://theme-graphene.odoo.com/page/blocks. It allows you to make the perfect square blocks (with responsive height) and combine with 2x height and/or 2x width using this plugin http://desandro.github.io/masonry/

Just an idea. Maybe for a different snippet

yajo:
@dgzurita Definitely a great idea for this snippet, but I think it could be achieved in a later version.

pedrobaeza:
Please write down the suggestion in the "Known issues / Roadmap" section

lfreeke:
:+1: (test)

yajo:
Seems like I need only one more approval. Any charitable soul out there?

pedrobaeza:
I have tested, but I'm not able to change mosaic piece height. I click on "Customize > Change Images Height", but it remains the same. It also fails with the initial provided value.

Please fix it.

yajo:
@pedrobaeza Can you provide debug info please, it works for me on runbot, and the UI test works too:

bloggif_567938f2b1961

pedrobaeza:
I tried with Firefox. Are your video made with Chrome? Maybe a blocker like NoScript can screw up the thing?

yajo:
No, I use Firefox. Probably NoScript will make it fail, since it needs js. Try disabling addons please.

pedrobaeza:
It should be then.

:+1:

PR #134, opened 2015-12-23
[8.0][website_slides][FIX] Use compiled CSS and fix fullscreen PDF proportion

yajo:
Odoo 9.0 autocompiles `.less` files, but v8 does not, so we have to serve the precompiled CSS.

Also there was a bug with fullscreen PDF files, which were showing up like this:

captura de pantalla de 2015-12-23 11-26-44

With the backported fix, they look like this:

captura de pantalla de 2015-12-23 11-27-14


pedrobaeza:
Thanks for the fix

:+1:

rafaelbn:
:+1:

PR #136, opened 2015-12-28
[8.0][IMP][website_snippet_marginless_gallery] Allow to avoid trimming images.

yajo:
The module by default trims images to show no spaces among them, but this allows the user to show the full images as shown here:

bloggif_56812553bf955


yajo:
Just to make reviewers understand the purpose behind this patch, it's very useful for displaying equally-sized logos.

Without the patch, when using the marginless gallery snippet for that, it displays like this:

captura el 2016-02-23 a las 12 11 11

With the patch, it is useful to make logos with different sizes and proportions look alike:

captura el 2016-02-23 a las 12 10 59

pedrobaeza:
:+1:

oscarolar:
:+1:

PR #129, opened 2015-12-10
8.0 website blog types and snippet

gfcapalbo:
Extension and website snippet for blog posts.
           **website_blog_more_post_types**     adds the possibility of viewing blog posts in the list in three ways: only title (default) , title + teaser or full blog post in list.  The blog post types can be mixed in the same blog.
           **website_snippet_blog_display_post** : a snippet to visualize a blog post on your website.

PR #141, opened 2016-01-07
[8.0] Improve module website_breadcrumb

JonathanNEMRY:
This PR provides a new snippet that allows user to add the breadcrumb where you want into the page
This also provides some improvements and fixes with the two differents criteria: "revert" & "include-self"


website_breadcrumb/views/templates.xml
```diff
@@ -6,6 +6,13 @@
 <data>
 
 <template id="breadcrumb" name="Breadcrumb snippet">
+    <t t-call="website_breadcrumb.breadcrumb-generator">
+        <t t-set='include_self' t-value='False'/>
+        <t t-set='revert_order' t-value='True'/>
+    </t>
+</template>
+
+<template id="breadcrumb-generator" name="Breadcrumb snippet generator">
```
yajo: The usual standard is to use `_` instead of `-` in ids.

Also I think that with this change, _website_breadcrumb_ should depend on _website_snippet_breadcrumb_ and not the other way around. If you want automatic breadcrumbs in the top of your pages, you will not care about the ability to put them manually or not; but if you want to put them manually, you will be disturbed by automatic breadcrumbs appearing in many pages.

website_node_branded/README.rst
```diff
@@ -0,0 +1,52 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+====================
+website_node_branded
+====================
+
+Once installed, you can use the t-ignore-branding attribute on a <t> node to control if node is branded or qweb-active (bears a
+@data-oe-model or a @t-* *which is not t-field* as t-field does not section out views)
```
yajo, line 10: I'm not sure if it is my poor English or my lack of knowledge of QWeb's internals, but I don't understand what is this module doing. What's the branding stuff you are ignoring?

A more elaborated description with an example of a template and what should happen with that `t-ignore-branding` vs what would happen without it would help a lot.

website_snippet_breadcrumb/static/src/js/website_snippet_breadcrumb.editor.js
```diff
@@ -0,0 +1,80 @@
+(function () {
```
Yajo: This should have the AGPL header.

website_snippet_breadcrumb/__openerp__.py
```diff
@@ -0,0 +1,19 @@
...
+# © 2015 ACSONE SA/NV (<http://acsone.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    'name': "Website Snippet Breadcrumb",
+    'summary': """Website Snippet Breadcrumb""",
+    'author': "Acsone SA/NV",
+    'website': "http://www.acsone.eu",
+    'category': 'Website',
+    'version': '8.0.1.0.0',
+    'depends': [
```
yajo, line 12: Missing license tag.

website_snippet_breadcrumb/views/snippets.xml
```diff
@@ -0,0 +1,40 @@
...
+                    <div class="oe_snippet_thumbnail">
+                        <img src="/website_snippet_breadcrumb/static/src/img/snippets/s_breadcrumb.png"
+                             class="oe_snippet_thumbnail_img"/>
+                        <span class="oe_snippet_thumbnail_title">
+                            Breadcrumb
+                        </span>
+                    </div>
+                    <section class="oe_snippet_body website_snippet_breadcrumb" data-revert-order='true' data-include-self='false'>
+                        <div class="parametricTemplate">
+                          <p>Dynamic Breadcrumb: The content will be replaced after saving the page</p>
```
yajo: I think a `ol.breadcrumb>li` placeholder would be more user friendly than a normal paragraph, to have a better preview of how will it look like.

yajo:
Tested in runbot, and it works pretty fine. :blush:

website_breadcrumb/views/templates.xml
JonathanNEMRY: Hi @Yajo 
IMO the dependency order is correct like that too because `website_breadcrumb` provides the base and `website_snippet_breadcrumb` provides an added value. I suggest to overwrite the template into the `website_snippet_breadcrumb` in order to disable it by default. By this way we switch into manual mode and only care about the dynamic snippet.
This could be a quick solution what do you think?

yajo: Adding a module that depends on another and disables what it does makes less sense than adding a module that adds something upon the previous one. That's why for me it seems more logical if the `website_snippet_breadcrumb` adds the template and logic (it should have the python part too), and then `website_breadcrumb` simply puts that on every page automatically.

If you still prefer to use manual mode, then please implement it in `website_breadcrumb` directly, to make usage consistent no matter if `website_snippet_breadcrumb` is installed. It should be enabled by default, since that's the original purpose of the module. Remember to update its README too.

lmignon: @Yajo IMO we can find pros and cons for the 2 POV. :+1: with the proposal of @JonathanNEMRY to disable the template of `website_breadcrumb` when we install `website_snippet_breadcrumb`. This proposal seems to be a good compromise to take care of your concern.

website_node_branded/README.rst
lmignon, line 10: @Yajo In Odoo if your snippet insert a `<t>`  element, the element will be removed when you save the page an not interpreted by the QWeb processor when the page is rendered. This technical module allows you to bypass this limitation. The breadcrum snippet is basic and only inserts a `<t t-call=...` element into the page https://github.com/acsone/website/blob/website_snippet_breadcrumb-jne/website_snippet_breadcrumb/static/src/js/website_snippet_breadcrumb.editor.js#L37
The credits of this Idea comes from Therp BV (https://github.com/OCA/website/pull/129)

yajo, line 10: If so, then why the snippet needs to inject it by js? Can't it be done directly? It's not going to be removed after all...

lmignon, line 10: @Yajo In this case, the '<t t-call' element  is injected by js because it's possible to configure the behavior of the snippet.

yajo, line 10: Ah OK. Please explain all of this, preferably with an example, in the README.

paolovalier:
tested on runbot, seems not working on blog pages but it's something great

paolovalier:
sorry, my mistake: it's all about configuring the website menus as stated in the README
my :+1:

website_node_branded/__openerp__.py
```diff
@@ -0,0 +1,18 @@
...
+{
+    'name': "website_node_branded",
+
+    'summary': """
+        Technical module providing a way to mark a node in a template as not
+        banded""",
+    'author': 'ACSONE SA/NV,'
+              'Odoo Community Association (OCA)',
+    'website': "http://acsone.eu",
+    'category': 'Uncategorized',
```
pedrobaeza, line 13: Please put a category here. You can use the same as the module website.

website_snippet_breadcrumb/README.rst
```diff
@@ -0,0 +1,69 @@
...
+ * save then you can see your breadcrumb
+
+Note that the breadcrumb is based on the menu hierarchy: Only pages referenced
+into a menu will trigger the computation of the breadcrumb
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/162/8.0
+
+.. repo_id is available in https://github.com/OCA/maintainer-tools/blob/master/tools/repos_with_ids.txt
```
pedrobaeza, line 35: Remove this paragraph

website_snippet_breadcrumb/README.rst
```diff
@@ -0,0 +1,69 @@
...
+.. branch is "8.0" for example
+
+For further information, please visit:
```
pedrobaeza, line 38: Remove this paragraph

pedrobaeza:
The breadcrumb icon will be deformed in modules views, so please provide an square image.

website_node_branded/README.rst
gfcapalbo, line 10: @lmignon  Wouldn't it be nice to split   website_node_branded  in it's own PR?  I have some   documentation to contribute i think it would be cleaner./

lmignon:
@JonathanNEMRY can you rebase your work on https://github.com/OCA/website/pull/236 and takes into account the latest remarks? Thanks in advance.

pedrobaeza:
@JonathanNEMRY  Are you going to work on last comments?

simahawk:
2yrs old, no answer, closing.
@JonathanNEMRY feel free to re-open if needed, thanks!

PR #143, opened 2016-01-11
[Fix] Use the URL field of the slide

jlaloux:
When you are over HTTPS and you want to load a slide. Chrome stops it because it tries to load the slide without HTTP. By using the field website_url coming from the slide model, if the URL of the slide is with HTTPS, you'll not have this error anymore.

image


sergio-teruel:
@jlaloux All your web is under https?? or only some sections?

sergio-teruel:
@jlaloux Please rebase to fix runbot errors

yajo:
I think you are getting hit by a core bug fixed in https://github.com/odoo/odoo/pull/10363/. Could you try please?

pedrobaeza:
No answer for so long, so closing.

PR #148, opened 2016-01-24
[ADD] website_product_share

andhit-r:
This module will add functionality to enable/disabled share button on each product page. This module will be useful if user would like to have share button on all product page. Without this module share button have to be added manually via share button snippet one by one


pedrobaeza:
Thanks for the contribution, Andithia.

As this is e-commerce stuff, please make the PR against e-commerce repo.

andhit-r:
Will do it, thanks @pedrobaeza

PR #149, opened 2016-01-24
[ADD] website_blog_share

andhit-r:
This module will add functionality to enable/disabled share button on each blog post. This module will be useful if user would like to have share button on all blog post. Without this module share button have to be added manually via share button snippet one by one.

Share button will appear after blog subtitle


andhit-r:
Closed PR to restart travis

andhit-r:
Restrat travis + runbot

andhit-r:
@pedrobaeza @elicoidal this is similar to https://github.com/OCA/e-commerce/pull/98, only this add share button to blog post

PR #162, opened 2016-02-06
[ADD][8.0] website_event_share

andhit-r:
This module will add functionality to enable/disabled share button on each event details. This module will be useful if user would like to have share button on all event details. Without this module share button have to be added manually via share button snippet one by one.

Share button will appear after event location


pedrobaeza:
Please put this one in OCA/event

andhit-r:
Moved to https://github.com/OCA/event/pull/34

Issue #164, opened 2016-02-15
[8.0] [website_portal] Login redirection regression


LeartS:
Commit https://github.com/OCA/website/commit/9141fe1b5166f32e10c87abdf120cebf5b190f1b introduced a regression. It absolutely should not redirect to `/` in every case, in particular it overrides redirects defined in the url, breaking for example login and signup from checkout page or every possible customization of the redirect from frontend modules/template modifications with the HTML editor.

pedrobaeza:
cc @sergio-incaser

LeartS:
Actually just noticed it was introduced in https://github.com/OCA/website/commit/5afd318fad7c2ddfb7e0fff096aac9a236d0f4cd cc @carlos-incaser

sergio-teruel:
@LeartS @pedrobaeza Yes your are right, I'll fix it...

PR #165, opened 2016-02-15
[FIX] website_portal: Respect redirect into url.

sergio-teruel:
This fix the #164 issue.
Please @LeartS can you review?
cc @pedrobaeza

### DIFF
--- a/website_portal/controllers/main.py
+++ b/website_portal/controllers/main.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+from urlparse import urlparse, parse_qs
+from urllib import urlencode
 from openerp import http
 from openerp.http import request
 from openerp import tools
@@ -99,12 +101,14 @@ class WebsiteAccount(http.Controller):
 class AuthSignup(AuthSignupHome):
 
     def redirect(self, res, **kw):
-        if request.session.uid:
-            user = request.env['res.users'].browse(request.session.uid)
-            redirect = kw.get('redirect', '')
-            if not user.has_group('base.group_user') and\
-                    (redirect == '' or redirect[-5:] == '/web?'):
-                    return http.redirect_with_hash('/')
+        redirect = kw.get('redirect', '')
+        if request.session.uid and (not redirect or '/web?' in redirect):
+            params = parse_qs(urlparse(redirect).query, keep_blank_values=True)
+            return_url = params.pop('redirect', ['/'])[0]
+            if '/web?' in return_url:
+                return_url = '/'
+            return_url = '%s?%s' % (return_url, urlencode(params))
+            return http.redirect_with_hash(return_url)
         return res
 
     @http.route(website=True, auth="public")
--- a/website_portal/controllers/main.py
+++ b/website_portal/controllers/main.py
@@ -98,21 +98,21 @@ class WebsiteAccount(http.Controller):
 
 class AuthSignup(AuthSignupHome):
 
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Seems OK, but I'm not expert in this. Let's see what @LeartS says.

website_portal/controllers/main.py
```diff
@@ -98,21 +98,21 @@ def details_form_validate(self, data):
...
-    def web_login(self, *args, **kw):
-        res = super(AuthSignup, self).web_login(*args, **kw)
+    def redirect(self, res, **kw):
         if request.session.uid:
             user = request.env['res.users'].browse(request.session.uid)
-            if not user.has_group('base.group_user'):
-                return http.redirect_with_hash('/')
+            redirect = kw.get('redirect', '')
+            if not user.has_group('base.group_user') and\
+                    (redirect == '' or redirect[-5:] == '/web?'):
```
hbrunn: i believe you mean `redirect[:5]` here?

LeartS:
I don't know. What if a user wants to redirect to `/web` or doesn't want the module to redirect to `/` ? or what it the default changes from `/web?` to something else, then this will stop working. Actually it could not work in some cases already, as the default redirection also has the query string, so for example it could be that it doesn't work with debug enabled (`redirect` would be `/web?debug=1`).
IMO the module should not explicitly redirect at all; i.e. it should not call `http.redirect_with_hash`.
It should just change the default redirect value from `/web?` to `/` (actually I would redirect to `/my/home` by default) when the user is a share user.

The problem is that the default redirect is set in the qwebcontext value in the login form, before we know what user has logged in. So, in my opinion, we could either:
- Always make the default redirect to `/` or `/my/home`, even for internal users: there will be the website black top bar to access the backend client!
- Leave all `web_login` as it is and inherit instead `web_client`: there the user is authenticated and we can check if he is a portal/share user, and in that case redirect him to frontend

The first option would simply be something like this:

```
from openerp.addons.web.controllers.main import Home


class WebsitePortalHome(Home):

    @http.route()
    def web_login(self, redirect=None, **kw):
        if not redirect:
                redirect = '/my/home' + request.httprequest.query_string
        return super(WebsitePortalHome, self).web_login(redirect=redirect, **kw)
```

This just sets the default redirect value passed to the standard login method, it can be made inhertiable and customizable (for example taking out the logic that determines the path in a function) and doesn't overwrite the standard `web_login` logic by calling explicitly a redirect, it just changes the default value.

website_portal/controllers/main.py
sergio-teruel: No, i only want five last positions. Only redirect to '/' in case that url is 'base.url/web?'.

LeartS: What if there is a querystring? Wouldn't `redirect` be for example `/web?debug=#` ?

hbrunn: that's what i thought too. I think to do this right, we need urlparse or the like

sergio-teruel:
Respect redirec to '/' or 'my/home'... i think that in v9 is redirected to '/' so I would respect that decision.
Now I redirect with the params in url, so if the url contains '?debug=1', this parameter is passed into returned url. Any direction that comes in the parameter 'returned' is returned.

LeartS:
I still think the module should not redirect at all but just change the default value and let the original function handle it.
But this can be an eventual future improvement, in the meantime :+1:

hbrunn:
same here :+1:

PR #166, opened 2016-02-17
[ADD] Adding new module website_membership_contact_info

mikevhe18:


pedrobaeza:
The name should reflect the real feature of the module. Something like _website_membership_contact_visibility_.

I also prefer to not showing the contact part instead of showing "This information is private", but maybe you can add a selection for choosing it: Visible in Website > "Yes", "No", "Mark as private".

mikevhe18:
@pedrobaeza thx for the review. I'll change the modul name soon.

Actually, I just wanted to tell people if anyone who wants to keep track of the contact information of one member, can contact the administrator. So that's why i put the message "Contact information has been marked as private. For more information please contact administrator" on the website

pedrobaeza:
Yeah, but you know this world is very diplomatic sometimes, so there are some people that can get angry if they think "I cannot see this information because they have marked it as private". It's the same as in Facebook, your contacts doesn't know what you have hidden. They only see what they not. That's why I suggest to make this selection field.

If you can't invest more time, I understand. Just put this comment on Known issues / Roadmap for anyone that wants to improve this.

PR #168, opened 2016-02-18
[9.0] Remove website_portal which was a backport

guewen:
Now it conflicts with the core `website_portal`.


pedrobaeza:
:+1:

pedrobaeza:
I have to mention that there are some improvements made on the v8 website_portal, but they must be added in v9 via new extension modules.

rafaelbn:
I think @sergio-incaser @carlos-incaser and @yajo must review this PR

guewen:
> I have to mention that there are some improvements made on the v8 website_portal, but they must be added in v9 via new extension modules.

I agree.

We should treat backports like the OCB branch:
- For bugfixing, they should be proposed upstream
- For improvements, they should be added in a separate module so we don't lose them on the next version

Also, I should note that the controller has been renamed to `WebsiteAccount` here and it is `website_account` in the official 9.0 branch, so the addons are not swappable. This is how I discovered the presence of this addon here, when I got this error:

```
  File "/home/gbaconnier/.../parts/enterprise/website_contract/controllers/main.py", line 10, in <module>
    from openerp.addons.website_portal.controllers.main import website_account
```

pedrobaeza:
Yeah, but that changes were done because they are too tight in the core, and a total rewrite was needed. We have already made the same with _account_bank_statement_import_. It's not ideal, but it's the way to not be constrained by Odoo rules also in a version totally controlled by OCA. Now, each version goes its way...

yajo:
:+1:

yajo:
I'd suggest you to extract a diff between core and this and consider what should be encapsulated in a new module, what should go upstream, and what should be discarded. But anyway this must be removed.

sergio-teruel:
:+1:  We will review the differences and if is needed we will do a new module..

PR #169, opened 2016-02-24
[8.0][FIX][website_cookie_notice] Cookie message translation

antespi:
Remove enters and spaces to avoid problems with translation of cookie message

PR #170, opened 2016-02-24
[8.0][FIX][website_cookie_notice] Cookie message translation (2)

antespi:
This PR is related with #169 
The problem was not only spaces, also translation definition of res_id was not correct:

```
#: view:website:website.layout
```

This must be

```
#: view:website:website_cookie_notice.message
```

It seems that this problem was caused by @oca-transbot here https://github.com/Antiun/website/commit/550d148f19aeb689789e7fabc94bdd9ba9049931#diff-b4e6e0424c922617ebb787e6c6464dd3

We should review how Transifex generates .po files

### DIFF
--- a/website_cookie_notice/i18n/el_GR.po
+++ b/website_cookie_notice/i18n/el_GR.po
@@ -19,16 +19,16 @@ msgstr ""
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
 #. module: website_cookie_notice
-#: view:website:website.layout
+#: view:website:website_cookie_notice.message
 msgid ". To disable them, configure your browser properly. If you keep using this website, you are accepting those."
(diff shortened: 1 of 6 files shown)


antespi:
Well, checking this issue with @pedrobaeza we saw that Transifex comments are not overwritten when loading translations from repo.
So we have re-load .pot again in Transifex, in order to not reproduce this error again

pedrobaeza:
Solved partially in the commit https://github.com/OCA/maintainer-quality-tools/commit/43ae2fa87b3397aa6fab0e067c6fa824b12a76c6. The solved part is about the template updating.

I had to make a direct commit because it was the only way to try without having to build an entire test repo.

Now trying to see the best way to update translations files.

pedrobaeza:
OK, next step also checked. The translation files upload is working as expected, because if you empty any of the translation strings, Transifex respects the translation stored on the server, so no possible desynchronization. The reason why the language hasn't been updated in this PR is because the template was not updated, and thus there was no occurrence for that string.

I merge this PR anyway for having this instantaneously and not having to wait to the synchronization.

antespi:
Thanks @pedrobaeza

PR #172, opened 2016-03-01
[FIX] website_portal_sale: Order list indicates "Invoiced" state when…

eantones:
… order is "In progress" and not necessarily invoiced yet.

### DIFF
--- a/website_portal_sale/templates/website_portal_sale.xml
+++ b/website_portal_sale/templates/website_portal_sale.xml
@@ -70,7 +70,7 @@
                             <td><span t-field="order.date_confirm"/></td>
                             <td>
                                 <t t-if="order.state == 'progress'">
-                                    <span class="label label-info"><i class="fa fa-fw fa-clock-o"/> Invoiced</span>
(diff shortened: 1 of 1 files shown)


pedrobaeza:
That was backported from v9, so maybe there's a reason for that label. @sergio-incaser @carlos-incaser, do you know it?

sergio-teruel:
In odoo  v9.0 this state is marked as invoiced:
https://github.com/OCA/OCB/blob/9.0/addons/website_portal_sale/views/templates.xml#L80
However i think that @eantones has the reason, but @pedrobaeza i think that this PR also must be proposed to Odoo/Ocb 9.0.
@eantones can you propose to Odoo?  Thanks
For this PR :+1:

pedrobaeza:
OK then for this :+1:

And for the patch to Odoo, they don't usually accept any translation change on stable branch, but you can try.

carlosdauden:
:+1:

eantones:
Ok, thanks!!

I'll try to propose it in odoo/ocb 9.0 or odoo/odoo 9.0 directly.

Thanks again

PR #179, opened 2016-03-11
[IMP] website_portal_sale: add the discount on the order line

eantones:
The discount is shown on the order line if the current user is a member of sale.group_discount_per_so_line group


pedrobaeza:
Same technique as redsys one, so :+1:

eantones:
:+1:

antespi:
:+1:

PR #181, opened 2016-03-15
[8.0][FIX] Update debug info.

yajo:
It was outdated. Does not affect performance but debugging.


pedrobaeza:
:+1: Fast-tracking as obvious

Issue #185, opened 2016-03-31
[8.0] [website_calendar_snippet] DST error


tarteo:
The start and end date are not calculated with DST.

pedrobaeza:
I close this issue as there is a PR opened, and we can comment there possible problems,

PR #186, opened 2016-03-31
[8.0][website_calendar_snippet] DST error fix

tarteo:
Fixes the problem explained in #185.

### DIFF
--- a/website_calendar_snippet/static/src/js/website_calendar_block.js
+++ b/website_calendar_snippet/static/src/js/website_calendar_block.js
@@ -45,12 +45,12 @@
                                 for(var i = 0; i < events.length; i++) {
                                     events[i].start = Date.parseExact(events[i].start, "yyyy-MM-dd HH:mm:ss");
                                     events[i].end = Date.parseExact(events[i].end, "yyyy-MM-dd HH:mm:ss");
-									var start_offset = events[i].start.getTimezoneOffset();
-									var end_offset = events[i].end.getTimezoneOffset();
-									start_offset = start_offset - (start_offset * 2);
-									end_offset = end_offset - (end_offset * 2);
-									events[i].start.addMinutes(start_offset);
-									events[i].end.addMinutes(end_offset);
+                                    var start_offset = events[i].start.getTimezoneOffset();
+                                    var end_offset = events[i].end.getTimezoneOffset();
+                                    start_offset = start_offset - (start_offset * 2);
+                                    end_offset = end_offset - (end_offset * 2);
+                                    events[i].start.addMinutes(start_offset);
+                                    events[i].end.addMinutes(end_offset);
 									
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Can you point me to an example that I can try on runbot so I can see before and after?

rafaelbn:
@tarteo could you clarify this PR? Thanks

tarteo:
@rafaelbn The current module does not take DST (https://en.wikipedia.org/wiki/Daylight_saving_time) into account. This PR fixes that issue.

lasley:
@tarteo - I'm not so sure if this adds up because the delta is created in UTC from Time A and Time B. These UTC times were generated via Tz aware context Datetimes on the way from browser to Odoo, which would account for Time A being in DST and Time B being outside of it.

I can, however, see a theoretical issue if User A is in a timezone that does not have DST. They schedule an appointment for User B that is in a DST tz. Looking at the calendar_event code, I believe that is an edge case that would yield a perceived global DST issue.

Can you please provide a before/after example? 

P.S. - I hold an eternal hate for whoever decided DST was a good idea.

tarteo:
@lasley

>  I'm not so sure if this adds up because the delta is created in UTC from Time A and Time B. These UTC times were generated via Tz aware context Datetimes on the way from browser to Odoo, which would account for Time A being in DST and Time B being outside of it.

When the calendar event (or any object with a date) is created it converts the tz aware date to tz unaware date (database). When shown in e.g. a form the tz unaware date gets client side converted to tz aware date (correct me if I'm wrong)

> I can, however, see a theoretical issue if User A is in a timezone that does not have DST. They schedule an appointment for User B that is in a DST tz. Looking at the calendar_event code, I believe that is an edge case that would yield a perceived global DST issue.

The required offset is calculated client side, so I think this is not an issue (maybe I misunderstood your issue).

Before / after:

dstfix

http://www.timeanddate.com/time/change/netherlands/amsterdam

lasley:
@tarteo TBH I see this as a core calendar issue due to lack of true time zone support. The calendar seems to be fully naive in even the most simple instances, such as:
- Admin User in `US/Hawaii` creates appointment for 11:00 
- Admin adds Demo User as attendee
- Demo User in `US/Pacific` sees 11:00

tarteo:
@lasley, dst is applied on the web_calendar module, e.g. I make an appointment for 11:00; on the core calendar it shows the correct time, but on the website_calendar_snippet it shows 10:00 (depending on the dst). The module (website_calendar_snippet) is already applying the user's time zone.

lasley:
@tarteo - Whoops seems I missed this comment. 

Got it, I understand now. Your fix does work, and makes the times correlate w/ what you would see in the backend calendar.

This is still technically a datetime naive calendar, but so is the real one - which is an issue for another day/PR. 👍  from me

ehahouimohssine:
:+1:

PR #106, opened 2015-11-11
Porting  website_cookie_notice and website_legal_page to 9.0

eLBati:
see commits

PR #193, opened 2016-04-20
Change link to protocol relative wikipedia in website_cookie_notice

### LINKED ISSUE #144: [website_cookie_notice] insecure content after click on cookie notice
Hi,

My odoo server runs behind nginx. All traffic redirect to https. With website_cookie_notice installed, after visitor click on cookie notice insecure content warning always shows. Tried to uninstall website_cookie_notice, no insecure content warning shows.

lasley:
This fixes #144 by switching the http://www.allaboutcookies.org/ link to Wikipedia on cookies instead. 

Original link does not provide an SSL version, and also includes spammy/non-professional items such as singles ads. I was not able to find a source more complete than Wikipedia on this topic.


lasley:
Also I wasn't quite sure on protocol regarding the headers in files that I am not editing as part of the commit. I went ahead and changed the header in the file I edited to new OCA, please let me know if I need to do on all files.

website_cookie_notice/models/res_company.py
```diff
@@ -62,7 +46,7 @@ class Company(models.Model):
         string="cookieResetButton")
     cookieWhatAreTheyLink = fields.Char(
         string="cookieWhatAreTheyLink",
-        default="http://www.allaboutcookies.org/")
+        default="//wikipedia.org/wiki/HTTP_cookie")
```
pedrobaeza, line 49: Why are you using partial URL (without protocol)?

pedrobaeza:
Shorten headers in always a plus!

website_cookie_notice/models/res_company.py
lasley, line 49: This is called a protocol relative URI, which will adapt to the protocol you are viewing the site on. I didn't want to force HTTPS on non-HTTPS sites, and adding `http:` will yield the certificate error that was reported in the issue

lasley:
Cool shortening now. Also, I accidentally based and submitted on 9.0. I'll consider this the point of no return, and will just upgrade to 9.0 under this PR + provide the backport for 8.0 in another PR. My bad!

eLBati:
Thank you @lasley , did you check PR https://github.com/OCA/website/pull/106 before?

lasley:
Hah nope, good catch @eLBati - that definitely supersedes this and one I just submitted for legal. Will review on that instead, thanks!

eLBati:
@lasley I still have to address last comment https://github.com/OCA/website/pull/106#issuecomment-193658128
But, if your PRs about these modules are more complete than mine, maybe better to continue with yours and to close mine

lasley:
@eLBati - My legal is more up to date, but your cookies is better - I didn't notice any of the 8.0 changes before upgrading mine 😦

lasley:
Maybe you should base off of my legal (#194), then keep your's open for just cookies?

eLBati:
@lasley here it is https://github.com/OCA/website/pull/195

PR #197, opened 2016-04-29
[8.0][IMP][FIX][website_event_register_free] Better template reusability.

yajo:
This makes the free registration form a separate template that can be moved to another place easily in themes.

In the mean time, I added some cosmetic whitespace and fixed a bug that did not display the register button on events that had no max seats set.

@Tecnativa.

### DIFF
--- a/website_event_register_free/view/templates.xml
+++ b/website_event_register_free/view/templates.xml
@@ -6,36 +6,65 @@
                 <link rel="stylesheet" href="/website_event_register_free/static/src/css/website_event_register_free.css"/>
             </xpath>
         </template>
+
+        <template id="free_registration_form">
+            <form
+                t-attf-action="/event/#{event.id}/register/register_free"
+                method="post"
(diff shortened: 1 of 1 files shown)


website_event_register_free/view/templates.xml
```diff
@@ -6,36 +6,65 @@
...
+                                    No seats available
+                                </span>
+                            </td>
+                        </tr>
+                    </tbody>
+                </table>
+                <button
+                    type="submit"
+                    class="btn btn-primary btn-lg pull-right"
+                    t-if="event.seats_available or not event.seats_max">
```
pedrobaeza, line 53: seats_max field isn't defined in another module?

yajo, line 53: it's core

pedrobaeza:
:+1:

rafaelbn:
Thanks :+1:

PR #200, opened 2016-05-20
[8.0][ADD][website_seo_redirection] Choose your URLs!

yajo:
This module allows you to modify any URL that points to any controller to whatever you want.

@Tecnativa


website_seo_redirection/models/website_seo_redirection.py
```diff
@@ -0,0 +1,104 @@
...
+        self._url_format_check("destination")
+
+    @api.multi
+    def _url_format_check(self, field_name):
+        display = self._fields[field_name].string
+        for s in self:
+            value = s[field_name]
+            if not value.startswith("/"):
+                raise ValidationError(_("%s must start with `/`") % display)
+            for char in "?&=#":
```
elicoidal, line 51: I think this is not restrictive enough: http://www.netregister.biz/faqit.htm#1
You might want to have a look at this: http://stackoverflow.com/questions/2894902/check-for-a-valid-domain-name-in-a-string

yajo, line 51: The purpose is to redirect internally, not to any other domain.

yajo:
This should finally be working fine! @pedrobaeza

yajo:
I'm afraid Travis is wrong this time.

website_seo_redirection/models/website_seo_redirection.py
elicoidal, line 51: You still need to make sure that the users do not make any wrong input.
Probably not a blocker anyway

andhit-r:
Tested on runbot :+1:

.travis.yml
```diff
@@ -26,6 +26,8 @@ env:
   - TESTS="1" ODOO_REPO="OCA/OCB"
   - TESTS="1" ODOO_REPO="odoo/odoo" INCLUDE="website_signup_legal_page_required"
   - TESTS="1" ODOO_REPO="OCA/OCB" INCLUDE="website_signup_legal_page_required"
+  - TESTS="1" ODOO_REPO="odoo/odoo" EXCLUDE="website_seo_redirection"
```
pedrobaeza: If you do this, you're not testing the module, so this is cheating.

pedrobaeza: I see that there's other line that tests all modules, so the module is indeed being tested, but there's no too much sense then to have another test line that excludes it.

pedrobaeza:
Travis is not lying. It's warning you about some Pylint problems:

```
website_seo_redirection/models/ir_http.py:5:1: F401 'NotFound' imported but unused
website_seo_redirection/models/ir_http.py:6:1: F401 'conf' imported but unused
website_seo_redirection/models/ir_http.py:7:1: F401 'routing_map' imported but unused
```

and

```
website_seo_redirection/__init__.py:1: [W7903(javascript-lint), ] static/src/js/website_seo_redirection.tour.js:5:1: Use the function form of "use strict".
website_seo_redirection/__init__.py:1: [W7903(javascript-lint), ] static/src/js/website_seo_redirection.tour.js:7:5: 'openerp' is not defined.
website_seo_redirection/__init__.py:1: [W7903(javascript-lint), ] static/src/js/website_seo_redirection.tour.js:44:4: 'jQuery' is not defined.
```

.travis.yml
yajo: This is on purpose. From now on, this module will intercept all HTTP requests of Odoo when installed, and Travis installs all modules. It is supposed to work fine, but if it does not, you could become crazy to find the problem. It's there to help locating the problem in case it fails, assuming my code is not perfect.

pedrobaeza: Well, till now, we haven't done this, because it adds more time to each build, and if needed, it's done when we find a red build, but not always. You can add the lines as commented, and put a comment saying that if you suspect the module web_seo_redirection can cause the failing, uncomment that lines to try if this build is successful.

yajo: I'm pretty sure this time is different. I think if we are adding a module that modifies all of Odoo's behavior, obviously the cost of adding this line is that there's another build to try, but the benefit is that in case of failure, the :x: will be absolutely visible by anyone and will effectively save the contributor a big amount of time, not only in seeing where's the bug, but also in digging through the `.travis.yml` file to find a comment that somebody left there assuming people reads that file before pushing their PR.

I think this module is somehow like the whole OCB repo: it is supposed to be 100% compatible with existing installations, but the only way to ensure that is to test it on every build.

But I still can add a comment to it to explain why it's necessary.

pedrobaeza: I understand your concerns, but I prefer to let it commented. People will know how to act in case of error.

website_seo_redirection/README.rst
```diff
@@ -0,0 +1,73 @@
...
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/186/8.0
+
+Known issues / Roadmap
+======================
+
+* The frontend *Promote* view should be extended to display and modify this
+  information more easily.
```
rafaelbn, line 35: If not, only Odoo administrators can make these changes :cry: . This is usually done by website editors without configuration permissions.

rafaelbn:
Tested :+1: , very needed in frontend Promote
http://3147973-200-b9576b.runbot2.odoo-community.org/odoo-is-awesome-tecnativa-are-odoo-experts
:smile:

website_seo_redirection/README.rst
yajo, line 35: Yes, quite sad, but we're out of budget right now AFAIK.

lasley:
👍

paolovalier:
tested on runbot :+1: thanks

rafaelbn:
cc @rlizana @jctrey

Issue #85, opened 2015-10-14
Migration to version 9.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0
# Modules to migrate
- [ ] website_analytics_piwik - https://github.com/OCA/website/pull/145
- [x] website_anchor_smooth_scroll - By obulkin - #243
- [ ] website_backend_views
- [ ] website_blog_facebook_comment
- [x] website_blog_excerpt_img - By @Yajo - #298 
- [x] website_blog_mgmt - By @sergio-teruel - #147
- [x] website_blog_share - By @cubells - #291 
- [x] website_breadcrumb - By @khaeusler - #216 
- [x] website_canonical_url - By @Yajo - #299
- [x] website_crm_privacy_policy - By @Yajo - #302
- [ ] website_calendar_snippet - It should be renamed to _website_snippet_calendar_ - By @tarteo - #317 
- [x] website_container_fluid - By @Yajo - #300
- [x] website_cookie_notice - #195
- [x] website_crm_quick_answer - By @Yajo - #321
- [x] ~~website_event_register_free~~ - In core
- [x] ~~website_event_register_free_with_sale~~ - In core
- [ ] website_hr_department
- [x] website_img_bg_style - By @cubells - #290 
- [x] website_legal_page - By @lasley - #194
- [x] website_logo - ~~By @lasley - #192~~ - By @StephanRozendaal - #245 
- [ ] website_menu_by_user_status - By @hurrinico - #275 
- [ ] website_menu_multilang
- [x] website_no_crawler - By @RoelAdriaans-B-informed - #100
- [ ] website_portal_purchase - By @lukebranch
- [ ] website_sale_order_company - To be moved to OCA/e-commerce
- [ ] website_sale_unsaleable_options - To be moved to OCA/e-commerce
- [X] website_sale_vat_required - To be moved to OCA/e-commerce - By @Yajo - https://github.com/OCA/e-commerce/pull/174
- [x] ~~website_sale_collapse_categories~~ - Removed as it's integrated on v9 core
- [x] website_seo_redirection - By @lasley - #219
- [ ]  website_signup_legal_page_required - By @archetipo - #272 
- [x] website_snippet_anchor - By @Yajo - #288
- [x] website_snippet_big_button - By @Yajo - #303 
- [x] website_snippet_marginless_gallery - By @Yajo - #368 
- [ ] website_supplier_list - By @mikevhe18 - #286

[...]

PR #202, opened 2016-05-30
[MIG][9.0][website_portal_purchase] Migrate

yajo:
- This module was depending on a customized backport of website_portal. Now it uses core website_portal, so it cannot display contents separated by tabs. ~~That will come in a different module.~~ (It was being difficult: https://github.com/OCA/website/pull/203)
- Removed .pot file.
- Updated .js file to new format.
- Updated XML tags and XMLID.
- Added missing window action.
- Refactored to depend on `website_portal_v10`
- Relicensed to LGPL with permission from @sergio-teruel (original author).

WIP until:
- [x] #204

@Tecnativa @lukebranch #85


lukebranch:
@Yajo ,

Thanks for tagging me. I'll test this over the next couple of days.

yajo:
Please somebody mark this as WIP. I'm updating it to be based on `website_portal_v10` (#204)

yajo:
Refactored, description updated.

rafaelbn:
Please rebase

rafaelbn:
@Yajo please review travis errors

pedrobaeza:
@rafaelbn It's only a flake8 error. Please review it functionally.

rafaelbn:
Hi  @Yajo @pedrobaeza I cannnot test it, open runbot I get error "Could not execute command 'sass'"
Please review this for testing. Thanks

antespi:
Please remove `datetime` python module from `controllers/main.py`

```
website_portal_purchase/controllers/main.py:6:1: F401 'datetime' imported but unused
```

website_portal_purchase/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+# (c) 2015 Antiun Ingeniería S.L. - Sergio Teruel
+# (c) 2015 Antiun Ingeniería S.L. - Carlos Dauden
```
lasley, line 3: IMO adding company copyright + author is a bit excessive, and is going to begin to make our awesome short headers like the old long ones again. I know this was in the 8.0 & I'm not blocking for it on this PR, but I'm wondering if we're setting a bad precedence here?

@pedrobaeza - am I being overly anal here? It seems to me this was the intent of removing the author header, and instead focusing towards the contributors of the ReadMe.

lasley, line 3: These also do not follow the correct copyright guidelines and will all need to be changed to say `Copyright` instead of `(c)`, which further reinforces my issue with this copyright proliferation. 

`Copyright` is the only legal term that asserts a copyright, so I am blocking for that.

lasley:
Awesome thanks @Yajo - code looked good outside of a comment on the copyrights. 

Building runbot for a functional now, consider code 👍  outside of the blocking change of `(c)` to `Copyright` in all headers & a tentative thumb on functional if I don't report back with bugs

website_portal_purchase/__init__.py
rvalyi, line 3: @lasley I vote for keeping company + author. Company is the legal thing. Now author is the technical guy to contact and it's a reward for the work. Don't forget that all this code is given away for free, but I think it's important for our ecosystem to educate companies that they lower their risk and support the ecosystem if they hire the people who write the code instead of the marketing middlemen. Authors can change of company and it's important their work can still be tracked, file by file. IMHO we should keep the contributors of the README for those who contributed a very significant proportion of the module otherwise nothing is trackable and the meritocracy is broken.

lasley, line 3: Isn't the work instead attributed via the ReadMe & the Git commits themselves? We have line-by-line attribution built in, and I have personally found these copyrights difficult to maintain & dirty the diff. 

The copyright line is strictly for legal assertion of copyright. Adding a name to it after the company is ambiguously representing that the person also holds a copyright to that code, which is False assuming it was work for hire. If it was not work for hire, the copyright should be held by the individual only.

rvalyi, line 3: git holds the information indeed but it's way more cumbersome to retrieve. Also if a file got 50 one lines patch from 50 people git will give you 50 names that's not what you need when you need to quickly spot who is the author of the file. Adding authors should not be ambiguous for the copyright thing. The company holds the copyright indeed; it's a legal fact. The author guy if mentioned is just for knowing who to contact or hire. So I vote for keeping the current system.

lasley, line 3: It's probably a bad place to start this convo here now that I'm thinking about it, I'm starting a thread in contributors mailing list instead so we can hash this out.

I am still blocking over the `(c)`, as it is not a legal assertion of copyright. Whether the author is ambiguous is TBD and non-blocking.

rvalyi, line 3: ok. Well I think it was already discussed at the OCA, but why not. As for (c) instead of Copyright, I'm not sure if it's bad but at Akretion we use "Copyright" for all our contribs.

lasley, line 3: the `(c)` or `©` has to be accompanied by `Copyright` in order to be valid. I'm honestly not sure why they exist 😑

pedrobaeza, line 3: About multiple authors, I'm all with @rvalyi. Each person, from the same company or other, has to be included in the header of the file when his/her contribution is significant. About the word "Copyright", it was issued some time ago, but there's a lot of code inside OCA that uses the abbreviation. Personally, I don't care about that word, because in open source is very meaningless (in contrary to atribution), but if it's a blocking one, we can change. To use the abbreviation is to save chars for PEP8.

lasley, line 3: I remember that convo regarding the symbol vs the word, which is why I was quoting here. I guess it's a question of whether we upgrade things like this to current OCA style during a migration or not. Current templates have the word as a result of the convo.

lasley, line 3: But I also just ran across like 10 PRs with the symbol. So I'm also feeling lazy.

yajo, line 3: @lasley you made me read this: http://www.gnu.org/licenses/gpl-faq.html#RequiredToClaimCopyright

So [if you reside in a country adhered to the Berne convention](https://www.firstwriter.com/copyright/berne_convention.shtml), [your copyright gets automatically assigned to you until 50 years after you died](https://en.wikipedia.org/wiki/Berne_Convention#Copyright_term).

Besides that, to publish to OCA you sign the CLA, what makes you a "half copyright" holder. Given this is *GPL-licensed and you only have half of the copyright, and the other half is automatically assigned to you by international laws, I have to agree adding the copyright is irrelevant.

But I don't like removing it when other people have preferred to add it. I'll just change (c) and keep these discoveries in mind for future modules.

lasley, line 3: The OCA CLA actually lets you maintain all copyright, it just hands over some licensing and packaging rights to OCA. But yeah that nuance is kind of out of scope here anyways.

You're right about the implicit copyright too, which boils back down to my suggestion that the real copyright is the Git blame log.

So I've slept on this, as well as dove into a bunch of PRs looking at examples. Don't worry about changing the `(c)`, it's too proliferated through everything else. I feel like if I blocked here I would need to block everywhere else, and I simply don't have the energy for something that isn't even being used correctly.

carlosdauden: Same https://github.com/OCA/sale-workflow/pull/302#discussion_r78152583

carlosdauden:
:+1:

PR #191, opened 2016-04-18
New Module - Website Form Recaptcha v9

lasley:
Hi Friends,

I was about to release this in the LasLabs repo, then realized there are probably way more people than just me having this problem.

This PR includes two modules
- `website_form_recaptcha` - Provides ReCaptcha validation, and field via JS override similar to date fields (use `o_website_form_recaptcha` class on any el inside of form to trigger replace
- `website_crm_recaptcha` - Adds ReCaptcha field to standard contact form

PR #208, opened 2016-06-06
New Module - Website CRM Recaptcha v9 (Depends #191)

lasley:
This PR contains `website_crm_recaptcha`, which utilizes `website_form_recaptcha` widget to provide ReCaptcha field in Contact Us form.

This is dependent on #191

### DIFF
--- a/website_crm_recaptcha/README.rst
+++ b/website_crm_recaptcha/README.rst
@@ -1,0 +1,68 @@
+.. image:: https://img.shields.io/badge/license-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+=======================
+Website CRM - ReCaptcha
+=======================
+
+Adds a Recaptcha validation to Contact form on Website
+
+
+Configuration
+=============
+
+* Obtain ReCaptcha key from `Google <http://www.google.com/recaptcha/admin>`_
+* Add site key to `recaptcha.key.site` system parameter
+* Add secret key to `recaptcha.key.secret` system parameter
+
+
+Usage
+=====
+
+To use this module, you need to:
+
+* Go to the Contact form on Website
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/186/9.0
+
+
+
+Known Issues / Road Map
+=======================
+
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues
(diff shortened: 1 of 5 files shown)


rafaelbn:
Hi @lasley I see 2 commits here. If this PR depends on 191 then this PR must fails until 191 is merged.
cc @Yajo

lasley:
No problem, thanks @rafaelbn

lasley:
Rebased w/ finalized dependency to allow for functional review - @Yajo

pedrobaeza:
@lasley, is there any possibility of providing a demo API key and a demo page with the recaptcha for having directly something to try? (something similar to https://github.com/OCA/website/blob/8.0/website_container_fluid/demo/pages.xml)

lasley:
@pedrobaeza  Hmm I'm not necessarily opposed to creating a ReCaptcha key pair for Runbot. The domains need to be whitelisted though, so this would require updating if we ever add a new Runbot. That said, does the below cover our current setup?
- runbot1.odoo-community.org
- runbot2.odoo-community.org

I had no idea I could make pages as just a demo, but this makes sense in hindsight. I don't have a problem adding that either.

pedrobaeza:
Uhm, if we need to lock the source URL that's something very uncomfortable, because the idea is to have this demo to be used in any DB with demo data. That's why I ask for a demo key provided by Google with the freedom to be used in any website as demo.

lasley:
From what I understand there isn't such a thing as a demo key, and there isn't a way to add wildcard domains. 

Any real key, however, will work on a local hostname such as `localhost`, `127.0.0.1`, etc. It just says "Running in Local Only Mode" or something similar. The one caveat would be someone attempting to test on a live dev server, or another Runbot.

Completely open to suggestions here, having a fully functional demo is definitely the best situation.

lasley:
To note - the hostname is determined by the client so outside of the end testing user editing their host file to point a local address over to a Runbot, we're stuck 😦

pedrobaeza:
Oh, such a shame. OK, there's no other possibility, so it's good enough as is then. Thanks for the insights.

lasley:
Yeah it really irked me as well, but totally worth having the feature to cut down on some CRM and Signup spam. We're going to have the same issue with an Akismet module I'm creating as well 😑

lasley:
Rebased now that dependency has been merged. Following are some test credentials that we can use on Runbot for the moment in order to provide a real functional review. I'll just expire when this is merged.

Site: 6Ld0ZiMTAAAAAF8NhlvRru5p7VDpdHOh5mB9y1xY
Secret: 6Ld0ZiMTAAAAAOoQkRA7BPvkW3VnhzrvsWGNGrR1

pedrobaeza:
It works like a charm :+1:

yajo:
Awesome :+1:

PR #214, opened 2016-06-21
[ADD][9.0] Create new module website_field_autocomplete

lasley:
This module adds an abstract autocomplete field to Odoo website. Activate by adding the `s_website_autocomplete` class to any text input, and provide the necessary data vars (`data-model` being the only required).

I almost feel like this is something that should exist, so maybe I'm doing something wrong or missing something, but I definitely couldn't find any examples of an autocomplete on website.

One downfall of my implementation here is the use of Model & the underlying jsonRPC, which forces a user session for results. This means that in its current form this cannot be used on the product search input, for example, because the desired result would also have the public user receiving an autocomplete.

The only way I could think of to circumvent the use of model would be to just add something in the controller, which seems sloppy. I dunno though, may be worth it.

Let me know if you have any thoughts/suggestions!

### DIFF
--- a/website_field_autocomplete/README.rst
+++ b/website_field_autocomplete/README.rst
@@ -23,6 +23,7 @@ To use this module, you need to add an input field with the
            class="js_website_autocomplete"
            data-query-field="name"
            data-display-field="name"
+           data-value-field="id"
            data-limit="10"
            data-domain='[["website_published", "=", true]]'
            data-model="product.template"
@@ -30,19 +31,21 @@ To use this module, you need to add an input field with the
 
 Following is a breakdown of the available attributes & their defaults:
 
-+--------------------+---------------------------------------------+-------------+----------+
-|  Attribute         |  Description                                |  Default    | Required |
-+====================+=============================================+=============+==========+
-| data-model         | Model name to query                         |             | True     |
-+--------------------+---------------------------------------------+-------------+----------+
-| data-query-field   | Field to query when searching               | name        | False    |
-+--------------------+---------------------------------------------+-------------+----------+
-| data-display-field | Field to display (defaults to query-field)  | query-field | False    |
-+--------------------+---------------------------------------------+-------------+----------+
-| data-limit         | Limit results to this many                  | 10          | False    |
-+--------------------+---------------------------------------------+-------------+----------+
-| data-domain        | Additional domain for query                 | []          | False    |
-+--------------------+---------------------------------------------+-------------+----------+
++--------------------+---------------------------------------------+---------------+----------+
+|  Attribute         |  Description                                |  Default      | Required |
++====================+=============================================+===============+==========+
+| data-model         | Model name to query                         |               | True     |
++--------------------+---------------------------------------------+---------------+----------+
+| data-query-field   | Field to query when searching               | name          | False    |
++--------------------+---------------------------------------------+---------------+----------+
+| data-display-field | Field to display                            | query-field   | False    |
++--------------------+---------------------------------------------+---------------+----------+
+| data-value-field   | Field to use as form value                  | display-field | False    |
++--------------------+---------------------------------------------+---------------+----------+
+| data-limit         | Limit results to this many                  | 10            | False    |
++--------------------+---------------------------------------------+---------------+----------+
+| data-domain        | Additional domain for query                 | []            | False    |
++--------------------+---------------------------------------------+---------------+----------+
 
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
@@ -53,8 +56,7 @@ Following is a breakdown of the available attributes & their defaults:
 Known Issues / Road Map
 =======================
 
-* Replace jQuery UI Autocomplete w/ HTML5 Datalist when `ready for production
-<http://caniuse.com/#feat=datalist>`_.
+* Replace jQuery UI Autocomplete w/ HTML5 Datalist when `ready for production <http://caniuse.com/#feat=datalist>`_.
 
 
 Bug Tracker
--- a/website_field_autocomplete/README.rst
+++ b/website_field_autocomplete/README.rst
@@ -53,7 +53,8 @@ Following is a breakdown of the available attributes & their defaults:
 Known Issues / Road Map
 =======================
 
-* Use of Model/jsonRPC requires a user session.
+* Replace jQuery UI Autocomplete w/ HTML5 Datalist when `ready for production
+<http://caniuse.com/#feat=datalist>`_.
 
 
 Bug Tracker
--- a/website_field_autocomplete/README.rst
+++ b/website_field_autocomplete/README.rst
@@ -19,27 +19,27 @@ To use this module, you need to add an input field with the
 
     <input type="text"
            class="s_website_autocomplete"
-           data-queryField="name"
-           data-displayField="name"
+           data-query-field="name"
+           data-display-field="name"
            data-limit="5"
            data-domain='[["website_published", "=", true]]'
            data-model="product.template"
            />
 
 
-+-------------------+--------------------------------------------+------------+----------+
-|  Attribute        |  Description                               |  Default   | Required |
-+===================+============================================+============+==========+
-| data-model        | Model name to query                        |            | True     |
-+-------------------+--------------------------------------------+------------+----------+
-| data-queryField   | Field to query when searching              | name       | False    |
-+-------------------+--------------------------------------------+------------+----------+
-| data-displayField | Field to display (defaults to queryField)  | queryField | False    |
-+-------------------+--------------------------------------------+------------+----------+
-| data-limit        | Limit results to this many                 | 10         | False    |
-+-------------------+--------------------------------------------+------------+----------+
-| data-domain       | Additional domain for query                | []         | False    |
-+-------------------+--------------------------------------------+------------+----------+
++--------------------+---------------------------------------------+------------+----------+
+|  Attribute         |  Description                                |  Default   | Required |
++====================+=============================================+============+==========+
+| data-model         | Model name to query                         |            | True     |
++--------------------+---------------------------------------------+------------+----------+
+| data-query-field   | Field to query when searching               | name       | False    |
++--------------------+---------------------------------------------+------------+----------+
+| data-display-field | Field to display (defaults to query-field)  | query-field | False   |
++--------------------+---------------------------------------------+------------+----------+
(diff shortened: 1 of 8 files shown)


website_field_autocomplete/views/assets.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+    Copyright 2016 LasLabs Inc.
+    @license AGPL-3 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+
+<odoo>
+    <template id="assets_recaptcha" name="website_form_recaptcha Assets" inherit_id="website.assets_frontend">
```
elicoidal: please adapt the id and name

elicoidal:
small details.
:+1:

lasley:
Haha whoops, copy paste miss - thanks for the catch!

elicoidal:
:) thanks for the contribution!

lmignon:
:+1: LGTM (Code review no test)

website_field_autocomplete/static/src/js/field_autocomplete.js
```diff
@@ -0,0 +1,57 @@
...
+
+    start: function() {
+
+      let self = this;
+
+      this.$target.autocomplete({
+        source: function(request, response) {
+          // Must use var, let yields undefined
+          var QueryModel = new Model(self.$target.data('model'));
+          let queryField = self.$target.data('queryField');
```
yajo: While I love `let`, I think it's not yet ready for production use. http://caniuse.com/#feat=let

website_field_autocomplete/static/src/js/field_autocomplete.js
```diff
@@ -0,0 +1,57 @@
...
+          if (!queryField) {
+            queryField = 'name';
+          }
+          let displayField = self.$target.data('displayField');
```
yajo: You can just append `|| queryField` and avoid lots of `if`s

website_field_autocomplete/static/src/js/field_autocomplete.js
```diff
@@ -0,0 +1,57 @@
...
+
+odoo.define('website_field_autocomplete.field_autocomplete', function(require){
+  "use strict";
+
+  var snippet_animation = require('web_editor.snippets.animation');
+  var Model = require('web.Model');
+
+  snippet_animation.registry.field_autocomplete = snippet_animation.Class.extend({
+
+    selector: '.s_website_autocomplete',
```
yajo: These kind of selectors tend to start by `js_` instead. It's a non-written standard for Odoo.

website_field_autocomplete/static/src/js/field_autocomplete.js
```diff
@@ -0,0 +1,57 @@
...
+
+    start: function() {
+
+      let self = this;
+
+      this.$target.autocomplete({
+        source: function(request, response) {
```
yajo, line 57: I'd love to see this method as a method directly on the animation object, to be able to better extend it on submodules.

website_field_autocomplete/static/src/js/field_autocomplete.js
```diff
@@ -0,0 +1,57 @@
...
+          let limit = self.$target.data('limit');
+          if (!limit) {
+            limit = 10;
+          }
+          let domain = [[queryField, 'ilike', request.term]];
+          let add_domain = self.$target.data('domain');
+          if (add_domain) {
+            domain = domain.concat(add_domain);
+          }
+          QueryModel.call('search_read', [domain, [displayField]], {limit: limit})
```
yajo: When you do [this](https://github.com/OCA/website/pull/214/files#r68006144), then please return this deferred.

website_field_autocomplete/README.rst
```diff
@@ -0,0 +1,86 @@
...
+
+
++-------------------+--------------------------------------------+------------+----------+
+|  Attribute        |  Description                               |  Default   | Required |
++===================+============================================+============+==========+
+| data-model        | Model name to query                        |            | True     |
++-------------------+--------------------------------------------+------------+----------+
+| data-queryField   | Field to query when searching              | name       | False    |
++-------------------+--------------------------------------------+------------+----------+
+| data-displayField | Field to display (defaults to queryField)  | queryField | False    |
```
yajo: I'd prefer `data-query-field` and `data-display-field`. Personal liking though, you are not required to satisfy it :yum:

website_field_autocomplete/static/src/js/field_autocomplete.js
lasley: Well that saddens me. I was reading something that it was recommended for all new JS to utilize `let` instead, but this must have been strictly Node. Sometimes I wish there was a better delineation, but I guess that's the point 😛 

Also thanks for the website, this is much better than the one I was using.

website_field_autocomplete/static/src/js/field_autocomplete.js
lasley: Will change - Out of curiosity, is this an Odoo module standard in order to not conflict with Odoo core namespacing? I used the `s_` because it seems that's what Odoo SA is using such as `s_website_form`

website_field_autocomplete/static/src/js/field_autocomplete.js
lasley: Would you mind elaborating a bit here? Isn't the `.then(function` slightly below this a deferred? Am I supposed to `QueryModel.then(call...`?

hbrunn: returning the deferred object(s) a function creates is a good habit because then other code can attach to this deferred and do stuff on completion/failure. Makes your code better by increasing modularity, it's a major pita if you need to do this and upstream code fails to publish the relevant deferred object(s). @Yajo has painful experience with that if I remember correctly

lasley: Hmm interesting, I've never thought of a deferred for use like this. It totally makes sense though because then we can call the `super`, catch the result, append whatever handlers/mutations we need, then continue down the line. Freaking brilliant!

lasley:
All requested updates made. Thanks for the review @Yajo and to @hbrunn for providing the deferred elaboration

website_field_autocomplete/static/src/js/field_autocomplete.js
yajo: I have seen many different prefixes: js, s, o, oe... That happens when there's no written standard, but after working some time for it, it seems o_ is used for CSS selectors and js_ for JS-only ones. Personal perception only, though.

website_field_autocomplete/static/src/js/field_autocomplete.js
yajo: That's their purpose after all. Although you cannot actually modify a deferred's "test" function, but you can append handlers as @hbrunn said.

website_field_autocomplete/static/src/js/field_autocomplete.js
```diff
@@ -10,44 +10,39 @@ odoo.define('website_field_autocomplete.field_autocomplete', function(require){
...
+    /* Query remote server for autocomplete suggestions
+     * @param request object Request from jQueryUI Autocomplete
+     * @param response function Callback for response, accepts array of str
+     */
+    autocomplete: function(self, request, response) {
+      var QueryModel = new Model(self.$target.data('model'));
+      var queryField = self.$target.data('query-field') || 'name';
+      var displayField = self.$target.data('display-field') || queryField;
+      var limit = self.$target.data('limit') || 10;
+      var domain = [[queryField, 'ilike', request.term]];
```
yajo: To make it even better, you could move most of these declarations to `start()` and save them as variables under `this`. Submodules would not need to redeclare them.

website_field_autocomplete/static/src/js/field_autocomplete.js
lasley: Cool thanks for the elaboration - I'll keep this in mind going forward

lasley:
Again great idea @Yajo - implemented!

yajo:
Great, now the last step to make it perfect would be to add a simple demo page that shows its functionality to test in runbot. Could you add it please?

yajo:
I tested this on runbot and it's great. You got my :+1: :+1: even without the demo data.

I'd like however if you could add to roadmap to replace jQueryUI's autocomplete for [native HTML5 datalist feature when it is ready for production](http://caniuse.com/#feat=datalist).

yajo:
It does not work when logged out. Is that expected?

lasley:
Yeah that's in the roadmap and in my PR notes - this is due to use of Model, which uses underlying JsonRPC & thus requires a non-public user session. 

Do you know of some abstract controller that could do something similar? I would really hate to make one specifically for this purpose, but I know my client is going to ask for autocomplete on their E-Commerce search as soon as I implement this in the area of immediate concern.

lasley:
Now that you said something about the limitation, it bugged me. Implemented a controller so this will now work w/o a user session. I assume this is secure enough due to the underlying model permissions?

I also added a `this.fields`, which can be manipulated in child modules just in case they want to get some more data back than just the display field.

lasley:
As a thought experiment, I extended this module (will submit that PR in a day or two).

Turns out I should also be returning the records in the promise in order to allow promise chaining. Done!

I also added a `data-value-field` to allow for separate labels and values to allow for the use case of requiring an ID for form submission or the like.

website_field_autocomplete/demo/field_autocomplete_demo.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+    Copyright 2016 LasLabs Inc.
+    @license AGPL-3 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+
+<odoo>
+    <template id="field_autocomplete_demo" name="Field Autocomplete Demo" inherit_id="website.homepage">
```
yajo: Do you think you could put this in a separate demo page (as others you find in the runbot)?

lasley:
Demo data moved to a separate page

yajo:
Functional test does not work, but I guess it's because of permissions of the public user again, so :+1: (although if you use a different model for the demo, it would be better; not sure which though).

rafaelbn:
Please @lasley add a funcional demo which I can test it works please before merging. Thanks a lot! Great work!

rafaelbn:
Super:

PR #217, opened 2016-06-25
[ADD][9.0] New Module website_field_autocomplete_related

lasley:
This PR depends on #214 and was really just a thought project that kind of rabbit holed into something useful.

New module extends `website_field_autocomplete` to allow for related autocomplete fields. Use case being an instance where you have a few different contexts you would like to search on, then want to fill in the related fields based on the results.

The related fields do not necessarily need to be autocompletes, they only need to have the same `data-relate-group`. 

I included demo data this time too!

This is my use case that I'm playing with. Top two are autocomplete fields, everything else is just an input. All fields are related, thus the data is filled in when the autocomplete is selected. Sort of like a ghetto onchange:

image


lasley:
Rebased now that dependency has been merged.

lasley:
@rafaelbn - Would you mind putting a WIP tag on this please? As I was implementing this into a complex form, I realized there needs to be isolation between send/receive + support for relationals.

pedrobaeza:
Changed

lasley:
Thanks @pedrobaeza !

lasley:
This is ready, but is now dependent on #228

lasley:
Rebased; this is ready for review

lasley:
This is affected by #240

lasley:
Rebased for 403 fix

lasley:
Crap this was broken by switching the base library to use `search_read` during PR reviews on the base module. Back to a work in progress here 💥

lasley:
Oops so seems this did work, I just screwed up something with the original rebase onto the changes made in `website_field_autocomplete` that caused me to go down a dumb path. Fixed base and reverted my rabbit hole - this is ready to review

lasley:
@t3ddftw please review

website_field_autocomplete_related/tests/test_controller.py
```diff
@@ -0,0 +1,73 @@
+# -*- coding: utf-8 -*-
+# © 2016 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import mock
+from contextlib import contextmanager
+
+from openerp.tests.common import TransactionCase
+
+from openerp.addons.website_field_autocomplete_related.controllers import main
```
tedsalmon: Why not a relative import?

tedsalmon (changes requested):
Everything looks good except some issues lint was warning of:

1. The following files have mode +x
```
./__init__.py
./__openerp__.py
./README.rst
./controllers/__init__.py
./controllers/main.py
```
2. Import in tests - My comment is inline.

lasley:
Thanks for the review @t3ddftw - all points fixed.

tedsalmon (approved):
Thanks @lasley, LGTM 👍

yajo (changes requested):
I can't understand what is this for 😕 

I see the usage section, but I don't really understand it much; I go to the demo page and behavior is weird: autocompletes only on company and URL. When I use URL it fills an ID (not an URL); when I use company it goes blank after choosing one...

Maybe a working demo would show what's this for...

lasley:
Oops sorry for the missed comment @Yajo. There was an issue here that broke the existing `website_field_autocomplete` module, which was the URL to ID thing you noticed. I pushed the same fixes here that I did for the v10 branch, so this should work now.

Regarding what this does - it's just an autocomplete mechanism that will also fill in other fields. We use it pretty heavily in the prescription checkout process in medical due to lack of real forms. The demo should now work, so it probably will also make more sense.

yajo:
I think the demo does not work yet...
The URL works fine, yes, but when I autofill a company, the field gets emptied.

lasley:
@Yajo - Oops it seems I was just trying to do something unsupported with the company field & I've been using the name one for testing. Removed the company field from docs and demo.

Just a note, the issue is that I haven't built dot notation in on the search side of things, so while the company can receive the `company_id.name` it cannot search by it.

yajo (approved):
OK now I have been able to test it properly. The point I was missing is that I had to be logged in, otherwise the effect is not noticed.

Guess what? https://github.com/OCA/maintainer-quality-tools/issues/471 strikes back! I'm tired of it 😭

Ignoring travis, LGTM.

lasley:
> The point I was missing is that I had to be logged in, otherwise the effect is not noticed.

Ohhhh yeah ok so I can fix this if I switch the demo up to URL model instead of partner. Remembering back on the base module, the logged in issue was what caused me to switch that to URL too.

simahawk:
anyone willing to finish this?

simahawk:
outdated and David is not working on Odoo anymore

PR #219, opened 2016-06-30
[MIG][9.0] website_seo_redirection to v9

lasley:
This PR migrates website_seo_redirection to v9. 

Not many changes required, mainly just the JS API and moving from Settings => Configuration to Settings => Technical.


rafaelbn:
Tested in runbot :+1: 

http://3149713-219-fa8895.runbot2.odoo-community.org/odoo/seo

I'm not sure if that is the best place to add SEO Redirections. What do you think about moving it to "Website Admin" menu?

lasley:
Good call @rafaelbn - done

rafaelbn:
Hi @lasley , what do you think about achiving this https://github.com/OCA/website/pull/219/files#diff-71d0f4748b2f745452375d5662d46aa1R34 :smile:

rafaelbn:
You can test how it works in promote with this module: https://github.com/treytux/odoo-website/tree/master/website_url_friendly

lasley:
@rafaelbn - Hmm that makes sense. So from a high level, I would just be adding a section in the Promote window of website that will allow you to configure slugs to redirect to the page that you are currently on.

I'm somewhat stumped though - it seems I need to extend `website.seo.HtmlPage`, but that interface is not being exposed in the new JS via a return:
- The sample - https://github.com/treytux/odoo-website/blob/master/website_url_friendly/static/src/js/website.seo.js#L22
- Website SEO Returns - https://github.com/odoo/odoo/blob/c20ec2ced61555515752dc8b24236499cce4643d/addons/website/static/src/js/website.seo.js#L570

Is there another way for me to extend a class inside of a module that hasn't directly exposed it? I feel like there should be, but I'm still learning the nuances between Odoo 8 and 9 JS conversions.

rafaelbn:
Maybe @Yajo could give you a clue

yajo:
[It's a JS Qweb template](https://github.com/odoo/odoo/blob/c20ec2ced61555515752dc8b24236499cce4643d/addons/website/static/src/xml/website.seo.xml). Those are extended in a similar way to Python Qweb ones, but using jQuery selectors to target the nodes; [just check the docs](https://www.odoo.com/documentation/9.0/reference/qweb.html#javascript). I hope it helps :wink:

lasley:
Thanks @Yajo - I've got it injected, but there's another part that they are doing in the example that I cannot seem to replicate. No matter though- I'm now angry enough at it that I will likely have it figured out in a few week or so 😉 

We can consider this a WIP if we want to block for this feature @rafaelbn

rafaelbn:
Let's say 1 month. After that we will add the feature to known issue and merge as is.

rafaelbn:
Please @kutyel take a look here. Could you help?

lasley:
@rafaelbn Ok so I've looked deeper here, and I will have to create a whole new page instead of being able to just inject in the existing promotion page like was done in the example.

This isn't just a simple website_form thing though, and will require some tinkering on my part - namely with creation and use of JS QWeb templates, with some sort of inheritance from the website_form JS.

TBH the scope of my project is to utilize this for an automated conversion from other systems, so the interface isn't incredibly important to me (Wordpress => Odoo migration). I made a ticket for this, but it got pushed back a few sprints due to some other timelines. 

I propose we merge as is, then once the ticket comes up on my end & if it's not already done, we will rock it out. Thoughts?

rafaelbn:
Hi @rlizana @jctrey could you help here? :smile:

rafaelbn:
Hi @rlizana @jctrey could you help here? :smile: 

We would like to achive this functionality: https://github.com/OCA/website/pull/219/files#diff-71d0f4748b2f745452375d5662d46aa1R34

rafaelbn:
Let's merge this one. We can improve it later. :+1: 
cc @Yajo @pedrobaeza @lasley

yajo:
Yes, the plan is clear, we just need the time and budget.

yajo:
:+1: Code review. Functional not possible due to runbot not available.

yajo:
:+1: Functional review

PR #223, opened 2016-07-06
[8.0][IMP][website_snippet_country_dropdown] Choose user's country by default

yajo:
With this patch, when the currently logged-in user has a country in its profile, it will be chosen by default for the flag.

@Tecnativa.

### DIFF
--- a/website_snippet_country_dropdown/views/snippets.xml
+++ b/website_snippet_country_dropdown/views/snippets.xml
@@ -54,7 +54,7 @@
 <template id="country_dropdown" name="Country Dropdown">
     <t t-set="default_value" t-value="default_value or ''"/>
     <t t-set="countries" t-value="request.env['res.country'].search([])"/>
-    <t t-set="default_country" t-value="default_country or countries.search([('code', '=', default_value[:2])]) or website.company_id.country_id or countries[:1]"/>
+    <t t-set="default_country" t-value="default_country or countries.search([('code', '=', default_value[:2])]) or request.env.user.country_id or website.company_id.country_id or countries[:1]"/>
     <div class="input-group col-md-12 js_country_dropdown" t-ignore="true">
         <div class="input-group-btn flag-button">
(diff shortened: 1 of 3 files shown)


website_snippet_country_dropdown/views/snippets.xml
```diff
@@ -54,7 +54,7 @@
 <template id="country_dropdown" name="Country Dropdown">
     <t t-set="default_value" t-value="default_value or ''"/>
     <t t-set="countries" t-value="request.env['res.country'].search([])"/>
-    <t t-set="default_country" t-value="default_country or countries.search([('code', '=', default_value[:2])]) or website.company_id.country_id or countries[:1]"/>
+    <t t-set="default_country" t-value="default_country or countries.search([('code', '=', default_value[:2])]) or request.env.user.country_id or website.company_id.country_id or countries[:1]"/>
```
pedrobaeza, line 57: You should check first if default_value is set, or you will end up with a search of an empty code with unexpected results.

yajo, line 57: As you can see, `default_value` defaults to an empty string. Odoo saves empty values as `False`, not as empty strings, so searching for empty strings will always return a empty recordset, unless your database is broken which is not this module's problem IMO.

You can test that by going to sales > localization > countries, creating a country with an empty code, and searching for "country code is equal to (and leave search box blank)".

pedrobaeza, line 57: OK then. A bit twisted, because it's very easy to include `..or (default_value and countries.search([('code', '=', default_value[:2])])) or...` for readability purposes, but it works as you say.

yajo, line 57: Thanks for the review anyway :wink:

pedrobaeza:
:+1:

carlosdauden:
:+1:

elicoidal:
:+1:

PR #228, opened 2016-07-09
[9.0][IMP] website_field_autocomplete: Upgrade for modularity

lasley:
Fairly straightforward upgrade here- I just split the controller data getter into two methods & added a `self.record_ids` representing the search data in order to allow for easier manipulation before handing back to client.

While I was there I noticed I was being stupid and not defining the dataType in the AJAX call, so I fixed that too.

### DIFF
--- a/website_field_autocomplete/controllers/main.py
+++ b/website_field_autocomplete/controllers/main.py
@@ -17,15 +17,30 @@ class Website(Website):
         methods=['GET'],
         website=True,
     )
-    def _get_autocomplete_data(self, model, **kwargs):
-        res = []
+    def _get_field_autocomplete(self, model, **kwargs):
+        """ Return json autocomplete data """
         domain = json.loads(kwargs.get('domain', "[]"))
         fields = json.loads(kwargs.get('fields', "[]"))
         limit = kwargs.get('limit', None)
+        res = self._get_autocomplete_data(model, domain, fields, limit)
+        return json.dumps(res.values())
+
+    def _get_autocomplete_data(self, model, domain, fields, limit=None):
+        """ Gets and returns raw record data
+        Params:
+            model: Model name to query on
+            domain: Search domain
+            fields: List of fields to get
+            limit: Limit results to
+        Returns:
+            Dict of record dicts, keyed by ID
+        """
+        res = {}
         if limit:
             limit = int(limit)
-        for rec_id in request.env[model].search(domain, limit=limit):
-            res.append({
+        self.record_ids = request.env[model].search(domain, limit=limit)
+        for rec_id in self.record_ids:
+            res[rec_id.id] = {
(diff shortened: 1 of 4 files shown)


rafaelbn:
It is intended that auto-complete fields only work if you are logged or also for public users?

lasley:
It should work for logged ~~in~~ out users, however the demo will not due to it being on res.partner. I couldn't think of a model that would work though without bringing in a dependency.

rafaelbn:
model=website.menu?
Funcionaly :+1: 
As this is an improvement you can write in the tittle [9.0][IMP] this could make that more people review if they are using the module.
@sergio-teruel @carlosdauden @yajo what do you think?

lasley:
Good call w/ website.menu, that does indeed work.

Will do on the [IMP] tag - I've been trying to pay attention to what everyone else is doing there in order to conform, but guidance never hurts either 😄

website_field_autocomplete/controllers/main.py
```diff
@@ -17,15 +17,30 @@ class Website(Website):
...
+            Dict of record dicts, keyed by ID
+        """
+        res = {}
         if limit:
             limit = int(limit)
-        for rec_id in request.env[model].search(domain, limit=limit):
-            res.append({
+        self.record_ids = request.env[model].search(domain, limit=limit)
+        for rec_id in self.record_ids:
+            res[rec_id.id] = {
```
yajo: Why don't you make the widget call `search_read()` through JSONRPC and forget about the controllers?

lasley: This was how I originally implemented it, until you made a comment about public access ;)

JsonRPC requires a non-public user session to operate. Downside of the implementation there was requiring a session, thus this would not work without logging in. It now does, as can be seen by the demo on `website.menu`

yajo: Ah sorry I thought I already asked that but did not see where. I have a floating goblin here that tells me that it should be done that way, but controllers are OK too.

Anyway, your code seems to be reimplementing the `search_read` method. Why not just use it?

lasley: Haha yeah, it was in the original PR for this module.

I am definitely reimplementing `search_read`, but is there actually a controller `search_read` that I can use instead? 

The issue with using JsonRPC & `search_read` is that it doesn't work when accessing anonymously - https://www.odoo.com/documentation/9.0/api_integration.html#logging-in

yajo: I was meaning that you can just use `search_read` here.

lasley: Ohhhh yup you're totally right. Dunno why I didn't think about that, sorry for being so thick there 😛

yajo:
:+1:

PR #230, opened 2016-07-15
[9.0][website_snippet_country_dropdown] Migrate.

yajo:
Module migrated to v9.

:fearful: Snippets framework have changed a lot in v9!

@Tecnativa


rafaelbn:
Tested, please open runbot and check what happens with website, it says "could not execute commnad 'sass'

yajo:
Most probably because Sass is not installed in runbot. @pedrobaeza Can you help here please?

rafaelbn:
Cannot test this one @pedrobaeza

rafaelbn:
Go to runbot -> Open - Error: "could not execute commnad 'sass'
cc @pedrobaeza

website_snippet_country_dropdown/README.rst
```diff
@@ -0,0 +1,100 @@
...
+All variables you can use to modify its behavior:
+
+* ``complete_field`` to give it a name. Usually it will match the field name.
+* ``default_value`` for the ``complete_field``.
+* ``countries`` as a ``res.country`` ORM recordset.
+* ``default_country`` as a ``res.country`` record.
+* ``no_country_placeholder``.
+
+You can view an example in
+https://github.com/OCA/e-commerce/tree/8.0/website_sale_checkout_country_vat.
```
pedrobaeza: You should change this to a relative reference (only mention website_sale_checkout_country_vat in OCA/e-commerce)

pedrobaeza:
Sorry, I can't install binary files on runbot servers. Only Python libraries. You should contact @gurneyalex for that task.

gurneyalex:
@Yajo @rafaelbn @pedrobaeza I just installed ruby-sass on runbot (and forced a rebuild of the PR).

pedrobaeza:
@gurneyalex, it seems that there's still missing a binary called bootstrap-sass.

antespi:
@gurneyalex This is what we do to have all sass dependencies installed:

```
apt install -y ruby ruby2.3 ruby2.3-dev
gem install bootstrap-sass compass sass
```

gurneyalex:
@antespi @yajo I just installed these using gems, and triggered a rebuild. I'm not sure the required PATH update has been propagated up to the Odoo servers of the runbots, though. Please ping me if you see errors on the new build.

yajo:
It seems to work.

gurneyalex:
:+1:

rafaelbn:
Hi @Yajo , I tried to add snippet but is in red all the time. Could you try to check it please. See screenshots:

yajo:
Please keep in mind that you can **never** drag a field to a controller-dependant form. If the controller is not ready to accept such field, it could either do not do what you expect it to do, or simply blow up.

Because of that, Odoo seems to even forbid dragging the snippet into a `<form>` element that is inside an editable area (such as the one in the demo page). Not sure why, but anyway since the main purpose of this module is to provide a basis for other modules to use it in controller-enabled forms, and that has to be done from a module, I think I will just add that to known issues.

rafaelbn:
:+1:

PR #234, opened 2016-07-22
[8.0][IMP][website_legal_page] Make template texts inheritable.

yajo:
If you need to change anything other than the URLs in this template, it's near to impossible.

I provide some hooks for that here. Just in case your legal pages are hosted somewhere else, or you don't have a legal advice, etc.

You should notice no change in UI.

@Tecnativa.

### DIFF
--- a/website_legal_page/__openerp__.py
+++ b/website_legal_page/__openerp__.py
@@ -23,7 +23,7 @@
 {
     'name': "Website legal page",
     'category': 'Website',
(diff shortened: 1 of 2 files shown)


pedrobaeza:
:+1:

rafaelbn:
Thanks!! :smile: :+1:

elicoidal:
:+1:

PR #240, opened 2016-07-29
[website_seo_redirection][FIX] Avoid some weird 403 errors.

### LINKED ISSUE #222: [8.0][website_slide] 403: Forbidden
module: website_slides
version: 8.0

**Steps to reproduce**
Install the module and go to the page which should open the home page slide.

On the demo page:
http://3150062-8-0-2e960e.runbot2.odoo-community.org/slides
there is a warning
403: Forbidden
When I installed I got the same warning.

**Current behavior**
403: Forbidden

**Expected behavior**
Homepage slide

Please tell me how to fix it? Thank you!

yajo:
This ~~hopefully~~ fixes #222.

@Tecnativa

### DIFF
--- a/website_seo_redirection/models/ir_http.py
+++ b/website_seo_redirection/models/ir_http.py
@@ -16,6 +16,10 @@ class IrHttp(models.AbstractModel):
         if not hasattr(request, "jsonrequest"):
             wsr = request.env["website.seo.redirection"]
 
+            # Requests at this point have no user, must remove `env` to force
+            # Odoo recompute it next time a controller needs it, with its user
+            del request.env
+
             try:
                 # Redirect user to SEO version of this URL if possible
                 return wsr.redirect_auto()


pedrobaeza:
Buy this way you are removing also a possible context dictionary. You should instead add the user if it doesn't exist. Is that possible?

yajo:
I'm just removing the environment, but context in `request.context` still exists. `request.env` is a lazy property, removing it makes Odoo rebuild it next time it is used, so other modules won't notice. And this hook resides before core dispatching, so there are very low chances that someone is using `request.env.context` before this.

pedrobaeza:
OK then, :+1:

lasley:
TIL on the lazy `env` computation 👍

pedrobaeza:
Fast-tracking as this is a blocking issue.

PR #241, opened 2016-07-29
[website_seo_redirection][FIX][9.0] Avoid some weird 403 errors.

lasley:
This cherry picks fix for #222 for v9

cc @Yajo @pedrobaeza


pedrobaeza:
:+1:

rafaelbn:
:+1:

rafaelbn:
I tested in runbot without this fix and with this fix. Bug https://github.com/OCA/website/issues/222 is solved also in v9. I think this can be a quick merge

pedrobaeza:
Let's do it then!

PR #242, opened 2016-08-12
[ADD][9.0] website_version

nicolas-petit:
Added base code from Odoo S.A.

Updated python code to make it PEP-8 compliant.
Added authoring information on python files.
Added a README.rst following the models present on OCA modules.

Fixed a bug when making a new version.
New versions ended up not being linked with the new view, and the old
version was linked instead.
This also caused unwanted publishing of new versions made from master.


website_version/__openerp__.py
```diff
@@ -0,0 +1,48 @@
...
+        'views/website_version_views.xml',
+        'views/res_config.xml',
+        'data/data.xml',
+    ],
+    'demo': [
+        'data/demo.xml',
+    ],
+    'qweb': ['static/src/xml/*.xml'],
+    'application': False,
+}
```
carlosdauden: Blank line

website_version/controllers/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+import main
```
carlosdauden: Blank line

website_version/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
```
yajo: Please use reduced headers everywhere.

website_version/__openerp__.py
```diff
@@ -0,0 +1,48 @@
...
+{
+    'name': 'Website Versioning',
+    'category': 'Website',
+    'summary': 'Allow to save all the versions of your website and allow to perform AB testing.',
+    'version': '9.0.1.0.0',
+    'description': """
+OpenERP Website CMS
+===================
+
+        """,
```
yajo: You have a README, drop this please.

website_version/__openerp__.py
```diff
@@ -0,0 +1,48 @@
...
+    'depends': ['website', 'marketing', 'google_account', 'mail'],
+    'author': 'Odoo S.A., Nicolas Petit (Clouder)',
```
yajo: Add OCA

website_version/__openerp__.py
```diff
@@ -0,0 +1,48 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
+{
+    'name': 'Website Versioning',
```
yajo: Module name (website_version) is the same as in Enterprise version? If yes, then please rename it to avoid conflicts. Something like website_version_ce for example.

website_version/__openerp__.py
yajo: Add EOF chars everywhere please.

website_version/controllers/__init__.py
yajo: from .

website_version/data/data.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<openerp>
```
yajo: Replace with `<odoo>` (here and everywhere) please

website_version/i18n/website_version.pot
```diff
@@ -0,0 +1,1640 @@
+# Translation of Odoo Server.
+# This file contains the translation of the following modules:
+#	* website_version
```
yajo: Remove unneeded .pot file

website_version/models/experiment.py
```diff
@@ -0,0 +1,227 @@
...
+
+    @api.multi
+    @api.constrains('website_id', 'experiment_version_ids')
+    def _check_website(self):
+        for exp in self:
+            for exp_ver in exp.experiment_version_ids:
+                if not exp_ver.version_id.website_id.id == exp.website_id.id:
+                    raise ValidationError('This experiment must have versions which are in the same website')
+
+    def read_group(self, cr, uid, domain, fields, groupby,
```
carlosdauden: Can you define this in new API or is it problematic?

website_version/__openerp__.py
nicolas-petit: Thanks for the feedback.
The latest commits should have fixed the issue.

nicolas-petit:
Thanks @carlosdauden and @Yajo for the feedback.
The latest commits should have fixed the issues you  both raised.

yajo:
Given this is a ~4700 lines PR, it would be great if you get all bots :green_heart: before doing more insightful reviews.

nicolas-petit:
Updated code to pass the automated test suite.

lasley:
Thanks for the submission @nicolas-petit - Is this a port of Odoo code? IMO we need to use the original commits similar to an upgrade to v9 that doesn't contain the module. You can do this by adding the Odoo git as another remote, then operating on it just like any other branch.

We also need the CLA disclaimer as noted in https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#backporting-odoo-modules

Also, you cannot add a copyright on top of something that is infinitely copywritten (Odoo SA uses 2016-TODAY, which creates an infinite copyright). I do not know the subtleties of being able to add an end date to their's though.

YannickB:
Hello @lasley , yes this is a port of Odoo code, the issue has been discussed on the mailing-list : https://www.odoo.com/groups/community-59/community-19495417?mode=thread&date_begin=&date_end=

We shall add the CLA disclaimer as you suggested and stick to your suggestions regarding the copyright.

Another issue regarding the licence : 
-No licence in the module we worked on
-As mentioned in the mailinglist, the download page were mentioning GPL licence
-I believe it was an error, the module was probably AGPL at this time (pre V9 release)
-Since it was first developed by Odoo SA, I believe it would be respectful to stick with the licence of Odoo CE, so that's why we are suggesting LGPL.
This is a mess, we really don't know which licence we should use. Any thoughts regarding this matter guys?

lasley:
I saw that mailing list entry, I've been plagued by the problem of joining these things too late and not being able to reply to anything 😦 

If there are doubts on whether this was actually LGPL & could possibly be GPL, I believe we would need to use GPL because LGPL allows use of software w/out source release and GPL does not.

yajo:
Could you point to the exact commit from where you extracted the module? If so, we could browse the tree at that point and see the LICENSE file

nicolas-petit:
@Yajo We can't as far as I know.

The code was downloaded from https://apps.openerp.com, on a page that doesn't exist anymore apparently.
Unless someone knows a way to find the version from a zipped source that contains no .git, I'm afraid it's not possible to point to a commit.

https://www.odoo.com/groups/community-59/community-19517711?mode=thread&date_begin=&date_end=
If you look at Martin Trigaux's answer above, he states that in theory the module is under AGPL.
Can we base our reasoning on this ?

lasley:
Theory and licensing don't really mix IMO. Unfortunately this starts getting into the grey area where I start consulting my IP attorney as well, but he's expensive.

Let's take a look at the zip, can you put it on a Google Drive or something?

pedrobaeza:
I think you don't need the zip. You can check master branch, the commit before the removal of the module.

nicolas-petit:
This would be the commit that moves website_version into entreprise.
https://github.com/odoo/odoo/commit/a11bab12264fc59a6c8f865dc0aa1099ba83c2f8

The parent commit seems to be this one: 
https://github.com/odoo/odoo/commit/9f1cb9ed97258c224127f5ff503886fad0abf5b2

pedrobaeza:
This seems to be LGPL code then, as Odoo moves their base code to that license before.

yajo:
Great @nicolas-petit! That's what I was meaning. Now you can extract the commit history as @lasley said in https://github.com/OCA/website/pull/242#issuecomment-240412319.

[If you review LICENSE at that point, you can see it was AGPL](https://github.com/odoo/odoo/blob/9f1cb9ed97258c224127f5ff503886fad0abf5b2/LICENSE).

pedrobaeza:
Although the LICENSE file, they already expresses their change to LGPL, and as the author can switch license, I would got for LGPL, but in this case, it's not too much relevant anyway.

nicolas-petit:
@Yajo @lasley Will that last message from the OCA CLAbot pose a problem ?

As I see it there is no way to make Odoo developers sign.

pedrobaeza:
No problem while the disclaimer is added in the README.

nicolas-petit:
@pedrobaeza I was actually making those changes when you made your comment.
They've been commited.

nicolas-petit:
Is there anything else that I need to fix?

pedrobaeza:
Well, as this module is very huge and requires a lot of time to be reviewed, it's not easy to get enough reviewers. Maybe you can review other PRs and say that review yours in payback.

website_version_ce/README.rst
```diff
@@ -0,0 +1,72 @@
...
+   :alt: License: LGPL-3
+
+
+**This module is a backport from Odoo SA and as such, it is not included in the OCA CLA. That means we do not have a copy of the copyright on it like all other OCA modules.**
+
+===============
+Website Version
+===============
+
+This module extends the functionality of the website by adding versionning
```
lasley: s/versionning/versioning

website_version_ce/controllers/main.py
```diff
@@ -0,0 +1,221 @@
...
+from openerp.tools.misc import DEFAULT_SERVER_DATETIME_FORMAT
+
+GOOGLE_ANALYTICS_CONFIGURED = 2
+GOOGLE_ANALYTICS_PARTIALLY_CONFIGURED = 1
+GOOGLE_ANALYTICS_NOT_CONFIGURED = 0
+
+
+class VersioningController(Website):
+
+    @http.route('/website_version_ce/change_version', type='json', auth="user",
```
lasley, line 16: What is the `_ce` suffix? IMO we should remove that

website_version_ce/controllers/main.py
```diff
@@ -0,0 +1,221 @@
...
+                website=True)
+    def change_version(self, version_id):
+        request.session['version_id'] = version_id
+        return version_id
+
+    @http.route('/website_version_ce/create_version', type='json', auth="user",
+                website=True)
+    def create_version(self, name, version_id=None):
+        if not name:
+            name = datetime.datetime.now().strftime(
```
lasley: `fields.Datetime.now()`

website_version_ce/controllers/main.py
```diff
@@ -0,0 +1,221 @@
...
+        # Check if client_id and client_secret are set for Google
+        gs_obj = request.env['google.service']
+        gm_obj = request.env['google.management']
+
+        client_id = gs_obj.get_client_id(
+            'management', context=kw.get('local_context'))
+        client_secret = gs_obj.get_client_secret(
+            'management', context=kw.get('local_context'))
+        if not client_id or not client_secret:
+            dummy, action = request.registry.get(
```
lasley: `request.env`?

website_version_ce/controllers/main.py
```diff
@@ -0,0 +1,221 @@
...
+                    request.uid,
+                    'website_version_ce',
+                    'action_config_settings_google_management'
+                )
+            return {
+                "status": "need_config_from_admin",
+                "url": '',
+                "action": action
+            }
+        url = gm_obj.authorize_google_uri(from_url=kw.get('fromurl'),
```
lasley: `fromurl` is killing my PEP8 senses. Possible to `from_url` instead?

website_version_ce/data/data.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
```
lasley: `<odoo noupdate="1">` & remove data tag

website_version_ce/models/experiment.py
```diff
@@ -0,0 +1,245 @@
...
+        required=True,
+        ondelete='cascade'
+    )
+    frequency = fields.Selection(
+        [('10', 'Less'), ('50', 'Medium'), ('80', 'More')],
+        string='Frequency', default='50')
+    google_index = fields.Integer(string='Google index')
+
+
+class Goals(models.Model):
```
lasley, line 14: Please separate these models by file

website_version_ce/models/experiment.py
```diff
@@ -0,0 +1,245 @@
...
+                    if exp_ver.search([
+                        ('version_id.view_ids.key', 'in',
+                         [v.key for v in exp_ver.version_id.view_ids]),
+                        ('experiment_id', '!=', exp_ver.experiment_id.id),
+                        ('experiment_id.website_id', '=',
+                         exp_ver.experiment_id.website_id.id),
+                        ('experiment_id.state', '=', 'running')
+                    ]):
+                        raise ValidationError(_(
+                            'This experiment contains a view which ' +
```
lasley: no need for the `+`, strings inside of parenthesis join automatically

website_version_ce/models/experiment.py
```diff
@@ -0,0 +1,245 @@
...
+        return True
+
+    @api.multi
+    @api.constrains('website_id', 'experiment_version_ids')
+    def _check_website(self):
+        for exp in self:
+            for exp_ver in exp.experiment_version_ids:
+                if not exp_ver.version_id.website_id.id == exp.website_id.id:
+                    raise ValidationError(_(
+                        'This experiment must have versions which ' +
```
lasley: no need for the `+`, strings inside of parenthesis join automatically

website_version_ce/models/experiment.py
```diff
@@ -0,0 +1,245 @@
...
+                        'are in the same website'))
+
+    @api.model
+    def read_group(self, domain, fields, groupby, offset=0, limit=None,
```
lasley, line 88: `groupby` should be `group_by`

website_version_ce/models/experiment.py
```diff
@@ -0,0 +1,245 @@
...
+                   context=None, orderby=False, lazy=True):
```
lasley, line 89: `orderby` should be `order_by`

website_version_ce/models/experiment.py
```diff
@@ -0,0 +1,245 @@
...
+        """ Override read_group to always display all states. """
+        if groupby and groupby[0] == "state":
+            # Default result structure
+            # states = self._get_state_list(cr, uid, context=context)
```
lasley: Please remove useless code

website_version_ce/models/experiment.py
```diff
@@ -0,0 +1,245 @@
...
+            return super(Experiment, self).read_group(
+                domain, fields, groupby, offset=offset, limit=limit,
+                context=context, orderby=orderby, lazy=lazy)
+
+    @api.multi
+    def _get_version_number(self):
+        for exp in self:
+            exp.version_number = len(exp.experiment_version_ids) + 1
+
+    name = fields.Char(string="Title", required=True)
```
lasley: Field definitions should be at the top of the file

website_version_ce/models/experiment.py
```diff
@@ -0,0 +1,245 @@
...
+            ('running', 'Running'),
+            ('paused', 'Paused'),
+            ('ended', 'Ended')
+        ],
+        'Status', required=True, copy=False, track_visibility='onchange',
+        default='running')
+    goal_id = fields.Many2one('website_version_ce.goals', string="Objective",
+                              required=True)
+    color = fields.Integer('Color Index')
+    version_number = fields.Integer(compute=_get_version_number,
```
lasley: Please use strings for computation definitions

website_version_ce/models/experiment.py
```diff
@@ -0,0 +1,245 @@
...
+                                    string='Version Number')
+    sequence = fields.Integer('Sequence', required=True, default=1)
+    google_id = fields.Char(string="Google id")
```
lasley: This string attr is redundant to the default, please remove

website_version_ce/models/experiment.py
```diff
@@ -0,0 +1,245 @@
...
+
+    @api.model
+    def create(self, vals):
+        exp = {
+            'name': vals['name'],
+            'objectiveMetric': self.env['website_version_ce.goals'].browse(
+                [vals['goal_id']])[0].google_ref,
```
lasley: No need to `[0]`:

```
self.env['website_version_ce.goals'].browse([vals['goal_id']]).google_ref
```

website_version_ce/models/experiment.py
```diff
@@ -0,0 +1,245 @@
...
+            'status': vals['state'],
+            'variations': [
+                {'name': 'master', 'url': 'http://localhost/master'}]
+        }
+        version_list = vals.get('experiment_version_ids', [])
+        for version in version_list:
+            if version[0] == 0:
+                name = self.env['website_version_ce.version'].browse(
```
lasley, line 139: No need for the `[0]`

website_version_ce/models/google_management.py
```diff
@@ -0,0 +1,217 @@
...
+        return {
+            'name': experiment['name'],
+            'status': experiment['status'],
+            'variations': experiment['variations']
+        }
+
+    @api.model
+    def create_an_experiment(self, data, website_id):
+        gs_pool = self.env['google.service']
+        website = self.env['website'].browse(website_id)[0]
```
lasley: No need for `[0]`

lasley: No need for `[0]`

lasley: No need for `[0]`

website_version_ce/models/google_management.py
```diff
@@ -0,0 +1,217 @@
...
+                   web_property_id,
+                   profile_id,
+                   self.get_token())
+        headers = {'Content-type': 'application/json', 'Accept': 'text/plain'}
+        data_json = simplejson.dumps(data)
+        try:
+            x = gs_pool._do_request(url, data_json, headers, type='POST')
+            result = x[1]['id']
+        except Exception, e:
+            _logger.info(_('An exception occured during the google ' +
```
lasley: I feel like this should be a `_logger.error` or `_logger.exception`

website_version_ce/models/google_management.py
```diff
@@ -0,0 +1,217 @@
...
+                           'analytics rquest: %s') % e)
+            raise
+        return result
+
+    @api.model
+    def update_an_experiment(self, data, google_id, website_id):
+        gs_pool = self.env['google.service']
+        website = self.env['website'].browse(website_id)[0]
```
lasley: No need for the `[0]`

website_version_ce/models/google_management.py
```diff
@@ -0,0 +1,217 @@
...
+        )
+        headers = {'Content-type': 'application/json', 'Accept': 'text/plain'}
+        data_json = simplejson.dumps(data)
+
+        return gs_pool._do_request(url, data_json, headers, type='PUT')
+
+    @api.model
+    def get_experiment_info(self, google_id, website_id):
+        gs_pool = self.env['google.service']
+        website = self.env['website'].browse(website_id)[0]
```
lasley: No need for the `[0]`

website_version_ce/models/google_management.py
```diff
@@ -0,0 +1,217 @@
...
+        url = ('/analytics/v3/management/accounts/%s/' +
+               'webproperties/%s/profiles/%s/goals') % (
+            account_id,
+            web_property_id,
+            profile_id
+        )
+        return gs_pool._do_request(url, params, headers, type='GET')
+
+    @api.model
+    def delete_an_experiment(self, google_id, website_id):
```
lasley, line 137: These CRUD methods seem to have an incredible amount of code duplication. Would it be possible to move the identical logic into a helper method instead?

website_version_ce/models/google_management.py
```diff
@@ -0,0 +1,217 @@
...
+            web_property_id,
+            profile_id,
+            google_id
+        )
+        return gs_pool._do_request(url, params, headers, type='GET')
+
+    @api.model
+    def get_goal_info(self, website_id):
+        gs_pool = self.env['google.service']
+        website = self.env['website'].browse(website_id)[0]
```
lasley: No need for the `[0]`

website_version_ce/models/google_management.py
```diff
@@ -0,0 +1,217 @@
...
+            account_id,
+            web_property_id,
+            profile_id
+        )
+        return gs_pool._do_request(url, params, headers, type='GET')
+
+    @api.model
+    def delete_an_experiment(self, google_id, website_id):
+        gs_pool = self.env['google.service']
+        website = self.env['website'].browse(website_id)[0]
```
lasley: No need for the `[0]`

website_version_ce/models/ir_qweb.py
```diff
@@ -0,0 +1,63 @@
...
+
+from openerp import models
+
+
+class QWeb(models.AbstractModel):
+    """ QWeb object for rendering stuff in the website context
+    """
+    _inherit = 'ir.qweb'
+
+    def render(self, cr, uid, id_or_xml_id, qwebcontext=None, loader=None,
```
lasley, line 19: Please convert this method to new API

website_version_ce/models/ir_ui_view.py
```diff
@@ -0,0 +1,182 @@
+# -*- coding: utf-8 -*-#
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from lxml import etree
+from openerp import tools, fields, models, api
+
+
+class View(models.Model):
```
lasley: Class names of models should be UpperCamelCased version of model name

website_version_ce/models/ir_ui_view.py
```diff
@@ -0,0 +1,182 @@
...
+
+    _inherit = "ir.ui.view"
+
+    version_id = fields.Many2one(
+        'website_version_ce.version', ondelete='cascade', string="Version")
```
lasley, line 13: `required=True`?

website_version_ce/models/ir_ui_view.py
```diff
@@ -0,0 +1,182 @@
...
+    def publish(self):
+        """To delete and replace views which are in the website
+        (in fact with website_id)
+        """
+        self.ensure_one()
+        master_record = self.search([
+            ('key', '=', self.key),
+            ('version_id', '=', False),
+            ('website_id', '=', self.website_id.id)
+        ])
```
lasley, line 60: `limit=1`?

website_version_ce/models/ir_ui_view.py
```diff
@@ -0,0 +1,182 @@
...
+                    right_ids[k.key] = k.id
+                    priority[k.key] = 2
+                # priority:2 take the original view
+                elif k.version_id.id is False and k.website_id.id is False:
+                    if not priority.get(k.key) or priority.get(k.key) < 2:
+                        right_ids[k.key] = k.id
+                        priority[k.key] = 1
+        return [x for x in arch if x[1] in right_ids.values()]
+
+    # To active or desactive the right views according to the key
```
lasley: s/desactive/deactivate

website_version_ce/models/ir_ui_view.py
```diff
@@ -0,0 +1,182 @@
...
+    def toggle(self, cr, uid, ids, context=None):
```
lasley: Please switch to new API

website_version_ce/models/res_config.py
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-#
```
lasley: Please name this file snake_case version of the model name

website_version_ce/models/res_config.py
```diff
@@ -0,0 +1,35 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from openerp import fields, models, api
+
+
+class AnalyticsConfigSettings(models.TransientModel):
```
lasley: Class names should be UpperCamelCased version  of model name

website_version_ce/models/version.py
```diff
@@ -0,0 +1,98 @@
+# -*- coding: utf-8 -*-#
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from openerp import fields, models, api
+from openerp.exceptions import Warning as UserError
+from openerp.http import request
+from openerp.tools.translate import _
+
+
+class Version(models.Model):
```
lasley, line 10: Class names of models should be UpperCamelCased version of model name

website_version_ce/models/version.py
```diff
@@ -0,0 +1,98 @@
...
+    """ A version is a set of qweb views which differs from the qweb views in
+    production for the website.
+    """
+
+    _name = "website_version_ce.version"
+
+    name = fields.Char(string="Title", required=True)
+    view_ids = fields.One2many('ir.ui.view', 'version_id', string="View",
+                               copy=True)
+    website_id = fields.Many2one('website', ondelete='cascade',
```
lasley, line 20: `required=True`?

lasley, line 20: `required=True`?

website_version_ce/models/website.py
```diff
@@ -0,0 +1,101 @@
+# -*- coding: utf-8 -*-#
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+import json
+from openerp import fields, models, api
+from openerp.http import request
+import random
+
+
+class NewWebsite(models.Model):
```
lasley: Class names of models should be UpperCamelCased version of model name

website_version_ce/models/website.py
```diff
@@ -0,0 +1,101 @@
...
+
+    _inherit = "website"
+
+    tuto_google_analytics = fields.Boolean("How to get GA key and View ID")
+    google_analytics_view_id = fields.Char('View ID')
+    google_management_authorization = fields.Char('Google authorization')
+
+    @api.model
+    def get_current_version(self, context=None):
```
lasley: context is provided in env when using new api, please remove

website_version_ce/models/website.py
```diff
@@ -0,0 +1,101 @@
...
+                pond_sum = 0
+                for exp_snap in exp.experiment_version_ids:
+                    result.append([int(exp_snap.frequency)+pond_sum,
+                                   exp_snap.version_id.id])
+                    pond_sum += int(exp_snap.frequency)
+                if pond_sum:
+                    # Setting master default frequency at 50
+                    pond_sum += 50
+                    # Setting cookie to avoid problems when db restarts
+                    main_experiment[exp.google_id] = str(0)
```
lasley: Type casting a static int to a string seems a bit weird IMO, why not just `"0"`?

website_version_ce/static/src/js/menu_version.js
```diff
@@ -0,0 +1,294 @@
...
+                    wizard.appendTo($('body')).modal({"keyboard" :true});
+                    wizard.on('hidden.bs.modal', function () {$(this).remove();});
+                });
+            }
+        });
+        wizardA.on('hidden.bs.modal', function () {$(this).remove();});
+    },
+
+    change_version: function(event) {
+        var version_id = parseInt($(event.target).closest("li").data("version_id"));
```
lasley, line 78: event.currentTarget

website_version_ce/static/src/js/menu_version.js
```diff
@@ -0,0 +1,294 @@
...
+        if(! version_id){
+            version_id = 0;//By default master
+        }
+        ajax.jsonRpc( '/website_version_ce/change_version', 'call', { 'version_id':version_id }).then(function (result) {
+                location.reload();
+            });
+    },
+
+    delete_version: function(event) {
+        var version_id = parseInt($(event.target).closest("li").data("version_id"));
```
lasley, line 88: event.currentTarget

website_version_ce/static/src/js/menu_version.js
```diff
@@ -0,0 +1,294 @@
...
+                        });
+                    });
+                    wizardA.on('hidden.bs.modal', function () {$(this).remove();});
+                }
+            });
+    },
+
+    publish_version: function(event) {
+        var version_id = base.get_context().version_id;
+        var name = $('#version-menu-button').attr('data-version_name');
```
lasley: `$('#version-menu-button').data('version_name');`

lasley (changes requested):
Wow this is super long - my eyes are bleeding from this review. I made it to about halfway through the JS. Will pick this back up later today, but figured it was worth posting the comments I already added.

website_version_ce/static/src/js/menu_version.js
```diff
@@ -0,0 +1,294 @@
...
+        ajax.jsonRpc( '/website_version_ce/diff_version', 'call', { 'version_id':version_id}).then(function (result) {
+            var wizardA = $(qweb.render("website_version_ce.publish_message",{message:_.str.sprintf("Publish Version %s", name), list:result}));
```
lasley: jQuery vars should be proceeded with a `$` - `var $wizardA`

website_version_ce/static/src/js/menu_version.js
```diff
@@ -0,0 +1,294 @@
...
+                location.reload();
+            });
+    },
+
+    delete_version: function(event) {
+        var version_id = parseInt($(event.target).closest("li").data("version_id"));
+        var name = $(event.currentTarget).parent().children(':last-child').text();
+        ajax.jsonRpc( '/website_version_ce/check_version', 'call', { 'version_id':version_id }).then(function (result) {
+                if (result){
+                    var wizard = $(qweb.render("website_version_ce.message",{message:_.str.sprintf("You cannot delete the %s version because it is in a running or paused experiment", name)}));
```
lasley: jQuery vars should be proceeded with a `$`

website_version_ce/static/src/js/menu_version.js
```diff
@@ -0,0 +1,294 @@
...
+                    wizard.appendTo($('body')).modal({"keyboard" :true});
+                    wizard.on('hidden.bs.modal', function () {$(this).remove();});
+                }
+                else{
+                    var wizardA = $(qweb.render("website_version_ce.delete_message",{message:_.str.sprintf("Are you sure you want to delete the %s version ?", name)}));
```
lasley: jQuery vars should be proceeded with a `$`

website_version_ce/static/src/js/menu_version.js
```diff
@@ -0,0 +1,294 @@
...
+                    wizardA.appendTo($('body')).modal({"keyboard" :true});
+                    wizardA.on('click','.o_confirm', function(){
+                        ajax.jsonRpc( '/website_version_ce/delete_version', 'call', { 'version_id':version_id }).then(function (result) {
+                            var wizardB = $(qweb.render("website_version_ce.message",{message:_.str.sprintf("The %s version has been deleted.", result)}));
```
lasley: jQuery vars should be proceeded with a `$`

website_version_ce/controllers/main.py
YannickB: I think it's to avoid conflict with the enterprise website_version, but maybe website_version_ce is not the best name. Any suggestion ?

nicolas-petit: It stands for "community edition", and @YannickB is right: it's made to avoid potential conflicts with the enterprise version.
Should we rename it again ? Maybe website_version_oca ?

rafaelbn:
Hi @nicolas-petit , could you please attend @lasley 's comments in order to go on with this PR? Thanks!

website_version_ce/models/experiment.py
nicolas-petit: https://github.com/odoo/odoo/blob/9.0/openerp/models.py#L2017

I cannot change the way odoo defines its internal model functions.

website_version_ce/models/experiment.py
nicolas-petit: https://github.com/odoo/odoo/blob/9.0/openerp/models.py#L2017

I cannot change the way odoo defines its internal model functions.

website_version_ce/controllers/main.py
pedrobaeza: Well, _website_version_oca_ is also not the best name, as is not a version developed by OCA people, but only the Odoo's one that is open source. So I think the suffix `_ce` is appropiated.

lasley: Alright I'm for it, was just making sure we had a good reason.

website_version_ce/models/ir_qweb.py
nicolas-petit: I'm unsure about this one, since that would make the id_or_xml_id assume the "ids" position, which is not what is used with this function: a single id or xml_id is used instead.

Despite searching, I've yet to find an example of redefinition with the new api.
I did find this issue raised on what appears to be an odoo-based project: http://gitlab.opengest.solutions/opengest/odoo/commit/e1deae7eab8be77d7bcb46a68560b86ffbb2277a?view=parallel
The issue suggests that trying to override the function with the new API causes context-related problems.

website_version_ce/models/ir_ui_view.py
nicolas-petit: The version_id is not required since only pages created for and from the website module will have one. Other ir.ui.view items will have none.

website_version_ce/static/src/js/menu_version.js
nicolas-petit: Actually, there is a difference between event.target and event.currentTarget.

Given the stackoverflow link below, and seeing how the listener might be the whole widget (see line 24 of menu_version.js), I'm thinking the uses of target and currentTarget are very intentional.

See here:http://stackoverflow.com/questions/5921413/difference-between-e-target-and-e-currenttarget

website_version_ce/static/src/js/menu_version.js
nicolas-petit: Actually, there is a difference between event.target and event.currentTarget.

Given the stackoverflow link below, and seeing how the listener might be the whole widget (see line 24 of menu_version.js), I'm thinking the uses of target and currentTarget are very intentional.

See here:http://stackoverflow.com/questions/5921413/difference-between-e-target-and-e-currenttarget

lasley (approved):
Changes look good (code no test). Thanks @nicolas-petit

pedrobaeza:
What about the open comments? Can you squash your commits a bit?

lasley:
@OCA/board - this is the versioning module I was speaking of earlier

website_version_ce/views/marketing_view.xml
```diff
@@ -0,0 +1,239 @@
...
+        <record model="ir.ui.view" id="experiment_form_view">
+            <field name="name">experiment.form</field>
+            <field name="model">website_version_ce.experiment</field>
+            <field name="arch" type="xml">
+                <form string="Experiment Form">
+                    <header>
+                        <field name="state" widget="statusbar" statusbar_visible="running,paused,ended" clickable="True"/>
+                    </header>
+                    <sheet string="Exp">
+                        <button icon="fa-line-chart" name="website_version_ce.action_analytic" String="Statistics" action="action"
```
astirpe, line 75: Here `String` is written with capital letter, should be `string` instead...

website_version_ce/models/ir_ui_view.py
```diff
@@ -0,0 +1,180 @@
...
+    def publish(self):
+        """To delete and replace views which are in the website
+        (in fact with website_id)
+        """
+        self.ensure_one()
+        master_record = self.search([
+            ('key', '=', self.key),
+            ('version_id', '=', False),
+            ('website_id', '=', self.website_id.id),
+            ('limit', "=", 1)
```
astirpe, line 59: Why this `('limit', "=", 1)` is added? I think it should be removed.

website_version_ce/models/ir_ui_view.py
```diff
@@ -0,0 +1,180 @@
...
+
+    # Activate or deactivate the right views according to the key
+    @api.multi
+    def toggle(self):
+        """ Switches between enabled and disabled statuses
+        """
+        recs = self.with_context(active_test=False)
+        for view in recs.browse(self.ids):
+            all_id = recs.search([('key', '=', view.key)])
+            for v in recs.browse(all_id):
```
astirpe, line 167: This should be:

```
for v in all_id:

```

website_version_ce/models/ir_ui_view.py
```diff
@@ -0,0 +1,180 @@
...
+                        priority[k.key] = 1
+        return [x for x in arch if x[1] in right_ids.values()]
+
+    # Activate or deactivate the right views according to the key
+    @api.multi
+    def toggle(self):
+        """ Switches between enabled and disabled statuses
+        """
+        recs = self.with_context(active_test=False)
+        for view in recs.browse(self.ids):
```
astirpe, line 165: This should be:

```
for view in recs:

```

website_version_ce/controllers/main.py
```diff
@@ -0,0 +1,219 @@
...
+    def change_version(self, version_id):
+        request.session['version_id'] = version_id
+        return version_id
+
+    @http.route('/website_version_ce/create_version', type='json', auth="user",
+                website=True)
+    def create_version(self, name, version_id=None):
+        if not name:
+            name = fields.Datetime.now().strftime(
+                DEFAULT_SERVER_DATETIME_FORMAT)
```
astirpe, line 27: Remove the `.strftime(...)` because `fields.Datetime.now()` already returns a string

website_version_ce/controllers/main.py
```diff
@@ -0,0 +1,219 @@
...
+    @http.route('/website_version_ce/google_access', type='json', auth="user")
+    def google_authorize(self, **kw):
+        # Check if client_id and client_secret are set for Google
+        gs_obj = request.env['google.service']
+        gm_obj = request.env['google.management']
+
+        client_id = gs_obj.get_client_id(
+            'management', context=kw.get('local_context'))
+        client_secret = gs_obj.get_client_secret(
+            'management', context=kw.get('local_context'))
```
astirpe, line 129: Pass the context by using `.with_context(...)` instead of `context=...`

website_version_ce/controllers/main.py
```diff
@@ -0,0 +1,219 @@
...
+        return name_list
+
+    @http.route('/website_version_ce/google_access', type='json', auth="user")
+    def google_authorize(self, **kw):
+        # Check if client_id and client_secret are set for Google
+        gs_obj = request.env['google.service']
+        gm_obj = request.env['google.management']
+
+        client_id = gs_obj.get_client_id(
+            'management', context=kw.get('local_context'))
```
astirpe, line 127: Pass the context by using `.with_context(...)` instead of `context=...`

website_version_ce/controllers/main.py
```diff
@@ -0,0 +1,219 @@
...
+                'website_version_ce',
+                'action_config_settings_google_management'
+            )
+            return {
+                "status": "need_config_from_admin",
+                "url": '',
+                "action": action
+            }
+        url = gm_obj.authorize_google_uri(from_url=kw.get('from_url'),
+                                          context=kw.get('local_context'))
```
astirpe, line 143: Pass the context by using `.with_context(...)` instead of `context=...`

website_version_ce/models/google_management.py
```diff
@@ -0,0 +1,217 @@
...
+    @api.model
+    def get_token(self):
+        icp = self.env['ir.config_parameter'].sudo()
+        validity = icp.get_param('google_%s_token_validity' % self.STR_SERVICE)
+        token = icp.get_param('google_%s_token' % self.STR_SERVICE)
+        if not (validity and token):
+            raise UserError(_("You must configure your account."))
+        if datetime.strptime(
+                validity.split('.')[0], DEFAULT_SERVER_DATETIME_FORMAT) < \
+                (fields.Datetime.now() + timedelta(minutes=3)):
```
astirpe, line 163: In commit https://github.com/nicolas-petit/website/commit/a9fe09568a44fc5ecc3b7cadc4eeb5425e39e95b you replaced `datetime.now()` with `fields.Datetime.now()`, but the latter returns a string: this will fail when you do `+ timedelta(...)`.

So you need to restore the `datetime.now()` as it was before.

website_version_ce/models/google_management.py
```diff
@@ -0,0 +1,217 @@
...
+    def do_refresh_token(self):
+        gs_pool = self.env['google.service']
+        icp = self.env['ir.config_parameter'].sudo()
+
+        rtoken = icp.get_param('google_%s_rtoken' % self.STR_SERVICE)
+        all_token = gs_pool._refresh_google_token_json(
+            rtoken, self.STR_SERVICE)
+
+        icp.set_param('google_%s_token_validity' % self.STR_SERVICE,
+                      fields.Datetime.now() +
```
astirpe, line 177: Replace `fields.Datetime.now()` with `datetime.now()` as it was before commit https://github.com/nicolas-petit/website/commit/a9fe09568a44fc5ecc3b7cadc4eeb5425e39e95b.

website_version_ce/models/google_management.py
```diff
@@ -0,0 +1,217 @@
...
+    @api.model
+    def set_all_tokens(self, authorization_code):
+        gs_pool = self.env['google.service']
+        all_token = gs_pool._get_google_token_json(
+            authorization_code, self.STR_SERVICE)
+        vals = dict()
+        vals['google_%s_rtoken' % self.STR_SERVICE] = all_token.get(
+            'refresh_token')
+        vals['google_%s_token_validity' % self.STR_SERVICE] = \
+            fields.Datetime.now() + \
```
astirpe, line 206: Replace `fields.Datetime.now()` with `datetime.now()` as it was before commit https://github.com/nicolas-petit/website/commit/a9fe09568a44fc5ecc3b7cadc4eeb5425e39e95b.

website_version_ce/models/google_management.py
```diff
@@ -0,0 +1,217 @@
...
+            timedelta(seconds=all_token.get('expires_in'))
+        vals['google_%s_token' % self.STR_SERVICE] = all_token.get(
+            'access_token')
+        icp = self.env['ir.config_parameter'].sudo()
+        icp.set_param('google_%s_rtoken' % self.STR_SERVICE, all_token.get(
+            'refresh_token'))
+        icp.set_param('google_%s_token_validity' % self.STR_SERVICE,
+                      fields.Datetime.now() + timedelta(
```
astirpe, line 214: Replace `fields.Datetime.now()` with `datetime.now()` as it was before commit https://github.com/nicolas-petit/website/commit/a9fe09568a44fc5ecc3b7cadc4eeb5425e39e95b.

website_version_ce/models/experiment.py
```diff
@@ -0,0 +1,218 @@
...
+            'name': vals['name'],
+            'objectiveMetric': self.env['website_version_ce.goals'].browse(
+                [vals['goal_id']]).google_ref,
+            'status': vals['state'],
+            'variations': [
+                {'name': 'master', 'url': 'http://localhost/master'}]
+        }
+        version_list = vals.get('experiment_version_ids', [])
+        for version in version_list:
+            if version == 0:
```
astirpe, line 138: `if version == 0` is not working here, since the variable `version` is a data structure.

See line https://github.com/odoo/odoo/commit/a11bab12264fc59a6c8f865dc0aa1099ba83c2f8#diff-0ffc8ccafa29d5b9e0575d21597768f2L116

website_version_ce/models/experiment.py
```diff
@@ -0,0 +1,218 @@
...
+    def _get_version_number(self):
+        for exp in self:
+            exp.version_number = len(exp.experiment_version_ids) + 1
+
+    @api.model
+    def create(self, vals):
+        exp = {
+            'name': vals['name'],
+            'objectiveMetric': self.env['website_version_ce.goals'].browse(
+                [vals['goal_id']]).google_ref,
```
astirpe, line 131: Remove external square brackets in the `.browse(...)` params. It should be `.browse(vals['goal_id'])`

website_version_ce/models/google_management.py
```diff
@@ -0,0 +1,217 @@
...
+        account_id = web_property_id.split('-')[1]
+        profile_id = website.google_analytics_view_id
+
+        return (gs_pool, website, web_property_id,
+                action_id, account_id, profile_id)
+
+    @api.model
+    def create_an_experiment(self, data, website_id):
+        (gs_pool, website, web_property_id, action_id, account_id,
+         profile_id) = self._setup_exp_variables(website_id, True)
```
astirpe, line 54: Using `_setup_exp_variables(...)` seem to break the functionality of this method and some other methods below. Could you check?

astirpe (changes requested):
@nicolas-petit thank you for proposing this module.
Please check my code review, most of the comments are about regressions introduced with https://github.com/nicolas-petit/website/commit/8a7164a3995ee35797db95015dbb18aeaa8e70cc

YannickB:
Hello all,

@nicolas-petit doesn't work on Odoo anymore, I strongly suggest that someone take ownership of this pull request. I don't know if there is an easy way to do it while retaining the history of the thread here, maybe we should make an optimistic merge so you can take the lead on this module proposal.

astirpe:
Yes, I agree. Is it possible to merge the job made until now, perhaps with `"installable": False`, as @YannickB suggests?

lasley:
I'll be superseding this PR later today.

website_version_ce/models/google_management.py
lasley: Can you elaborate on what is breaking?

lasley:
Superseded by #362

website_version_ce/models/google_management.py
astirpe: I was experiencing a crash and the traceback was pointing to `_setup_exp_variables`.
Are you able to create an experiment with the updated code? If yes, then you can ignore my previous comment.

lasley: Ummm the module seems broken in a lot of ways right now honestly, but once it compiles I'll give it a shot. Thanks!

PR #243, opened 2016-08-16
[9.0] [MIG] website_anchor_smooth_scroll: Migrate to v9

obulkin:
- Update JS to use v9 module system, including new web_editor hook for determining when a website page is fully loaded
- Improve module interaction with back/forward browser actions by updating browser history before scrolling
- Make small style and clarity changes to various module files

v8 of the module can be found at https://github.com/OCA/website/tree/8.0/website_anchor_smooth_scroll


lasley:
Sent email to cover @obulkin under @laslabs ECLA

website_anchor_smooth_scroll/__openerp__.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Antiun Ingeniería S.L. - Jairo Llopis
+# Copyright 2016 LasLabs Inc.
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Smooth Scroll for Website Anchors',
+    'summary': '''Replace default behavior for internal links (anchor only)
```
lasley: this is going to add a carriage return into the summary. Instead you can break the string, then add a new one below it and it will auto-concat

```
'Replace default behavior for internal links (anchor only)'
'with smooth scroll'
```

website_anchor_smooth_scroll/__openerp__.py
```diff
@@ -0,0 +1,29 @@
...
+     with smooth scroll
+    ''',
+    'version': '9.0.1.0.0',
+    'category': 'Website',
+    'website': 'https://laslabs.com/',
+    'author': '''Antiun Ingeniería S.L., Tecnativa, Odoo Community Association
```
lasley: Same comment re carriage returns. Would be best to just put one per line IMO.

website_anchor_smooth_scroll/demo/website_anchor_smooth_scroll.xml
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+    Copyright 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
+    Copyright 2016 LasLabs Inc.
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+
+<openerp>
```
lasley: New tag is `<odoo>`

website_anchor_smooth_scroll/views/website_anchor_smooth_scroll.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+    Copyright 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
+    Copyright 2016 LasLabs Inc.
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+-->
+
+<openerp>
```
lasley: `<odoo>`

website_anchor_smooth_scroll/static/src/js/website_anchor_smooth_scroll.js
```diff
@@ -0,0 +1,27 @@
...
+    var base = require('web_editor.base');
+
+    var smooth_scroll = function(event) {
+        event.preventDefault();
+        var anchor_fragment = event.target.hash;
+
+        // Do this before scrolling so that browser history accurately reflects scroll position at time of click 
+        history.pushState(null, document.title, anchor_fragment);
+
+        $('html, body').stop().animate({
```
lasley: IMO we should still return a promise from here, even though not needed due to rearrangement of the history push. This makes it so we are not changing the external interface + Promises are incredibly useful with inherited modules.

website_anchor_smooth_scroll/__openerp__.py
```diff
@@ -0,0 +1,29 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+{
+    'name': 'Smooth Scroll for Website Anchors',
+    'summary': '''Replace default behavior for internal links (anchor only)
+     with smooth scroll
+    ''',
+    'version': '9.0.1.0.0',
+    'category': 'Website',
+    'website': 'https://laslabs.com/',
```
lasley: We should leave the original website as attribution to the original authors. LasLabs merely upgraded this module, which is nothing in comparison to creating IMO

lasley:
Thanks @obulkin - comments inline, mainly nitpicking 😉

website_anchor_smooth_scroll/__openerp__.py
obulkin: That's a fair point. I initially tried to include both of them, but that was resulting in a broken link, so I settled for just linking to LasLabs. I'll revert back to the previous address.

website_anchor_smooth_scroll/static/src/js/website_anchor_smooth_scroll.js
obulkin: In v8, outside code could only get access to and make meaningful use of the promise by calling `website_anchor_smooth_scroll()` directly, so the equivalent here would involve exporting `smooth_scroll()` from the module. Is that what you have in mind?

lasley: Yup, which means this also needs to be exported https://github.com/OCA/maintainer-tools/pull/208/files#diff-6d42ed87f53a9be31c0530c5d6ae1e11R18

Best bet would probably be to define this module as `website_anchor_smooth_scroll`, then `return {smoothScroll: smooth_scroll}`

obulkin:
@lasley --- I've addressed your comments and pushed up the changes. Could you take another look?

lasley:
LGTM thanks @obulkin 👍

pedrobaeza:
:+1: It would be interesting to squash the 3 commits of "OCA transbot" in one, and your 2 commits for 9.0 migration also in one.

lasley:
I agree, please squash now that we have our two thumbs @obulkin & also squash the Transifex into one of their own.

obulkin:
I squashed my commits and pushed up the result, but I'm running into various issues when I try to squash the OCA Transbot commits because my changes have all been squashed into the merge commit resulting from the migration instructions (https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0), and the Transbot commits precede the merge. I've tried squashing the Transbot commits, then merging the 8.0 extract into 9.0, and rebasing my branch on top of the result, but this also caused problems that I was unable to work through.

Any advice, @pedrobaeza?

pedrobaeza:
You need to perform the squash on your current branch: Make `git rebase -i HEAD~10` to pull a rewrite of the last 10 commits (there is no need for too many in fact, but for assuring). Then you will see 3 with the text "OCA Transbot updated translations from Transifex". Preserve the first, and squash the rest making something like:

```
...
pick 6aba7a3 OCA Transbot updated translations from Transifex
s 0f6c675 OCA Transbot updated translations from Transifex
s b3ef819 OCA Transbot updated translations from Transifex
...
```

obulkin:
@pedrobaeza --- This is pretty much the first thing I tried. However, my commit doesn't appear in the resulting list, and squashing the Transbot commits results in a branch without my changes. I believe this is because my commit is the merge commit from https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-9.0, amended/squashed together with my changes.

pedrobaeza:
OK, if we have another thumb up or 5 days last, I'll do it when merging. Please ping me when this happens.

lasley:
Thanks @pedrobaeza - I gave this a shot too, and my git-fu failed me :expressionless: 

Call me an idealist, but I wonder if it's possible to make the transifex bot identify if it was the last commit and amend if that is the case.

pedrobaeza:
I'm afraid it can't be, because that amending will screw up all the `git pull` commands around the people that has already download the previous "OCA bot" commit. I'm going to merge this one already making the amend, speeding up a bit the review process to solve this, and if any problem, we can make other PRs fixing it.

pedrobaeza:
Done with a merge+rebase

lasley:
Thanks Git Masta Pedro! Also thanks for the breakdown re: no auto amend w/ Transifex, makes sense in hindsight

pedrobaeza:
Not so fast, there's something wrong with the last commit that screws up the merge operation: or I have the rebase, or I have your changes for v9... so I'm afraid I have to keep the 3 commits.

lasley:
Dang, that's the exact same thing we ran into 😦 

We hypothesized this happened because of the merge commit to bring the module onto the v9. I feel we may have had better luck if the module would have already been on v9 branch, because we wouldn't have had that pesky merge commit.

pedrobaeza:
I already tried that when redacting the migration guide, and it was even cleaner when dealing with modules that don't exist on v9, so I'm puzzled what was the problem. Anyway, let's go ahead and forget these extra 2 commits, which are not too much important.

PR #246, opened 2016-08-31
[IMP][website_seo_redirection] New features and bug fixes.

yajo:
- Fix a bug that was redirecting to disabled languages if the user's browser was in an installed language that was disabled for the website.
- Allow to have several origins for a destination.
- Disallow recursive redirections.
- Add both original and redirected URLs to web editor link search.
- Cannot have a redirection to yourself, now as a SQL constraint.
- Rename "Redirected URL" to "Destination URL" (seems more self-explicative).
- Add support for redirections that don't replace the controller (AKA shortcuts).
- Allow to choose a page URL from the "Promote" menu.
- Fix some styling issues, including `"use strict"` outside the function block.
- Tests, tests, tests (man, those JSONRPC ones were hard).

@Tecnativa

### DIFF
--- a/website_seo_redirection/models/website_seo_redirection.py
+++ b/website_seo_redirection/models/website_seo_redirection.py
@@ -1,10 +1,12 @@
 # -*- coding: utf-8 -*-
 # © 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from psycopg2 import IntegrityError
 
 from openerp import _, api, fields, models
-from openerp.http import request, local_redirect
+from openerp.http import local_redirect, request
 from openerp.exceptions import ValidationError
+
 from ..exceptions import NoOriginError, NoRedirectionError
 
 
@@ -15,7 +17,9 @@ class WebsiteSeoRedirection(models.Model):
     _description = "SEO controller redirections"
     _sql_constraints = [
         ("origin_unique", "UNIQUE(origin)", "Duplicated original URL"),
-        ("destination_unique", "UNIQUE(destination)", "Duplicated new URL"),
+        ("origin_destination_distinct",
+         "CHECK(origin != destination)",
+         "Recursive redirection."),
     ]
 
     origin = fields.Char(
@@ -25,10 +29,18 @@ class WebsiteSeoRedirection(models.Model):
         help="Path where the original controller was found.",
     )
     destination = fields.Char(
-        string="Redirected URL",
+        string="Destination URL",
         required=True,
         index=True,
         help="Path where the controller will be found now.",
+    )
+    relocate_controller = fields.Boolean(
+        default=True,
+        help="If you relocate the controller, the same page that was found in "
+             "the original URL will be now in the destination. Otherwise, "
+             "this will simply make requests be redirected, but a new "
+             "controller should be available in the destination URL if you "
+             "do not want a 404 error.",
     )
 
     @api.multi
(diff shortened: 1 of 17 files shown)


yajo:
Note to self: push PR to v9 when merged.

website_seo_redirection/models/website_seo_redirection.py
```diff
@@ -15,7 +17,9 @@ class WebsiteSeoRedirection(models.Model):
     _description = "SEO controller redirections"
     _sql_constraints = [
         ("origin_unique", "UNIQUE(origin)", "Duplicated original URL"),
-        ("destination_unique", "UNIQUE(destination)", "Duplicated new URL"),
+        ("origin_destination_distinct",
+         "CHECK(origin != destination)",
```
lasley, line 21: ~~Hmmm what about a cyclical recursion? Redirect A points to B which points to C, which points back at A? Out of scope?~~ this question was answered in like 20 lines

website_seo_redirection/models/website_seo_redirection.py
```diff
@@ -42,6 +54,21 @@ def _check_destination(self):
...
+    @api.constrains("origin", "destination")
+    def _check_not_recursive(self):
+        """Avoid infinite loops."""
+        for redirection in self:
+            origins = {redirection.origin}
+            while redirection:
+                redirection = self.search([
+                    ("origin", "=", redirection.destination),
+                ])
+                if redirection.origin in origins:
```
lasley, line 66: This assumes one record, but there isn't a limit in the search

website_seo_redirection/models/website_seo_redirection.py
```diff
@@ -42,6 +54,21 @@ def _check_destination(self):
...
+                    raise ValidationError(
+                        _("Recursive redirection is forbidden."))
+                origins.add(redirection.origin)
```
lasley, line 69: Won't this fail on the last iteration? Looking at the while loop, it will terminate when not `redirection` which would mean that at some point we're trying a `.origin` on an empty Recordset, `False` or `None`?

website_seo_redirection/models/website_seo_redirection.py
yajo: [It's impossible.](https://github.com/OCA/website/pull/246/files/6b9a8e5494aa80e8897c5cc8b22e737ca93d5552#diff-d011fe32656e50bc9cf8dc81716ac66aR19)

website_seo_redirection/models/website_seo_redirection.py
yajo: You got it, an empty recordset evals to False, which exits the loop. Since there's nothing left to do, adding `False` to `origins` won't hurt.

lasley: But the loop won't terminate until after `EmptyRecordset().origin`, which would be yielding an `AttributeError` because not singleton?

yajo: No, emtpy recordset fields return `False` always.

It would raise `AttributeError` if I asked for `redirection.trololo`

lasley: Oh snap TIL! I've been adding so many unnecessary guards 😑

lasley:
Awesome features @Yajo - Travis seems to be failing but 👍  once green

yajo:
Not sure what is failing, it works in localhost, I'll add some logging to inspect. BTW I just realized I forgot to activate the promote menu stuff...

lasley:
I didn't notice missing promote menu during functional tests, whoops. Maybe add a note in ReadMe about promote menu as well?

yajo:
It should work now.

pedrobaeza:
:+1:

PR #250, opened 2016-09-04
[9.0][MIG] website_backend_views to v9

lasley:
This upgrades website_backend_views to v9.

It should work, although I haven't tested much. This is a WIP because with the fix I found in OCA/web#402 for [public data sessions](https://github.com/OCA/web/pull/402/commits/33ee3ae91e66e687af7cab6195383ffd2ac2456d), I believe this is the key for OCA/web#403 instead of the standalone session module.

### DIFF
--- a/website_backend_views/__init__.py
+++ b/website_backend_views/__init__.py
@@ -1,22 +1,6 @@
 # -*- coding: utf-8 -*-
-##############################################################################
-#
-#    OpenERP, Open Source Management Solution
-#    This module copyright (C) 2015 Therp BV <http://therp.nl>.
-#
-#    This program is free software: you can redistribute it and/or modify
-#    it under the terms of the GNU Affero General Public License as
-#    published by the Free Software Foundation, either version 3 of the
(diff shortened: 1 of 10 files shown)


simahawk:
@lasley sorry, just to clarify: do I need to wait for the linked issues to be closed or can I test the porting right away? tnx

lasley:
@simahawk - I just pushed the version that I have been functionally testing before implementing the public session. This should work in v9 to the same capacity that the v8 one did, and I would most certainly love additional confirmation 😄

simahawk:
`website.assets_frontend.js:836 Uncaught TypeError: openerp.init is not a function` this is what I get :(
The whole error in debug mode is [this](https://www.dropbox.com/s/vdfiesb2te5ll2h/bkend-views-err.png?dl=0).

simahawk:
closing as it's on 9.0 and no action since more than 2 yrs and @lasley left us a long time ago :sob: 
Feel free to re-open it if anyone needs it, thanks!

PR #251, opened 2016-09-05
[IMP][website_slides] Update backport.

yajo:
Get lots of FIX and IMP from upstream since the backport was done.

I know Travis won't like long Python lines from Odoo, but that detail was what took me most **hours** because last backport fixed that, so I would like to know if there's any way to tell flake8 not to check that in all this backport, to make further updates easier. Any ideas?

@Tecnativa


pedrobaeza:
Can you check CIs status and squash all the commits from the translation agent?

rafaelbn (approved):
Reviewed, thanks

pedrobaeza (approved):

PR #254, opened 2016-09-08
[IMP][website_seo_redirection] New features and bug fixes. (#246)

yajo:
Forward #246 and #255 to v9.
- Fix a bug that was redirecting to disabled languages if the user's browser was in an installed language that was disabled for the website.
- Allow to have several origins for a destination.
- Disallow recursive redirections.
- Add both original and redirected URLs to web editor link search.
- Cannot have a redirection to yourself, now as a SQL constraint.
- Rename "Redirected URL" to "Destination URL" (seems more self-explicative).
- Add support for redirections that don't replace the controller (AKA shortcuts).
- Allow to choose a page URL from the "Promote" menu.
- Fix some styling issues, including `"use strict"` outside the function block.
- Tests, tests, tests (man, those JSONRPC ones were hard).

@Tecnativa

### DIFF
--- a/website_seo_redirection/README.rst
+++ b/website_seo_redirection/README.rst
@@ -12,6 +12,9 @@ URLs and allow you to improve the SEO.
 Usage
 =====
 
+Moving a page to another URL
+----------------------------
+
 Let's assume you created a blog post entry that is absolutely amazing and you
 want a fixed and beautiful URL for it. That entry is posted in
 ``https://example.com/blog/our-news-1/post/amazing-post-23``, but you want it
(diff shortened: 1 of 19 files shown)


pedrobaeza:
With the renaming capability of v9, it's still needed all the part of changing the URL?

yajo:
I did not try, but probably, if you don't want to lose SEO score, yes.

rafaelbn:
Hi @Yajo , travis fails

rafaelbn (changes requested):
Hi @Yajo , please solve branch conflicts and review travis. Thanks

pedrobaeza:
Can you rebase this and check my question, @Yajo?

yajo:
Let's see what does the bot say.

yajo:
BTW, about https://github.com/OCA/website/pull/254#issuecomment-245559390, if you use the renaming function, you get a 404 error in old URL, while with this you get a 301 redirection. I think it still has a meaning for SEO freaks.

yajo:
Well, bots are :white_check_mark:, but it's fake. They are affected by https://github.com/odoo/odoo/pull/13743. It seems Travis has no `sass` executable. Who should I ping for the fix?

pedrobaeza:
Not really fake, but not "covered" due to an Odoo bug. You can ping to the JS lead developer you met at Odoo experience to see, but we can't block this due to that question, or should we?

yajo:
It is fake because they should go :x:; that developer is already pinged in the issue.

However, [tests for this addon are passing](https://travis-ci.org/OCA/website/jobs/166716308#L1157), so no need to block this PR.

pedrobaeza (approved):
Hi, that was what I want to hear. As the problem stated by @rafaelbn is fixed, I merge.

PR #257, opened 2016-09-21
8.0 group old archives

gfcapalbo:
Modifies the archive, allows grouping per month for the current year,  groups by year for older posts.
Usefull for those who have many years of posts and don't want a very long archive list


website_blog_group_old_archives/controllers/website_blog.py
```diff
@@ -0,0 +1,75 @@
...
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+import datetime
+from openerp import http, tools
+from openerp.http import request
+from openerp.addons.website_blog.controllers.main import WebsiteBlog
+
+class website_blog(WebsiteBlog):
+
```
hbrunn: too many blank lines

website_blog_group_old_archives/controllers/website_blog.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def nav_list_grouped(self):
+        blog_post_obj = request.registry['blog.post']
```
hbrunn: no need to change this now, but the `request` object has a perfectly good property called `env`

website_blog_group_old_archives/controllers/website_blog.py
```diff
@@ -0,0 +1,75 @@
...
+        new_groups = blog_post_obj.read_group(
+            request.cr, request.uid,
+            [['create_date', '>=', beginning_of_year]],
+            ['name', 'create_date'],
+            groupby="create_date", 
+            orderby="create_date desc", 
+            context=request.context
+        )
+        for group in new_groups:
+            begin_date = datetime.datetime.strptime(
```
hbrunn: all of those can be simplified a lot by using `openerp.fields.Date{,time}.from_string`

website_blog_group_old_archives/controllers/website_blog.py
```diff
@@ -0,0 +1,75 @@
...
+        return {'new_groups': new_groups, 'old_groups': old_groups}
+
+    @http.route()
+    def blog(self, blog=None, tag=None, page=1, **opt):
+        result = super(website_blog, self).blog(
+                blog=blog, tag=tag, page=page, **opt
+            )
+        result.qcontext['nav_list_old_grouped'] = self.nav_list_grouped()['old_groups']
+        result.qcontext['nav_list_new_grouped'] = self.nav_list_grouped()['new_groups']
+	return result
```
hbrunn: this should break miserably

website_blog_group_old_archives/static/src/js/website_blog_group_old_archives.js
```diff
@@ -0,0 +1,7 @@
+//-*- coding: utf-8 -*-
+//© 2016 Therp BV <http://therp.nl>
+//License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+openerp.website_blog_group_old_archives = function(instance)
+{
+}
```
hbrunn, line 7: remove unneeded files

website_blog_group_old_archives/views/templates.xml
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<openerp>
+    <data>
+
+
+
```
hbrunn: remove blank lines

website_blog_group_old_archives/views/templates.xml
```diff
@@ -0,0 +1,34 @@
...
+        <template id="opt_blog_rc_history_group_old_archive"                                                        inherit_id="website_blog.opt_blog_rc_history">
```
hbrunn: remove extra spaces

hbrunn (changes requested):

hbrunn:
and fix https://travis-ci.org/OCA/website/jobs/161662840#L472

website_blog_group_old_archives/controllers/website_blog.py
gfcapalbo: @hbrunn why?  it is tested. i see an indent that isn't there in the code. do you mean that?

hbrunn: aha! this line uses a tab instead of spaces, that's why it looks weird here. Why this works eludes me to be honest, but I also can't find a definite source on how the python interpreter handles mixed spaces and tabs. Then it's only a violation of the rule to use spaces instead of tabs

gfcapalbo: i thought i had a vim plugin for that. evidently i dont. fixing

pedrobaeza:
As we avoid to put plural names on modules (https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#modules), I suggest you to call this module `website_blog_old_archive_grouping`.

PR #258, opened 2016-09-21
8.0 website blog tags variable dimension

gfcapalbo:
Extends the tag cloud availiable in website_blog to support variable dimensions of the tags, based on the number of blogs they are contained in.


website_blog_tags_variable_dimension/README.rst
```diff
@@ -0,0 +1,61 @@
...
+
+
+For further information, please visit:
+
+* https://www.odoo.com/forum/help-1
+
+Known issues / Roadmap
+======================
+
+* ...
```
hbrunn: remove empty sections

website_blog_tags_variable_dimension/README.rst
```diff
@@ -0,0 +1,61 @@
...
+
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* gfcapalbo <giovanni@therp.nl>
```
hbrunn: you might want to change this

website_blog_tags_variable_dimension/models/blog_tag.py
```diff
@@ -0,0 +1,87 @@
...
+        no blogposts with tags.
+        """
+        max_ranking = 1
+        for tag in tags:
+            max_ranking = max(
+                len(tag.post_ids), 
+                max_ranking
+            )
+        for this in self:
+            this.max_ranking = float(max_ranking)
```
hbrunn, line 29: replace the loop by self.update()
But as this is probably called very often and we hopefully will have tens of thousands of posts in a while, I'd propose to make this field stored anyways

website_blog_tags_variable_dimension/models/blog_tag.py
```diff
@@ -0,0 +1,87 @@
...
+
+    @api.multi
+    def get_font_size(self):
+        for this in self:
+            hits = len(this.post_ids.ids)
+            this.font_size_in_cloud = int(
+                (hits)/(this.max_ranking) * (
+                    this.max_font - this.min_font
+                ) + this.min_font
+            )
```
hbrunn: missing blank line

website_blog_tags_variable_dimension/models/blog_tag.py
```diff
@@ -0,0 +1,87 @@
...
+            hits = len(this.post_ids.ids)
+            this.font_size_in_cloud = int(
+                (hits)/(this.max_ranking) * (
+                    this.max_font - this.min_font
+                ) + this.min_font
+            ) 
+    @api.multi
+    def get_tag_cardinality(self):
+        for this in self:
+            this.tag_cardinality = len(this.post_ids.ids)
```
hbrunn, line 51: also a case for a stored field

website_blog_tags_variable_dimension/models/blog_tag.py
```diff
@@ -0,0 +1,87 @@
...
+
+    @api.multi
+    def compute_placement(self):
+        for this in self:
+            this.random_placement = random.choice([x for x in range(0, 4000)])
```
hbrunn: this is fantastically inefficient. Compare

``` bash
$ python -m timeit 'import random; random.choice([x for x in range(0, 4000)])'
10000 loops, best of 3: 112 usec per loop
$ python -m timeit 'import random; random.randint(0, 4000)'
1000000 loops, best of 3: 1.07 usec per loop
```

and use https://docs.python.org/2/library/random.html#random.randint

website_blog_tags_variable_dimension/static/src/js/website_blog_tags_variable_dimension.js
```diff
@@ -0,0 +1,7 @@
+//-*- coding: utf-8 -*-
+//© 2016 Therp BV <http://therp.nl>
+//License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+openerp.website_blog_tags_variable_dimension = function(instance)
+{
+}
```
hbrunn: remove empty files (also the css file above)

website_blog_tags_variable_dimension/views/templates.xml
```diff
@@ -0,0 +1,41 @@
...
+            </xpath>  
+        </template>
+
+
+
+        <template id="opt_blog_rc_tags_var_dim" name="Variable Dimension Tag Cloud"
+            inherit_id="website_blog.index_right" active="False" customize_show="True">
+            <!-- note keep this tag cloud allways on top  -->
+            <xpath expr="//div[@id='blog_right_column']/section[1]" position="before">
+                        <section class="mt32">
```
hbrunn: watch your indentation

website_blog_tags_variable_dimension/models/blog_tag.py
```diff
@@ -0,0 +1,87 @@
...
+        compute=compute_placement,      
+    )
+
+    random_color = fields.Char(
+        string="random color of tag in cloud",
+        compute=compute_color, 
+    )
+
+    def randomize_recordset(self):
+        return sorted(self, key=lambda x: x.random_placement)
```
hbrunn: I think you're better off replacing this by https://docs.python.org/2/library/random.html#random.sample (k = len(self)) and remove the function fields about randomness. As you only need them in your view, you're better off doing this in some function you call - this saves the field related overhead you don't need here

hbrunn (changes requested):
and check https://travis-ci.org/OCA/website/jobs/161664244#L472

website_blog_tags_variable_dimension/models/blog_tag.py
gfcapalbo: that means that al random selections even non integer ones should be done by hashtable that associates ints to values +randint  to harness the x10 factor improvement. i used random.sample for this and removed the whole field.

gfcapalbo: x100

rafaelbn:
Please @gfcapalbo could you check travis? thanks

website_blog_tags_variable_dimension/models/blog_tag.py
```diff
@@ -0,0 +1,83 @@
+# -*- coding: utf-8 -*-
+# © 2016 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+from openerp import api, fields, models
+import random
```
oscarolar, line 5: import first standard imports [guideline](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#imports)

oscarolar:
what is the point of having this empty file?:
`website_blog_tags_variable_dimension/static/src/css/website_blog_tags_variable_dimension.css`

website_blog_tags_variable_dimension/models/blog_tag.py
```diff
@@ -0,0 +1,83 @@
...
+class BlogTag(models.Model):
+    _inherit = 'blog.tag'
+
+    # doing it for all blogs. need to make it per-blog.
+    min_font = 10.0
+    max_font = 19.0
+
+    @api.multi
+    def get_blog_most_used_tag_cardinality(self):
+        tags = self.env['blog.tag'].search([])
```
susport, line 16: Replace this code:
tags = self.env['blog.tag'].search([])
for tag in tags:

Suggestion:
for tag in self:

website_blog_tags_variable_dimension/models/blog_tag.py
```diff
@@ -0,0 +1,83 @@
...
+                ) + this.min_font
+            )
+
+    @api.multi
+    def get_tag_cardinality(self):
+        for this in self:
+            this.tag_cardinality = len(this.post_ids.ids)
+
+    @api.multi
+    def compute_color(self):
```
susport, line 54: The compute methods start it with  _.
Example:  def _compute_color (self):

website_blog_tags_variable_dimension/models/blog_tag.py
gfcapalbo: @susport, attention, this is a particular function.  "self" could be  any recordset , I need to actually scan all tags. this function has to return max cardinality amongst all records.. so unfortunately the more elegant  expression is not possible, the search needs to happen.

website_blog_tags_variable_dimension/models/blog_tag.py
gfcapalbo: is this oca guideline?

PR #260, opened 2016-09-30
[8.0][IMP][website_canonical_url] Safer, prefer root, relative paths

yajo:
This patch makes this addon more SEO-friendly:
- Includes a safety fix to avoid possible `KeyError` when no lang in context (could this ever happen? not sure, but preventing is easier).
- Uses relative paths everywhere, so this works fine under any setup (HTTPS proxies, anyone?).
- In case you visit `/`, and that gets rerouted to your homepage (`/page/homepage` by default), then we canonicalize `/` instead. This is better because it fits more what the user would type in the URL bar, and removes stop words from the indexed URL.

@Tecnativa

### DIFF
--- a/website_canonical_url/__openerp__.py
+++ b/website_canonical_url/__openerp__.py
@@ -4,13 +4,11 @@
 
 {
     'name': "Website Canoncial URL",
-    'summary': """
-        Canonical URL in Website Headers
-        """,
-    'author': "initOS GmbH, Odoo Community Association (OCA)",
+    'summary': "Canonical URL in Website Headers",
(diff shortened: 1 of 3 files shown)


website_canonical_url/__openerp__.py
```diff
@@ -10,7 +10,7 @@
     'author': "initOS GmbH, Odoo Community Association (OCA)",
```
pedrobaeza: Include `Tecnativa` here

rafaelbn (approved):
Thanks

pedrobaeza (approved):

PR #269, opened 2016-10-22
[MIG][10.0] website_field_autocomplete

lasley:
- Bump version
- Change openerp namespace to odoo
- Add OCA icon
- Remove jQuery require

PR #270, opened 2016-10-22
[MIG][10.0] website_field_autocomplete_related

lasley:
This is a forward port of #217. It's a vertical-medical website dependency, so is a blocker to my upgrades there - hence the need to port before the v9 is merged.
- Rename manifest file
- Bump version
- Change openerp namespace to odoo
- Remove jQuery require

Dependent on:
- [x] #269

### DIFF
--- a/website_field_autocomplete_related/controllers/main.py
+++ b/website_field_autocomplete_related/controllers/main.py
@@ -2,8 +2,8 @@
 # Copyright 2016 LasLabs Inc.
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
 
-from openerp.http import request
-from openerp.addons.website_field_autocomplete.controllers.main import Website
+from odoo.http import request
+from odoo.addons.website_field_autocomplete.controllers.main import Website
 
 
 class WebsiteAutocomplete(Website):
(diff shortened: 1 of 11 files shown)


OsoTranquilo (approved):

HaTry (approved):
Work OK

Tardo:
@lasley The module 'website_field_autocomplete' doesn't work fine... if you select an option using arrow keys you see the ID instead text...
And.. why use jQuery Autocomplete and not Select2?

website_field_autocomplete_related/controllers/main.py
```diff
@@ -0,0 +1,32 @@
...
+class WebsiteAutocomplete(Website):
+
+    def _get_autocomplete_data(self, model, domain, fields, limit=None):
+        """ Perform recursive search_reads to get all related data """
+        res = super(WebsiteAutocomplete, self)._get_autocomplete_data(
+            model, domain, fields, limit,
+        )
+        self.record_ids = request.env[model].search(domain)
+        for field in fields:
+            if '.' in field:
```
andriusdaunoravicius: Since you have no else clause, you could filter fields and no if would be required

andriusdaunoravicius (approved):

lasley:
Oops sorry I missed the comments on this one!

@Tardo - Good catch, I've pushed a fix for the ID issue. 

Honestly I just didn't know select2 was implemented in Odoo when I made this, so I went with jQuery UI. I will likely switch that up in v11 if we still need it.

@andrius-preimantas - filter added

simahawk:
outdated and David is not working on Odoo anymore

PR #280, opened 2016-11-14
[FIX][10.0] Fixing ssl break and redirect for multilanguage websites

lasley:
v10 forward port of @PCatinean's #276


pedrobaeza (approved):
As this has been already reviewed in v8, we can fast-track it

PR #281, opened 2016-11-15
[FIX][REF][IMP][website_event_register_free_with_sale] Less requirements for free events.

yajo:
Before this patch, there was a bug.

1. Get some free tickets for an event:

   

2. Press "Order Now".
3. Leave checkout form blank, and press "Confirm".
4. Odoo required too many fields for free events:

   

I had to refactor it to fix it, because workflow was a bit blurry. Now it reuses more core Odoo facilities, and workflow is more straight-forward and better documented inline.

Also, when you were buying both free and non-free tickets, no confirmation was being sent to the user about the free part. This also is improved.

This PR also affects `website_event_register_free` because it is a requirement to make this work. Changes there are minimal, so it is included here instead of opening 2 PR.

@Tecnativa


pedrobaeza:
Please check Travis

pedrobaeza:
In v9, this should be moved to OCA/event

yajo:
Travis fails because there is an old api method in `website_event_register_free`. What comment should I put where to shut it up?

pedrobaeza (changes requested):
Two comments:
- The text that appears in price in test I have made is 0.00 €, not Free.
- When registering in a free event, I get as mandatory also the VAT.

yajo:
Did you try removing `website_sale_vat_required`?

pedrobaeza:
No, I didn't install it. That can be the problem.

yajo:
The "Free" problem I think can be because there are both free and non-free tickets in the event. Is that it?

pedrobaeza:
The error is not with old API, but:

```
website_event_register_free_with_sale/controllers/website_sale.py:23:1: F401 'openerp.addons.website_event_register_free.controllers.website_event.WebsiteEvent' imported but unused
```

pedrobaeza (approved):
When the lint error is fixed.

rafaelbn:
Hi, tested in runbot. 1 fix + 1 imp

I created 3 events:

1 - one without any tickets
2 - one with 1 ticket with price 0 euros
3 - one with a mix, 1 ticket with price 0 euros and a second ticket with 10 euros

Case 1:
- Fails as Odoo doesn't let to to register (I tested also with  website_sale_vat_required uninstalled, same result)
- In Odoo if you press "Register" with all fields empy it put all of the mandatory ones in RED, but in this case only shows an advice of the first field.

Check image bellow

Case 2 (included in the image) is OK



Case 3 is OK

yajo:
I only patched cases 2 and 3. I personally think case 1 is a conceptual failure of this addon, I'd be glad to fix that. Why am I able to register to events without tickets? Makes no sense, just add free tickets and then you get registration capabilities, don't you think?

pedrobaeza:
I also think that case 1 shouldn't be covered for avoiding a lot of code for a case that can easily be covered the other way.

yajo:
Should I take this chance to remove support for case 1, or do we let this for migration?

rafaelbn:
Remove support for case 1 then @Yajo

yajo:
it turns out [case 1 is exactly what the whole `website_event_register_free` is about](https://github.com/OCA/website/pull/281/files#diff-8acf543fe46c4dc59e7acf26d82038faR4)... :hankey: 

So, summarizing:
- `website_event_register_free` was born to avoid installing `website_event_sale` when you plan to have only free events.
- `website_event_sale` was born to allow you to sell tickets for events, be it for free or not. Only problem is if amount is zero, a zero-amount SO is created.
- `website_event_register_free` was born to allow you to have both installed without colliding.

Seems mostly like this:

modern software development

I guess I'll keep this as it is, but for next migration structure should be:
- `website_event_sale`, to sell tickets for events.
- `website_event_sale_free`, to skip the SO creation when its amount is zero because it is composed solely of free tickets.

Do you agree?

pedrobaeza:
No, the structure is correct, because you don't want the heavy website_sale when you only have free events.

yajo:
You are not installing `website_sale`, only `website_event_sale`. It's different: sell tickets.

hbrunn:
and guess what `website_event_sale` depends on: https://github.com/OCA/OCB/blob/8.0/addons/website_event_sale/__openerp__.py#L15

yajo:
I see...

Well, I'm more on the side of not reinventing things just to avoid dependencies. If dependencies come in, you can simply remove the "shop" link from your menu, or disable the "/shop" controller from your proxy, or you could even use [`website_seo_redirection`](https://github.com/OCA/website/tree/8.0/website_seo_redirection) to redirect `/shop` calls to `/event`..., and avoid maintaining these confusing addons.

You know... creating addon A to do the same as addon B but without installing addon C, and then addon D to disable addon A's most parts because addon B already makes that and it comes out you want both... It's a bit absurd honestly :confounded: :question:

pedrobaeza:
When planning a long-term strategy, maintaining the migration of that modules is a pain. That's why we prefer to avoid that dependency in this simple case. What we can do is to not depend on website_event_register_free when having also events with price.

yajo:
`website_sale` is a standard addon. If they really don't use it, then no data would need to be migrated, and if they do, then standard migration process should be enough. It's quite common that other of your customers have that installed.

I guess it's a matter of what you understand by "maintainability": maintaining mostly-useless addons, or maintaining addons that users don't use. Both have pros and cons...

> What we can do is to not depend on website_event_register_free when having also events with price

In any case, this is not possible :confused:. You need to depend on that to disable it, or you'd have 2 registration forms.

pedrobaeza:
I'm talking about having all the needed logic only on the second one. Obviously, both modules will be incompatible that way.

yajo:
My proposal is to dramatically change structure and drop support for selling tickets without installing the addon that actually was made to sell tickets, but IMHO, if you still want to maintain this structure, then it's best as it is now. Otherwise you'd introduce intermodule incompatibility and have to repeat code between those 2.

pedrobaeza:
Let's go then for this structure...

rafaelbn:
Travis is red

rafaelbn (approved):
Thanks! Much more better :-)

PR #283, opened 2016-11-15
[FIX][9.0] travis: Add SASS

lasley:
Possible fix to #282 

Question still stands as to why this issue just popped up, and why it's not showing on v8 or v10.

Issue #282, opened 2016-11-15
[9.0] Travis failing on SASS execution


lasley:
First reported in #279, also noticed in #277. I rebuilt 9.0 branch on Travis, and it is now failing in the same way: https://travis-ci.org/OCA/website/builds/167816895

```
2016-11-15 17:32:13,517 5314 ERROR openerp_test openerp.addons.base.ir.ir_qweb: Could not execute command 'sass'
```

pedrobaeza:
We have faced the same issue and tried to fix it with https://github.com/OCA/website/pull/277/files#diff-354f30a63fb0907d4ad57269548329e3R32

lasley:
You're still getting SASS issues there too though (albeit not as bad):

```
2016-11-15 11:32:27,007 7442 WARNING openerp_test openerp.addons.base.ir.ir_qweb: LoadError: cannot load such file -- bootstrap-sass
```

Do we have any idea what the SASS is actually being used for? I'm confused why bootstrap-sass is even a thing; Odoo uses bootstrap-less 😕

pedrobaeza:
AFAIK they used both. Then we also need to install another thing more...

lasley:
Closing, fixed in #284

PR #284, opened 2016-11-15
[FIX][9.0] travis: Add SASS

lasley:
Possible fix to #282 and resubmit of #283, which was to the wrong branch.

Question still stands as to why this issue just popped up, and why it's not showing on v8 or v10.


lasley:
Crap Travis is [building this on 10.0 too](https://travis-ci.org/OCA/website/jobs/176130522#L179). I'm now incredibly confused - The travis file [clearly says v9](https://github.com/laslabs/website/blob/5980601aa721c54a4970282cebef17f2925ba127/.travis.yml#L17).

lasley:
Amending the commit did the trick. I'll keep in mind for future. Now back to troubleshooting the real issue

lasley:
Alright this should be good assuming build passes. I'm still baffled why this error appeared out of nowhere, but I needed to update the Ruby version in Travis to support SASS, then add `bootstrap-sass` and `compass` as gem depends.

After that, I resolved the test issue in `_logo` and removed the v8 migrations.

lasley:
Now we're ping-ponging CIs here. Runbot is failing due to the use of the RVM command, which is by far the best way to upgrade Ruby on Travis. Version <2 Ruby is deprecated for `sass` and outright not supported for `compass`.

We'll need to add this either to Travis2Docker or to the Docker base image for Runbot. I think the image, but I'll wait for response from Vauxoo in [associated issue](https://github.com/Vauxoo/travis2docker/issues/74) before randomly hacking about.

yajo:
Shouldn't we be using t2d at this point?

pedrobaeza:
Travis2docker is not activated yet for all repositories. I don't know if website is one of them.

lasley:
Website is one of them, and the issue is specific to T2D - although we would have the same issue if stock Runbot. I'm working on fixing the base Docker image, and should have something submitted by EOD today

yajo:
Seems like runbot went :x:

lasley:
Runbot 🔴  was fixed in:
- https://github.com/Vauxoo/travis2docker/pull/77
- https://github.com/Vauxoo/docker-odoo-image/issues/165

We just need to wait for updates noted in https://github.com/OCA/runbot-addons/issues/105#issuecomment-262144364 then a rebuild will fix.

moylop260 (approved):
Using the last changes this pr work fine!

screen shot 2016-11-21 at 5 48 41 pm

(I'll add small comments non-blocking)

website_logo/controllers/main.py
```diff
@@ -80,5 +80,5 @@ def website_logo(self, dbname=None, **kw):
                             image, filename=imgname, mtime=mtime)
                         return response
             except Exception:
-                return http.send_file(placeholder(imgname))
+                pass
```
moylop260: Using [pep-0020](https://www.python.org/dev/peps/pep-0020):
 - Errors should never pass silently.

If there’s really nothing to do add a `_logger.error`

lasley: Hmmm guess I need to re-read that one, but I agree this was bad. `.error` it is, there is indeed nothing to do because the end return handles it

lasley:
Alright Runbot fixed, and green like @moylop260 noted. This is good for a merge I believe

pedrobaeza (approved):

moylop260:
🎉 🎈

PR #291, opened 2016-11-28
[9.0] [MIG] website_blog_share module

cubells:
* Updated README.rst

@Tecnativa


website_blog_share/__openerp__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Copyright 2016 Andhitia Rama <andhitia.r@gmail.com>
+# Copyright 2016 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Blog Share",
+    "version": "9.0.1.0.0",
+    "author": "Tecnativa,"
```
andhit-r: @cubells do not erase me from author :smile:

cubells:
Sorry @andhit-r 

Check if all is right now.

andhit-r:
No worries @cubells. Tested functionally on runbot :+1:

pedrobaeza (approved):

website_blog_share/views/website_blog.xml
```diff
@@ -0,0 +1,16 @@
...
+<odoo>
+    <template id="blog_share" name="Share Button" inherit_id="website_blog.blog_post_complete" active="False" customize_show="True">
+        <xpath expr="//h2[@t-field='blog_post.subtitle']" position="after">
+            <div class="oe_share s_share">
+                  <h3>Share
+                    <a class="oe_share_facebook" t-attf-href="https://www.facebook.com/sharer/sharer.php?u={url}" target="_Blank"><i class="fa fa-facebook-square"/> </a>
+                    <a class="oe_share_twitter" t-attf-href="https://twitter.com/intent/tweet?text={{blog.website_meta_title}}&amp;url={url}" target="_Blank"><i class="fa fa-twitter"/> </a>
+                    <a class="oe_share_linkedin" t-attf-href="http://www.linkedin.com/shareArticle?mini=true&amp;url={url}&amp;title={{blog.website_meta_title}}" target="_Blank"><i class="fa fa-linkedin"/> </a>
+                    <a class="oe_share_google" t-attf-href="https://plus.google.com/share?url={url}" target="_Blank"><i class="fa fa-google-plus-square"/> </a>
+                    <a href="mailto:?body={url}&amp;subject={title}"><i class="fa fa-envelope-o"> </i></a>
```
yajo, line 11: It would be nice to add [whatsapp](https://www.whatsapp.com/faq/en/android/28000012) and [telegram](http://telegram.wiki/tips:websharing) sharing.

yajo (approved):
Code OK. Functional not possible due to dead runbot, but I guess it should work fine.

website_blog_share/views/website_blog.xml
pedrobaeza: But that shouldn't be static... How do you tell your phone to share the content and make appear the share dialog? Is there a standard HTML way?

pedrobaeza:
@cubells, please squash Andithia's commits together, as they are meaningless.

website_blog_share/views/website_blog.xml
yajo: well, all others are hardcoded. No standard so far, although all browsers have the option nowadays.

pedrobaeza: Yeah, they are hardcoded, but I thought that maybe there was a way to handle it. WhatsApp link will only work on mobile with it installed, so I see it as not a good choice. The other can be an addition indeed. I'm thinking that it would best to have a general website_share module (or even use any from standard - does it exists?), and that module can be improved for adding services. Of course, there's the need of allowing to configure in which services to share.

yajo: yes, that would be great. Maybe we can add all this to roadmap to not block this migration.

pedrobaeza: Yes, of course this is not blocking. They were just some thoughts.

pedrobaeza:
Merging as the error is not in this module.

Issue #262, opened 2016-10-06
Migration to version 10.0


pedrobaeza:
# Todo

https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-10.0
# Modules to migrate
- [x] website_analytics_piwik - By @astirpe - #338 
- [x] website_anchor_smooth_scroll - By @Tardo #347 
- [x] website_canonical_url - By @simahawk - #449 
- [ ] website_backend_views
- [ ] website_blog_facebook_comment
- [ ] website_blog_mgmt
- [x] website_breadcrumb - By @JonathanNEMRY - #339 
- [x] website_container_fluid - By @Tardo #348
- [x] website_cookie_notice - By @hurrinico - #309 
- [x] website_crm_quick_answer - By @chienandalu - #356 
- [x] website_crm_privacy_policy - By @Yajo #359 
- [ ] website_crm_recaptcha - By @lasley - #268 
- [x] ~~website_event_register_free~~ Included in core
- [x] ~~website_event_register_free_with_sale~~ Included in core
- [x] website_field_autocomplete - By @lasley - #269 
- [ ] website_form_recaptcha - By @lasley - #267 
- [ ] website_forum_censored
- [ ] website_hr_department - By @JonathanNEMRY - #335 
- [x] website_legal_page - By @lasley  - #353  
- [x] website_logo - By @chienandalu - #357 
- [ ] website_menu_by_user_status - By @GillesTephaneMeyomesse - #324 
- [ ] website_menu_multilang - By @reinka - #352 
- [x] website_no_crawler - By @RoelAdriaans-B-informed - #301
- [x] website_odoo_debranding - By @Yajo - #360 
- [x] ~~website_payment_v10~~ - Removed as now is in core
- [ ] website_portal_contact - By @osotranquilo #350
- [ ] website_portal_purchase - By @Tardo - #351
- [x] ~~website_portal_v10~~ - Removed as now is in core
- [x] ~~website_sale_collapse_categories~~ - Removed as it's integrated on v9 core
- [x] website_sale_order_company - Moved to OCA/e-commerce - By @reinka https://github.com/OCA/e-commerce/pull/180
- [ ] website_sale_unsaleable_options - To be moved to OCA/e-commerce
- [x] website_seo_redirection - ~~By @mobilizer - #294~~ - By @lasley - #329 
- [x] website_signup_legal_page_required - By @Yajo - #249 
- [x] website_snippet_anchor - @Yajo #361 
[...]

PR #299, opened 2016-12-19
[9.0][website_canonical_url] Migrate.

yajo:
Did almost nothing from 8.0 branch. Just a squash on Transifex's commits, and most basic changes.

@Tecnativa


pedrobaeza:
What you can do is to propose this to master.

yajo:
I'd love to, but I think I'd need authorisation from @tremlin and @rami-wafaie, because that would imply a license change.

pedrobaeza:
OK, let's see what they say, but we can also rewrite the code from scratch (and we'll probably need it as v10 controller handling is different - even there are already included some things I think - like the pager).

pedrobaeza:
@Yajo, tests are not correct.

yajo:
https://github.com/OCA/server-tools/pull/666 should have fixed it, I rebased, let's see.

yajo:
Finally [tests are working](https://travis-ci.org/OCA/website/jobs/187151583#L1013-L1030) 😅 

However Travis went ❌ because of some weird thing... [Only OCB build fails with `website_portal_contact` tour, with no traceback](https://travis-ci.org/OCA/website/jobs/187151583#L1347-L1374)... It makes no sense, Odoo build works!

I think I need some help with that...

pedrobaeza:
I have no clue either. Have you checked commit history to see if there's an specific commit for that module in OCB that it's not in Odoo?

yajo:
But it is an OCA addon.

pedrobaeza:
Uhm, and it's failing also on main branch?

yajo:
Seems like not.

pedrobaeza:
Yeah, I see that it's failing in both Odoo and OCB. First step should be to fix the main branch.

yajo:
There was a problem with Sass indentation and encoding I fixed, please check again now, only OCB fails.

yajo:
Yay! #306 made Travis 💚

pedrobaeza:
Please squash a bit commits from 8.0

yajo:
Done.

pedrobaeza (approved):

cubells (approved):
:+1: 

LGTM

rruebner:
@Yajo We want to use this add-on for odoo 10. Is there a PR for odoo 10 already in progress? Otherwise we would migrate it to odoo 10.

website_canonical_url/models/website.py
```diff
@@ -0,0 +1,35 @@
...
+                url_parts[2] = lang + url_parts[2]
+                canonical_url = urlunparse(url_parts)
+        # Special case for rerouted requests to root path
+        try:
+            if (canonical_url.startswith("/page/") and
+                    self.menu_id.child_id[0].url == canonical_url):
+                canonical_url = "/"
+        except:  # pragma: no cover
+            pass
+        return canonical_url
```
rruebner, line 35: @Yajo Do you think it is enough to use relative paths as canonical links? Based on the [google docs](https://support.google.com/webmasters/answer/139066?visit_id=1-636293281057402117-902391869&hl=en&rd=1) you should avoid using relative urls.

yajo: I did not know about that, definitely a 💩, but it should be fixed.

yajo:
@rruebner you can check https://github.com/OCA/website/issues/262 to see migration attempts for v10. Normally the first thing you have to do when you want to migrate something is to go there and ask a maintainer to add you as the migrator for X addon, and when it's done ask them to link the PR. It's or way to avoid dupe efforts.

PR #302, opened 2016-12-20
[9.0] [WIP] website_crm_privacy_policy module migrate

cubells:
- Updated README
- Updated views
- Updated js

WIP

cc @Tecnativa


website_crm_privacy_policy/static/src/js/scripts.js
```diff
@@ -1,49 +1,36 @@
...
+ * Copyright 2016 Vicent Cubells <vicent.cubells@tecnativa.com>
  */
+odoo.define('website_crm_privacy_policy.crm_policy', function (require) {
+    'use strict';
 
-// Check jQuery available
-if (typeof jQuery === 'undefined') { throw new Error('Website CRM privacy policy addon requires jQuery') }
+    var core = require('web.core');
+    var ajax = require('web.ajax');
+    var _t = core._t;
```
yajo: Add `var animation = require('web_editor.snippets.animation');`

website_crm_privacy_policy/static/src/js/scripts.js
```diff
@@ -1,49 +1,36 @@
...
 
-+function ($) {
-    'use strict';
+    crm_policy.registry.accept_policy = snippet_animation.Class.extend({
```
yajo: `return animation.registry.accept_policy = animation.Class.extend({`

website_crm_privacy_policy/static/src/js/scripts.js
```diff
@@ -1,49 +1,36 @@
...
+        selector: '.s_website_form',
 
-    $(document).ready(function() {
-        var _t = openerp._t;
+        start: function() {
+            var self = this;
```
yajo: Delete this line.

website_crm_privacy_policy/static/src/js/scripts.js
```diff
@@ -1,49 +1,36 @@
...
+            this.$target.find('.o_website_form_send').on('click', function (e) {
```
yajo: Better get used to proxies: `this.$('.o_website_form_send').on('click', $.proxy(this.accept_policy, this));`

website_crm_privacy_policy/static/src/js/scripts.js
```diff
@@ -1,49 +1,36 @@
...
+                self.accept_policy(e);
+            });
+        },
 
         // Validate form
-        $('form[action="/crm/contactus"]').on('submit', function(e) {
-            // Validate privacy_policy is checked
-            if(!$('input[name="privacy_policy"]').is(':checked')) {
+        accept_policy: function(e) {
```
yajo: Other developers appreciate real names for variables, functions, etc. Use `event` instead of `e` for the sake of readability please.

website_crm_privacy_policy/static/src/js/scripts.js
```diff
@@ -1,49 +1,36 @@
...
+            e.preventDefault();  // Prevent the default submit behavior
+            this.$target.find('.o_website_form_send').off();  // Prevent users from crazy clicking
+
+            var self = this;
```
yajo: Remove this line, you are not using it.

yajo (changes requested):

website_crm_privacy_policy/static/src/js/scripts.js
```diff
@@ -9,28 +9,24 @@ odoo.define('website_crm_privacy_policy.crm_policy', function (require) {
...
-            });
+            this.$('.o_website_form_send').on('click', $.proxy(this.accept_policy, this));
         },
 
         // Validate form
-        accept_policy: function(e) {
-            e.preventDefault();  // Prevent the default submit behavior
+        accept_policy: function(event) {
+            event.preventDefault();  // Prevent the default submit behavior
             this.$target.find('.o_website_form_send').off();  // Prevent users from crazy clicking
```
yajo: The normal thing to do here is to disable the button instead of unbinding its events. Please make sure [`off()`](https://api.jquery.com/off/) is doing what you expect.

website_crm_privacy_policy/static/src/js/scripts.js
```diff
@@ -9,28 +9,24 @@ odoo.define('website_crm_privacy_policy.crm_policy', function (require) {
...
 
-            var self = this;
-            if(!this.target.find('input[name="privacy_policy"]').is(':checked')) {
-                e.preventDefault();  // Prevent form from submitting
+            if(!this.$target.find('input[name="privacy_policy"]').is(':checked')) {
```
yajo: You can use the `this.$('input[name="privacy_policy"]')` [shorthand available in animations](https://github.com/odoo/odoo/blob/9.0/addons/web_editor/static/src/js/snippets.animation.js#L12-L14).

website_crm_privacy_policy/static/src/js/scripts.js
```diff
@@ -9,28 +9,24 @@ odoo.define('website_crm_privacy_policy.crm_policy', function (require) {
     var core = require('web.core');
     var ajax = require('web.ajax');
     var _t = core._t;
+    var animation = require('web_editor.snippets.animation');
 
-    crm_policy.registry.accept_policy = snippet_animation.Class.extend({
+
+    return animation.registry.accept_policy = animation.Class.extend({
         selector: '.s_website_form',
```
yajo: You are affecting more forms than just the `website_crm` one, right? Shouldn't this be reflected in the addon name?

website_crm_privacy_policy/static/src/js/scripts.js
```diff
@@ -9,28 +9,24 @@ odoo.define('website_crm_privacy_policy.crm_policy', function (require) {
     var core = require('web.core');
     var ajax = require('web.ajax');
```
yajo: I think you are not using this dependency. Please check and drop it if so.

yajo (changes requested):

website_crm_privacy_policy/static/src/js/scripts.js
cubells: How can affect "only" the website_crm form?

yajo (changes requested):
I have tested in runbot and current behavior is plainly broken. Alert triggers twice and form is never sent.

However, I checked that [nowadays every major browser supports HTML5 validation](http://caniuse.com/#search=required), so how about simply removing all the JS stuff and replacing it by a `required="required"` attribute in the legal terms checkbox? That will trigger the native, beautiful and automatic browser validation with no extra work.

pedrobaeza:
@Yajo, please make the change as you have proposed.

yajo:
I just did a quick update, so you can see the idea, but did not test it. Let's wait for runbot and see.

yajo:
OK @pedrobaeza @cubells, this is interesting. With https://github.com/OCA/website/pull/302/commits/4dee2974f90ab3b3650b7f3ff538d0381082e3ac it works now.

The problem is not about using or not HTML5 validation, the problem is that the new [`website_form`](https://github.com/odoo/odoo/blob/9.0/addons/website_form) addon [disables it](https://github.com/odoo/odoo/blob/9.0/addons/website_form/static/src/js/website_form.js#L49).

I decided to write the long explanation of the problem as a bug in https://github.com/odoo/odoo/issues/14942, to keep it as a reference. Please check it out.

yajo (approved):

pedrobaeza (approved):
Well, there's no visual sign that you lack of clicking on the accept terms, only "Invalid form". Can this be due to a missing class? For now, I'm merging, and if it can be improved, we make another PR.

pedrobaeza:
In website_sale_require_legal we have a balloon:

seleccion_003

Why don't we have it here?

pedrobaeza:
I wait for merging until the question is solved.

yajo:
Yep, that's all explained in https://github.com/odoo/odoo/issues/14942.

The contactus form uses the `website_form` tools, which has these drawbacks. OTOH the checkout form is a pseudo-standard HTML5 form ([with the *pseudo* part fixed in that addon](https://github.com/OCA/e-commerce/blob/f8cad91db967bd83ee4e2d6384dce6b59af571a8/website_sale_require_legal/views/website_sale.xml#L23-L30)) and thus benefits from having a clear & standard UI feedback on validation error.

Avoiding the `website_form` dependency would mean to refactor the whole `website_crm` addon, so this should be fixed upstream in master. For now, this is what we have.

BTW, doesn't the label get red when you don't check it?

pedrobaeza:
No, it doesn't. You don't have any visual sign of what is incorrect, which is very anti-user friendly.

yajo:
Sorry to correct you but it does:

captura de pantalla de 2017-01-09 10-58-56

It gets red, and the field has an *. This is the expected behavior for a website_form-based form. I agree it's a 💩, but that must be fixed in https://github.com/odoo/odoo/issues/14942 😞

pedrobaeza:
OK, it's very weird indeed and I didn't even noticed about the text turning into red (I was expecting the checkbox is surrounded by red). Anyway, if there's no other solution, let's go this way.

PR #305, opened 2016-12-26
[8.0][FIX]website_snippet_country_dropdown:Avoid sending the country code …

jesusVMayor:
…to the server when the vat is empty.

Currently it avoids leaving the vat field empty altrough it is optional.


pedrobaeza:
@sergio-teruel @Yajo, can you check?

yajo:
Although this fix makes sense, this is not the place. It must be fixed in https://github.com/OCA/e-commerce/tree/9.0/website_sale_checkout_country_vat. This addon is purpose-agnostic, while the other one is targeted for the sale checkout process.

Please mention me when you open the corresponding PR to e-commerce, thanks!

PR #306, opened 2016-12-28
[9.0][FIX][website_portal_contact] Use demo user instead of portal one for tests.

yajo:
This fixes Travis for OCB build, by applying a workaround for https://github.com/odoo/odoo/pull/14859.

https://github.com/OCA/OCB/pull/575 introduced an incompatibility between OCB and Odoo builds, which results in having no portal user for tours.

@Tecnativa


pedrobaeza (approved):
As this greenifies the branch, I fast-track it

PR #307, opened 2017-01-02
[8.0] website_seo_redirection display internal error on multiple origin

kryskool:
Hi 

When the same destination have multiple origin like:

- /tinyerp -> /odoo 
- /openerp -> /odoo

When visit /odoo we have an internal error (except_orm) and failed to display website properly

I have made a test to reproduce the error, and patch to fix this issue

Regards,

### DIFF
--- a/website_seo_redirection/models/website_seo_redirection.py
+++ b/website_seo_redirection/models/website_seo_redirection.py
@@ -105,7 +105,9 @@ class WebsiteSeoRedirection(models.Model):
             ("destination", "=", path),
             ("relocate_controller", "=", True),
         ])
-        if not redirection.origin:
+        if len(redirection) > 1:
+            raise NoOriginError(_("Too many origin found for this redirection."))
+        elif not redirection.origin:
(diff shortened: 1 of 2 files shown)


pedrobaeza:
@Yajo, please check

website_seo_redirection/models/website_seo_redirection.py
```diff
@@ -105,7 +105,9 @@ def find_origin(self, redirected_path=None):
             ("destination", "=", path),
             ("relocate_controller", "=", True),
         ])
-        if not redirection.origin:
+        if len(redirection) > 1:
+            raise NoOriginError(_("Many origin found for this redirection."))
```
yajo, line 109: I understand the issue, but I think you are not addressing it right.

Here you are replacing one exception by another, but any exception will get the user to a 500 error anyway.

I think the correct path is to add a constraint that does not allow the web admin to create >1 redirections to the same destination if any of them has `relocate_controller==True`. This way, you will never get to this situation and then no exception will be raised.

yajo (changes requested):

PR #310, opened 2017-01-03
[10.0] Migrate website legal page to version 10.0

hurrinico:


eLBati (approved):
thanks

eugen-don:
@hurrinico 
thank you my friend, i was looking for this module... works like expected...

lasley (approved):

lasley (changes requested):
Oops actually looks like there's a few warnings 

https://travis-ci.org/OCA/website/jobs/190224344#L456

lasley:
Hint the warnings are due to us using `website.` in a few views, which is a namespace we don't own. I'm almost positive it should be as simple as removing that prefix from the ids & updating the link

pedrobaeza:
@hurrinico please fix what @lasley says and I'll merge.

eLBati:
@lasley @pedrobaeza @hurrinico changing
```
<template id="website.legal" name="Legal advice" page="True">
```
to
```
<template id="website_legal_page.legal" name="Legal advice" page="True">
```
would prevent page ```/page/legal``` to be served (404 error)

What do you suggest?

cc @antespi

lasley:
Hmmmm I see the problem. Odoo SA's [`/page` interface](https://github.com/odoo/odoo/blob/b228f548ca10e61d3b802fa624bdaf8a2f8b11b2/addons/website/controllers/main.py#L115) is enforcing the bad practice of editing namespaces you don't own.

A controller with an explicit route is my preferred. Interface or not, it is pretty commonly accepted practice to isolate everything by namespacing & breaking the good practice is not worth the 6-7 lines of code we'll save.

That's my opinion. Anyone else have thoughts?

antespi:
I agree with @lasley

JonathanNEMRY:
add setup here https://github.com/hurrinico/website/pull/1

hurrinico:
@lasley @antespi i'm quite busy at the moment, and that type of mod require to change all the module, but prs are welcom

eLBati:
@lasley I don't know how to implement what you suggest. Hints?

lasley:
@eLBati - I'll actually submit a PR in the next few days. I was just looking, and it seems I had a branch doing exactly this in v9 + updating the privacy policy in accordance with some advice I have received from legal counsel.

yajo (approved):
I feel adding custom controllers is an overkill.

This addon just adds 3 pages, and it uses the tool Odoo uses for pages. It lets submodules to know the URL for that page and make sure it exists; these pages are expected to be changed by the user. Seems legit IMHO; custom controllers and so on would be more code to maintain for basically no benefit.

OsoTranquilo (approved):

yajo:
@lasley This is working perfectly and is holding other PRs; mind if we merge as it is? You can update the texts later, and we can defer the controller conversation to future v11...

lasley:
@Yajo - merging as is will break the 10.0 Runbot build going forward. Is that something we're OK with?

yajo:
Hmm no, definitely. It seems it breaks due to 3 warnings:

```
2017-05-29 09:33:12,132 149 INFO openerp_test odoo.modules.loading: loading website_legal_page/views/website_legal.xml
2017-05-29 09:33:12,199 149 WARNING openerp_test odoo.addons.base.ir.ir_model: Creating the ir.model.data legal in module website instead of website_legal_page.
2017-05-29 09:33:12,199 149 INFO openerp_test odoo.modules.loading: loading website_legal_page/views/website_privacy.xml
2017-05-29 09:33:12,268 149 WARNING openerp_test odoo.addons.base.ir.ir_model: Creating the ir.model.data privacy in module website instead of website_legal_page.
2017-05-29 09:33:12,268 149 INFO openerp_test odoo.modules.loading: loading website_legal_page/views/website_terms.xml
2017-05-29 09:33:12,341 149 WARNING openerp_test odoo.addons.base.ir.ir_model: Creating the ir.model.data terms in module website instead of website_legal_page.
```

Hard to fix if we don't do custom controllers... OK, well if you can have it for today it would be great, we are at #ocacodesprint merging like crazy goats! 🐐

yajo (changes requested):
As said, we need to fix bots, which will probably include adding the controller.

pedrobaeza:
What is the state of this?

yajo:
You see, current design raises warnings, which kills runbot. @lasley suggested to add custom controllers and that's what we're waiting for.

pedrobaeza:
OK, thanks for the heads-up. Let's see if @hurrinico can work on this.

hurrinico:
Hi @pedrobaeza as i said few days ago, this mod is quite hard and at the moment i'm quite busy, so can be difficult to me redefine all the module

lasley:
I've actually already done this, just need to add tests. Will submit a PR today

pedrobaeza:
@lasley did you checked your code for making the PR?

lasley:
hurrinico#3 - TBD whether tests pass

pedrobaeza:
Everything is now green, so please update your reviews, @lasley and @Yajo

website_legal_page/controllers/main.py
```diff
@@ -0,0 +1,20 @@
...
+
+    @http.route([
+        '/legal',
+        '/legal/<string:page>',
+    ],
+        auth='public',
+        type='http',
+        website=True,
+    )
+    def show(self, page='legal'):
```
yajo, line 19: SEO-wise, if you have a `/legal` and a `/legal/legal` page that are the same, you get penalized.

I think you should better not add a default page and require it (404 error otherwise), or make it redirect to `/legal/legal` if `not page`.

Remember to add `**kwargs` so that we can add `?debug=assets` and so on.

Also, this controller's name should be `legal` instead of `show` IMHO.

website_legal_page/tests/test_controller.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class TestController(HttpCase):
+
+    def test_legal(self):
+        """ It should return a 200 for the default page. """
+        response = self.url_open('/legal')
+        self.assertEqual(
+            response.getcode(),
+            200,
```
yajo, line 15: This will change with my above comment.

website_legal_page/views/website_legal.xml
```diff
@@ -6,17 +6,17 @@
           inherit_id="website.layout_footer_copyright"
           name="Legal advice Link">
     <xpath expr="//footer//div/span[@t-field='res_company.name']" position="after">
-        <span>- <a href="/page/legal">Legal advice</a></span>
+        <span>- <a href="/legal">Legal Advice</a></span>
```
yajo, line 9: `/legal/advice` maybe? Better than `/legal/legal`, right?

website_legal_page/views/website_privacy.xml
```diff
@@ -6,17 +6,17 @@
           inherit_id="website.layout_footer_copyright"
           name="Privacy policy Link">
     <xpath expr="//footer//div/span[@t-field='res_company.name']" position="after">
-        <span>- <a href="/page/privacy">Privacy policy</a></span>
+        <span>- <a href="/legal/privacy">Privacy Policy</a></span>
```
yajo, line 9: It turns out that it is better for SEO having a explicit name than a short one (unless the explicit one is too long). What if we take the chance to rename this one to `/legal/privacy-policy`?

website_legal_page/views/website_terms.xml
```diff
@@ -6,11 +6,11 @@
           inherit_id="website.layout_footer_copyright"
           name="Terms of use Link">
     <xpath expr="//footer//div/span[@t-field='res_company.name']" position="after">
-        <span>- <a href="/page/terms">Terms of use</a></span>
+        <span>- <a href="/legal/terms">Terms of use</a></span>
```
yajo, line 9: This one could be `/legal/usage-terms` too.

yajo (changes requested):

rafaelbn:
@Yajo as @hurrinico said https://github.com/OCA/website/pull/310#issuecomment-307391503 he cannot finish this PR so please make a PR to his branch to go ahead

website_legal_page/controllers/main.py
lasley: `/legal` is the default route here. I guess technically `/legal/legal` would work too, but I'm pretty sure I rigged everything up as `/legal`

lasley: Oh I see, you're recommending explicitly denying (still early here, need coffee).

What about `/legal/overview` or something? I don't like the one duplicated

website_legal_page/views/website_legal.xml
lasley: 👍  will solve above

website_legal_page/views/website_terms.xml
lasley: I went with `terms-of-use` so that it more closely matches the link text

lasley:
Superseding with #353 in order to save a bunch of side merges/PRs

Issue #304, opened 2016-12-21
[9.0] website_snippet_anchor is blocking menu


lucode:
This module is blocking the editing of the menu from frontend and backend.
see the forum post:
https://www.odoo.com/de_DE/forum/hilfe-1/question/v9-website-menu-is-blocked-for-editing-111990

pedrobaeza:
Please try on runbot to see if you reproduce the problem.

rafaelbn:
Please @lucode try here http://3275938-9-0-3c6c72.runbot1.odoo-community.org/?db=3275938-9-0-3c6c72-all (admin/admin) and let us know if you can reproduce problem. Thanks

lucode:
It does reproduce the error in runbot. No editing of Menu entries possible.

pedrobaeza:
@Yajo, can you check?

yajo:
Right now I see this problem in main 9.0 branch build:

![captura de pantalla de 2017-01-05 12-58-18](https://cloud.githubusercontent.com/assets/973709/21679884/b06475f0-d346-11e6-8ed4-c579c174d9ef.png)

I know it from other times, let me add a couple of quick fixes and see if it works.

PR #311, opened 2017-01-05
[9.0][FIX][website_portal_purchase] Fix copyright header.

yajo:
Sass requires consistent indentation even for multiline comments, so I add it here.
I take the time too to change (C) for Copyright (which actually has a legal meaning).
This should fix https://github.com/OCA/website/issues/304#issuecomment-270628689.

@Tecnativa


pedrobaeza (approved):
Fast-tracking for fixing the branch. Weird bug anyway.

PR #316, opened 2017-01-30
[FIX][IMP][website_event_register_free_with_sale] Store ticket type and allow multiple tickets

yajo:
Previous implementation had these problems:

1. It did not store the ticket type along with the registration when it was a free one.
2. It did not allow to reserve multiple ticket types at once, tightly related to problem 1.

With this patch:

- `event.registration._prepare_registration()` is simplified because it does not require a recordset for the event.
- You can have several free tickets for the same event.
- You can reserve several free tickets along with several paid tickets in one step.
- The free ticket type is stored along with the reservation record.
- We only use 1 session key for all of this (`free_tickets`), which makes maintenance easier.

@Tecnativa


rafaelbn (approved):
Tested

pedrobaeza (approved):

PR #321, opened 2017-02-16
[9.0][MIG][website_crm_quick_answer]

yajo:
Migration includes:

- Drop filter and use the new domain field in the automated action.
- Migration script to avoid FK error when dropping it.
- Warn about non-immediate response.

This will only work after @odoo merges https://github.com/odoo/odoo/pull/15528, so I'll label as WIP until then. However, code is stable.

@Tecnativa


pedrobaeza (approved):

yajo:
Odoo rejected https://github.com/odoo/odoo/pull/15528, so it's now included in the addon, and will have to be this way until v11, so it's explained in README. Ready to review then.

cubells (approved):
:+1: 

LGTM

Code review and tested.

yajo:
Rebased, it should go green now.

PR #322, opened 2017-02-18
[ADD] website_blog_location

hbrunn:
Locations in website blogs
==========================

This module was written to allow you to attach a location to a blog entry.

Users will then be able to navigate to ``/blogmap/<blogid>`` and see a map where blog entries can be selected by location.

Usage
=====

- go to Knowledge/Blog posts
- select some blog post
- enter a latitude and longitude
- click the `Show in map` link and be happy to see your blog post there
- link to ``/blogmap/<blogid>`` in some menu or page
- alternatively, enable the right column for your blog in the frontend, and enable the option `Show in map`. This generates a link to the map

.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
    :target: https://runbot.odoo-community.org/runbot/186/8.0

Known issues / Roadmap
======================

* adding the latitude/longitude fields on the frontend would be a lot more user friendly
* use a map widget to select the location
* add a snippet to show the map
* custom markers would be nice
* some of the code already tries to be more generic. In the end, we should have some base module that shows maps for arbitrary records, and have this one depend on it
* in the end, there probably should be an openlayers base module where users can configure which mapping engine to use etc.


lasley:
@hbrunn - I am planning this exact same thing for product variants right now. 

Would you be opposed to me splitting some of this logic out into a base module ~~in maybe `server-tools`~~ in this repo? I'm thinking the `LocalizableMixin` and the template would help me a lot.

I'll obviously review this regardless, but figure it would save some effort if I split the base off first.

hbrunn:
on the contrary! If you can invest time in this (I consider this module more on the hacky side), how about splitting off from https://github.com/OCA/geospatial/tree/10.0/base_geoengine a module that takes care of the widget/website part but doesn't depend on postgis et al?

pedrobaeza:
This may benefit from the geoengine project

simahawk:
@hbrunn is this still something you are working on?

website_blog_location/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV <http://therp.nl>
```
nikul-serpentcs, line 2: Remove Copyright from all `__init__` file

website_blog_location/__openerp__.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+# © 2017 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Locations in website blogs",
+    "version": "8.0.1.0.0",
+    "author": "Therp BV,Odoo Community Association (OCA)",
```
nikul-serpentcs, line 7: Add also `Website` as per [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L10)

nikul-serpentcs (commented):
Improve Code

hbrunn:
this module was made for a pet project that I'm going to migrate to 10 or 12 soon, so probably I won't continue working on this, but on the v10 version of it. Thanks for the review, I'll pick them up during migration.

PR #324, opened 2017-03-08
[MIG] website_menu_by_user_status form v8 to v10

GillesTephaneMeyomesse:


bistaray (commented):
I installed (Enterprise Edition) and there was no menu added at "Settings > Configuration > Website Settings > Configure Website menus" - nor does this path exist, and I don't see any XML record adding such a menu.  The fields were added to the model, so by creating my own menu, I was able to edit the menu and set values for "Logged In" and "Not Logged In".

I tried a menu with Logged In checked, and Not Logged In unchecked.

I tried a menu with Logged In unchecked, and Not Logged In checked.

I tried a menu with both unchecked.

Each time - the menus kept showing no matter if I logged in or out.

Tried incognito mode so make sure there were no cache issues.

Is this module Community Only?

GillesTephaneMeyomesse:
Hey ! Sorry i updated Readme so first you must check you have activate developer mode. after go to Website Admin > Configuration > Settings > Configure website menus. Remove default filter to edit website menu.
This is a community module of course

eugen-don:
@GillesTephaneMeyomesse 

Hi i just checked that this only disables the visibility of the menu item but does not prevent the user from entering the menu by simply typing the url, can you update the code to display a 404 error if the user is not logged ?

GillesTephaneMeyomesse:
@eugen-don Hi it's possible to return display 404 error for unlogged users. I will make it on next feature

GillesTephaneMeyomesse:
Hi @eugen-don I update this module, now return 404 if page is hidden for public or logged user

leemannd:
Feedback from Odoo Enterprise. The menu is perferctly removed and the redirection doesn't work. So still I'm pretty happy. Thank you @GillesTephaneMeyomesse

tazus:
I test current branch:
1.  hide/show works
2. 404 redirect for restrict page not working.
3.  the link to the page not on the menu will display page not found, after un-install this module the link working again.

Suggestion:  adding th function to input/limit the URLs/links that the menu item will  show/not shown

GillesTephaneMeyomesse:
@tazus I'have, @leemannd I did a little cleaning on the code and custom the 404 page. Can you review and test this please ?

leemannd:
@GillesTephaneMeyomesse Tested in Odoo Enterprise:
- the menu hiding works for all menuitems
- the redirection works for all menuitems that contains '/page'. And won't work for '/shop', '/jobs', '/event'...
We should at least mention it.

As you want to make private parts of the website, maybe we could add a template for robots.txt like:
`  <template id="website.robots">
User-agent: *
Disallow: /MENUS_DISABLED*
Sitemap: <t t-esc="url_root"/>sitemap.xml
  </template>`

website_menu_by_user_status/controllers/main.py
```diff
@@ -0,0 +1,72 @@
...
+from odoo.addons.web.controllers.main import Home
+
+
+class Website(Home):
+
+    def _get_pages(self):
+        """
+        All pages width url
+        """
+        return request.env['website.menu'].sudo().search([]).filtered('url')
```
simahawk: this might be really expensive if you have a lot of menu items. Search for `[('url', '!=', False))]` should be better.

website_menu_by_user_status/controllers/main.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def _check_user_access(self):
+        """
+        Is user logged
+        """
+        return request.env.user == request.website.user_id or False
+
+    def _get_suggest_page(self):
```
simahawk: this could be cached per user and lang.

website_menu_by_user_status/controllers/main.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        All pages accessible by the user (logged or not)
+        """
+        if self._check_user_access():
```
simahawk, line 37: BTW couldn't this be achieved with some record rule?

simahawk (changes requested):
I'm not in favor of asking for refactoring while migrating. So, if the code has been ported as it is, fine. Otherwise, pls, consider my comments. :)

website_menu_by_user_status/controllers/main.py
GillesTephaneMeyomesse: Thank's

leemannd:
New feedback from an Enterprise Version. Forms don't work anymore. If you take the base `contact us` form you have a redirection after sending the form with `page/...` and it isn't in the allowed menus.

GillesTephaneMeyomesse:
@leemannd this is curious i'ill check this out. Thank you

GillesTephaneMeyomesse:
@leemannd I don't think this is a problem because page/... gives a redirection on the public pages

leemannd:
@GillesTephaneMeyomesse Maybe i've been explaining myself poorly and maybe it comes from the enterprise version, but there is something highly wrong.
This module, with the redirection, has been breaking forms in my tests.
1. Install this module on an enterprise version
2. Install also website_crm (odoo base module)
3. Go to the contact page `_yourInstance_/page/contactus`
4. You have a from to fill in. And when you do, the message and the form is processed, you are redirected to the page `_yourInstance_/page/website_crm.contactus_thanks` and it shows a 404 error.

What is your opinion about it?

EDIT: When installing this module, I made no other configurations and all routes are available for logged and non-logged users.

EDIT2: My previous explanation was poorly written.

leemannd (changes requested):
Bug found with the redirection and forms like `website_crm`. Explained before.
N.B: tested only with the enterprise version.

GillesTephaneMeyomesse:
@leemannd you have right let me fix that, thanks

GillesTephaneMeyomesse:
@leemannd I cleaned the code and added a small fix for this bug sorry. Can you do tests and review and test this please ?

website_menu_by_user_status/controllers/main.py
GillesTephaneMeyomesse: @simahawk I think it's possible but this is little hard to check if a user is logged in or not. Do you have an idea to do this ?

leemannd (commented):
The bug concerning the forms is corrected. Thanks!

There is still one concern on my side and it was part of my first comment. If you install modules like `website_sale` or `website_event`, the top menu is hidden. But this two modules are installing new routes like `_your_wesite_/shop` or `_yourWebsite_/event` and the redirection won't work in such cases.
At least it's worth a comment in the README?

GillesTephaneMeyomesse:
@leemannd  Thank you for your comments.
I will see the best way to manage these new routes added by other modules.

website_menu_by_user_status/controllers/main.py
simahawk: I have a PoC working w/ groups + rule. I think this module should be based upon this and get rid of the check on pages:  adds logic that is not perfomant and goes far beyond the purpose of the module. Page/routes permission should be granted by routes/pages logic themselves. Very likely we gonna get rid of the redirect issue too.

Can't work on this today but I'm going to get back w/ another PR most probably on Monday.

simahawk: there you go https://github.com/OCA/website/pull/394

GillesTephaneMeyomesse:
@simahawk Hi, your idea is very interesting ! But when i tested your module i noticed a security breach. suppose a user who has been removed the right to see the page contact_us it's true that it is no longer displayed on the top menu that is good but if you type the url your_instance/contact_us_force_access for example it will have a 404 suggesting  your_instance/contact_us and therefore he will have access to this page again ! Can you check that? Thank you

simahawk:
@GillesTephaneMeyomesse there's no security breach. As I told you this module should just take care of protecting menu items, not pages. Pages should be protected with their own access rule. 
The part of the page handling (what's in controllers) should be moved to another module and rewritten according to this logic.

One possible way is to add a computed boolean field to page to retrieve a related menu item and check its permission. Then you create a record rule that filters pages based on this flag.
By doing so you won't need to override controllers and add all that complex logic (which is not good regarding performance).

rafaelbn:
Hi @flotho you have here the module you were interested in v10

Hi @GillesTephaneMeyomesse ,

This module is already in v11 do you want to finish this PR or maybe is better to do the backport of the v11 module?

- [ ] Some improvements of v11 modules done in https://github.com/OCA/website/pull/491

Let me know please

Thanks

pedrobaeza:
Any news?

GillesTephaneMeyomesse:
Hi @rafaelbn, @pedrobaeza I don't kwow how this module where merged in v11 master with my name may be @dufresnedavid workon... I tried to fix in this migration one bad use case.
I think there is small security problem on this module as discussed with @simahawk and he had good idea in https://github.com/OCA/website/pull/394
My use case : 
1 - change "Contact us" access page to "User not logged"
2 - ok that work : Page menu is not accessible in web menu
3- but if user try this "http://localhost:8069/web/xxx" Odoo raise 404: Page not found! but suggests "Maybe you were looking for one of these popular pages ?" where we found "Contact us"
4 - and user can clic on "Contact us" and access

This is not good in my opinion because we said only accessible for "user logged"
Manifest is also wrong because it said "Go to Website > Configuration > Settings > Pages." but this menu is only accessible in "Activate the developer mode" and module use deprecated "@api.one"

simahawk:
Sorry, no time, no project related to such things lately...
I'm kind of a blocker here :stuck_out_tongue_closed_eyes:

I'll try to move fwd w/ https://github.com/OCA/website/pull/394 as agreed w/ Holger and ask everyone to rebase on top of that so we have a common implementation.

I don't like this complex machinery on pages...
OTOH I'm not particularly concerned (nor involved in projects) on v10.

So, if everyone here think that we can go ahead w/ current implementation: go ahead. I'll unlock my review.
We can tackle the issue from v11 only and whoever wants can backport later.

ddufresne:
@GillesTephaneMeyomesse @simahawk At the base, this module did not aim to secure pages but only to hide/show menu items based on the user being logged in or not.

I am myself confused by the relevancy of this module in its current state. However, we have (at Numigi) clients who use it. These clients are aware that hiding the menu item does not make the page secure and are OK with the mechanism. For this reason, I will fork this module in a private repo outside of OCA.

Let me know if you need reviews for an improved version. However, I will stop making development efforts on this.

PR #326, opened 2017-03-16
[10.0][MIG] website_menu_by_user_status

eugen-don:
Migrated Module to V. 10.0


pedrobaeza:
There is another migration of the same module in https://github.com/OCA/website/pull/324, so please review that one.

PR #294, opened 2016-12-09
[MIG][10.0] website_seo_redirection

mobilizer:
Migration to version 10.0

PR #329, opened 2017-03-18
[10.0][MIG] website_seo_redirection

lasley:
This supersedes #294. I was going to initially just add a commit into that branch, but the migration commit was out of order and the migration itself was in a merge commit. It was significantly easier to just perform the whole migration again myself & fix the tours in one fell-swoop.

@Yajo @laslabs - take a look at the tour changes for v10. Specific files to look at:
* test_ui.py
* website_seo_redirection.tour.js

High level tour differences to note:
* [New loading system](https://github.com/odoo/odoo/blob/10.0/addons/web_tour/static/src/js/tour_manager.js#L226) using `wait_for`
* `mode` is now `test`
* `path` is now `url`
* `element` is now `trigger`
* `title` is now `content`
* `waitFor` is now `extra_trigger`, if different from `trigger`
* `id` key is now first param of `.register`
* `steps` is now the third param of `.register`
* Tours now only operate on body, so triggering on meta tags is no longer possible 😦 

We also have a fairly large problem - the new tour system is incompatible with the old version of PhantomJS that Travis uses. [This is the error:](https://travis-ci.org/LasLabs/website/jobs/212562744#L838)

```
2017-03-18 22:48:30,763 5348 ERROR openerp_test odoo.tests.common: phantomjs: u"ReferenceError: Can't find variable: MutationObserver\n at http://127.0.0.1:8069/web/content/508-e4a58c9/web.assets_common.js:3668\n at http://127.0.0.1:8069/web/content/508-e4a58c9/web.assets_common.js:547\n at http://127.0.0.1:8069/web/content/508-e4a58c9/web.assets_common.js:541\n at http://127.0.0.1:8069/web/content/508-e4a58c9/web.assets_common.js:546\n at http://127.0.0.1:8069/web/content/508-e4a58c9/web.assets_common.js:548\n at http://127.0.0.1:8069/web/content/508-e4a58c9/web.assets_common.js:541\n at http://127.0.0.1:8069/web/content/508-e4a58c9/web.assets_common.js:546\n at http://127.0.0.1:8069/web/content/508-e4a58c9/web.assets_common.js:547\n at http://127.0.0.1:8069/web/content/508-e4a58c9/web.assets_common.js:541\n at http://127.0.0.1:8069/web/content/508-e4a58c9/web.assets_common.js:546\n at http://127.0.0.1:8069/web/content/508-e4a58c9/web.assets_common.js:547\n at http://127.0.0.1:8069/web/content/508-e4a58c9/web.assets_common.js:541\n at http://127.0.0.1:8069/web/content/508-e4a58c9/web.assets_common.js:546\n at http://127.0.0.1:8069/web/content/508-e4a58c9/web.assets_common.js:937 (in done)\n at http://127.0.0.1:8069/web/content/508-e4a58c9/web.assets_common.js:957\n(leaf frame on top)"
```
I included a hotfix to update Phantom to 2.1.14 in this Travis file as a proof of concept, which turned the 🔴  build 🍏 . I think it might be wise to include this in MQT instead though maybe.

Changelog:
* Replace openerp to odoo
* Bump version
* Upgrade tour test
* Fix security
* Update phantom in travis file


website_seo_redirection/tests/test_ui.py
```diff
@@ -2,15 +2,20 @@
...
+
+    post_install = True
+    at_install = False
+
+    def test_admin_tour(self):
         """Redirections work with default language."""
-        tour = "odoo.__DEBUG__.services['web.Tour']"
+        tour = "odoo.__DEBUG__.services['web_tour.tour']"
         self.phantom_js(
-            "/en_US",
```
yajo: I think I prefer if you leave this URL, just because it should store a cookie and trigger a redirection by itself to `/`, which was a special case this addon had to take into account if I don't remember badly... Is it possible?

yajo (changes requested):
Quite amazingly educational & cliean migration 😲, just one little detail.

website_seo_redirection/tests/test_ui.py
lasley: Sure - admittedly I only changed this to simplify the case as I was banging my head against the wall. I definitely meant to switch it back!

.travis.yml
```diff
@@ -27,11 +27,9 @@ virtualenv:
   system_site_packages: true
 
 install:
-  - git clone https://github.com/OCA/maintainer-quality-tools.git ${HOME}/maintainer-quality-tools
+  - git clone https://github.com/LasLabs/maintainer-quality-tools.git -b hotfix/update-phantom ${HOME}/maintainer-quality-tools
```
lasley: Do not merge with this line. Pending OCA/maintainer-quality-tools#437

simahawk (approved):
code: LGTM

yajo (approved):
Code: 👍 

Functional: if you test in v8 runbot, when you press *Promote* in a page, this section is added:

captura de pantalla de 2017-03-21 09-19-59

I tested in v9 runbot and it seems that feature did not get migrated to v9, but honestly it feels like a real need when you have it. It really makes the addon end-user friendly.

Could you port it?

I won't block it for this if you don't, at least it's better to have this than nothing...

lasley:
@Yajo - I feel like it would be better to add an explicit check to see if the URL exists first vs adding a warning. Something simple like a `requests.get` to the page looking for a 404 I think would be more worthy?

yajo:
Well, my point was more about exposing the feature to the end user under the promote form than about the warning itself. Indeed that would be a neat IMP 😉

lasley:
Ohhhhh I see - the part in the promote page is gone entirely! I thought you were just referring to the warning lmao ok yeah I'll forward port.

yajo:
a09c5f2 🤣 

PS: Still nothing on promote

jholze:
Hi @Yajo, i was checking this addon and it works really good. Very flexible. Maybe we will update our SEO suite to it. But just a question: When i create a permanent redirect, the addon renders a 200 http code instead of 301. Can you maybe explain why ? I check in chrome dev tools and only 200 status codes where displayed.

website_seo_redirection/models/website_seo_redirection.py
```diff
@@ -127,3 +159,32 @@ def redirect_auto(self, path=None, code=301, website=None, rerouting=None):
             dict(request.httprequest.args),
             True,
```
rruebner, line 160: @lasley, @Yajo Hi, I just tested the add-on for odoo 9 and it works so far, really good and simple solution. But all the redirections have only the status code 200 instead of 301.
In the function redirect_auto() you call local_redirect() at the end with only 4 parameters which results in keep_hash=True. Is it supposed to work in this way?
For testing I used keep_hash=False which results in a status code 301 for the redirections.

yajo: Is it possible that [you are not using Firefox](https://github.com/odoo/odoo/blob/9.0/openerp/http.py#L156-L164)?

rruebner: @Yajo Thats correct, I tested it with Chrome. I looked at the function redirect_with_hash(), I didn't understand the problem completely in this function but as I tested it with my temporary change it also worked in Chrome.

lasley:
Promote page has been fixed!

image

yajo:
Cool, but why did you remove the translation? 😆

lasley:
Because there is already an l18n folder including a bunch of translations, one of which was `es.po`. Am I misunderstanding something?

yajo (approved):
Oh, didn't realize.

pedrobaeza (approved):

pedrobaeza:
@lasley , waiting for your final squash at your discretion for doing the merge.

lasley:
Squashed!

PR #332, opened 2017-04-15
[ADD] website_hide

hbrunn:
Hide website
============

This module was written to restore the look and feel of Odoo without the ``website`` module while having it installed anyways. The reason you might want that are modules like `website_mail`, which use the HTML editor bundled with the website addon.

Further, this module synchronizes the default website's languages with the installed languages, in order to support translations using the website editor.

Usage
=====

After installation, you shouldn't see any website menus in the frontend any more, just the standard login screen.


hbrunn:
[travis error](https://travis-ci.org/OCA/website/jobs/222745596#L1529) unrelated

jesusVMayor (approved):

simahawk (approved):

PR #342, opened 2017-05-17
[ADD] website_sale_category_empty: Create module

tedsalmon:
* Add module to hide categories that have no products associated with them

Website Sale - Hide Empty Categories
=============================

Removes any category links from the Shop which have no products associated
to them.

Usage
=====

Simply install the module and refresh your view. You will see the empty
categories disappear

cc @lasley


website_sale_category_empty/__manifest__.py
```diff
@@ -0,0 +1,20 @@
...
+    "author": "LasLabs, Odoo Community Association (OCA)",
+    "license": "LGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "website_sale",
+    ],
+    "data": [
+        "views/website_sale_templates.xml",
+    ]
```
lasley: trailing comma

website_sale_category_empty/models/product_public_category.py
```diff
@@ -0,0 +1,17 @@
...
+from odoo import fields, models
+
+
+class ProductPublicCategory(models.Model):
+    _inherit = 'product.public.category'
+
+    product_ids = fields.Many2many(
+        comodel_name='product.template',
+        string='Products',
+        relation='product_public_category_product_template_rel',
```
lasley: Curious - why the explicit relation/col definitions? Are you hooking into a pre-existing join table?

lasley (changes requested):

website_sale_category_empty/models/product_public_category.py
tedsalmon: Yes, I'm hooking into a pre-existing M2M relation -- This just made it bi-directional :)

lasley: Sweet - can you add a note of the specific relation that you're hooking into please? This could be a pain point in migrations

tedsalmon: I've made a note of this in the README :)

lasley: I think a code comment would be more appropriate, directly above the field definition. As an engineer performing a migration, would you look in the ReadMe for this note while diagnosing a random issue with this field?

pedrobaeza:
Isn't better website_sale_hide_empty_category?

tedsalmon:
Good call, Pedro "Master of naming and reviews" Baeza ;P

pedrobaeza (approved):

lasley (approved):

PR #343, opened 2017-05-19
[10.0][FIX][website_seo_redirection] Fix url updates in promote panel.

rruebner:
If you want to use model functions in javascript you have to use **require("web.Model")** and **new Model("your.model").call()** instead of **require("web.session")** and **session.model("your.model").call()** now in odoo 10.


yajo (approved):
Great!

yajo:
Fast-tracking.

PR #353, opened 2017-06-13
[10.0] [MIG] website_legal_page

lasley:
I didn't want to keep submitting PRs to an edge branch, so I'm superseding #310 with this.

@Yajo - I completed your https://github.com/OCA/website/pull/310#pullrequestreview-43655093 in 8f3c7b28cd770fcbd719b16ada913d64074e3809, which I will squash into 	311c6ca before merge.


website_legal_page/controllers/main.py
```diff
@@ -0,0 +1,19 @@
...
+
+    @http.route([
+        '/legal/<string:page>',
+    ],
+        auth='public',
+        type='http',
+        website=True,
+    )
+    def show_legal_page(self, **kwargs):
+        return self.page('website_legal_page.%s' % page, kwargs)
```
yajo: This will fail! `page` is undefined.

website_legal_page/views/website_legal.xml
```diff
@@ -6,17 +6,17 @@
           inherit_id="website.layout_footer_copyright"
           name="Legal advice Link">
     <xpath expr="//footer//div/span[@t-field='res_company.name']" position="after">
-        <span>- <a href="/page/legal">Legal advice</a></span>
+        <span>- <a href="/legal/advice">Legal Advice</a></span>
     </xpath>
 </template>
 
-<template id="website.legal" name="Legal advice" page="True">
+<template id="legal" name="Legal Advice" page="True">
```
yajo: id should be `advice` instead, as in line 9

website_legal_page/views/website_privacy.xml
```diff
@@ -6,17 +6,17 @@
           inherit_id="website.layout_footer_copyright"
           name="Privacy policy Link">
     <xpath expr="//footer//div/span[@t-field='res_company.name']" position="after">
-        <span>- <a href="/page/privacy">Privacy policy</a></span>
+        <span>- <a href="/legal/privacy-policy">Privacy Policy</a></span>
     </xpath>
 </template>
 
-<template id="website.privacy" name="Privacy policy" page="True">
+<template id="privacy" name="Privacy Policy" page="True">
```
yajo: ID here must be `privacy-policy`, like in line 9

website_legal_page/views/website_terms.xml
```diff
@@ -6,11 +6,11 @@
           inherit_id="website.layout_footer_copyright"
           name="Terms of use Link">
     <xpath expr="//footer//div/span[@t-field='res_company.name']" position="after">
-        <span>- <a href="/page/terms">Terms of use</a></span>
+        <span>- <a href="/legal/terms-of-use">Terms of use</a></span>
     </xpath>
 </template>
 
-<template id="website.terms" name="Terms of use" page="True">
+<template id="terms" name="Terms of use" page="True">
```
yajo: And here, `terms-of-use`

website_legal_page/tests/test_controller.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class TestController(HttpCase):
+
+    def test_unknown(self):
+        """ It should return a 404 for unknown pages. """
+        response = self.url_open('/legal/no-page')
+        self.assertEqual(
+            response.getcode(),
+            404,
```
yajo: It would be nice to add a test like this for just `/legal`

website_legal_page/tests/test_controller.py
```diff
@@ -0,0 +1,24 @@
...
+        )
+
+    def test_privacy(self):
+        """ It should return a 200 for a defined page. """
+        response = self.url_open('/legal/privacy')
```
yajo: It would be nice to add a test like this for each of the 3 pages.
BTW, you said it would be `privacy-policy` instead. Check below.

yajo (changes requested):
Check bots please.

lasley:
Hah bots are part of the reason that I just took the PR over - I obviously tested nothing! Fixed.

yajo (approved):
OK, it seems good enough for me 😊

However, I still want to add a couple of warnings we might want to fix now:

1. Current design of pages will hit https://github.com/odoo/odoo/issues/15437.
2. Changing XMLIDS of legal pages will make old ones disappear and new ones get new. I'm pretty sure many people will not want to reach their lawyers again to re-fix their lost legal terms, so a migration script from <10.0 versions would be very desirable.

lasley:
Hmmm I'm not sure I totally understand the number 1 point. What do I need to do here in order to avoid that?

Regarding the migration script - isn't that better suited for OpenUpgrade?

pedrobaeza:
In this case I don't think we should provide separate views for "content containers", as someone may want to modify the entire legal page without keeping the default layout.

yajo:
> What do I need to do here in order to avoid that?

Yes, @pedrobaeza, but just in case @lasley wants to avoid that: just replace those `<div class="oe_content"/>` by `<t t-call="website_legal_page.${page}_content"/>` and adding that view as a separate one that just contains `<div class="oe_content"/>`.

yajo:
@lasley I took the freedom to squash your last commits to get it ready to merge. I also added the required migration script, so this should be ready to merge now.

pedrobaeza:
Merging then

lasley:
You're awesome, thanks @Yajo! Been backed up with the deploys and totally wasn't able to get back to this. Excellent community effort!

yajo:
Anyone interested on this please take a look at #358

PR #358, opened 2017-07-04
[FIX][website_legal_page] Fix links in reusable template

yajo:
Problem introduced when changing the controllers in #353.

Nobody noticed because the reusable template was not being reused yet, but I'm migrating addons that do and here's the patch.

This should be quite obvious.

@tecnativa


pedrobaeza (approved):
Indeed obvious. Fast-tracking.

PR #360, opened 2017-07-05
[MIG][website_odoo_debranding] Migrate to v10

yajo:
- Relicensed to LGPL.
- The template that does the job can be disabled by the admin.

@Tecnativa


chienandalu:
Testing on runbot:

The custimize option is not present:
image

Also it doesn't seem to unbrand the site by default:

image

yajo:
For some reason it was not installed on runbot by default. Try now please.

chienandalu (approved):
No it works as expected :+1:

pedrobaeza (approved):

Issue #334, opened 2017-04-25
[RFC] website_wow: Universal Snippet Animations


lasley:
This module will provide [WowJS](https://github.com/matthieua/WOW) animations, which is basically just a wrapper on top of the [animate.css](https://github.com/daneden/animate.css) library we all know and love. 

Note that there is commercial licensing for this, but as long as we're open source it's GPLv3 which works just fine for us! _To the LasLabs engineer implementing this - we cannot release this one under LGPL-3, so please use GPL-3._

We will inject the animations button after the existing [Customize](https://github.com/odoo/odoo/blob/fc2e80cb4bcc450762c7ac5cb82a3e2d88062b38/addons/web_editor/static/src/xml/snippets.xml#L31) menu in the Web Editor. Clicking this button will open a dialog (similar to the [media dialog](https://github.com/odoo/odoo/blob/4aab300070e65ca9bb0f7afe22150449baaad275/addons/web_editor/static/src/js/widgets.js#L86)) to select:
* Animation to use (from supported animate.css classes)
* Animation duration (`data-wow-duration`)
* Delay before animation start (`data-wow-delay`)
* Distance relative to browser bottom to start the animation from (`data-wow-offset`)
* Number of times animation is repeated (`data-wow-iteration`)

Once the dialog is saved, it will append the proper classes and data attributes to the snippet that originally triggered the dialog.

One thing I noticed with the animations when we implemented `animate.css` on our website is that if you are in edit mode, the animations do not start & the snippets never appear. If this is the case with WowJS as well, we will likely just disable the animations when in edit mode.

Anyone have any thoughts or pre-existing code we can start from?

lasley:
Derp dunno how I didn't know about [`website_animate`](https://www.odoo.com/apps/modules/10.0/website_animate/). It's not animate.css or WowJS, but it already exists!

lasley:
Sorry for the spasm open/close, but I see a lot of lacking animations in the Odoo SA version & I think I would prefer something a little more adopted (I ❤️  you `animate.css`). 

I also want something that I can test my themes against without an EE contract.

 As such, I'm keeping this on our project schedule.

lasley:
Closing to track in  #365

PR #365, opened 2017-07-19
[ADD] website_wow: Website Animations with WowJS

lasley:
This is a WIP to complete RFC #334 - Website animations on steroids.


website_wow/README.rst
```diff
@@ -0,0 +1,62 @@
+.. image:: https://img.shields.io/badge/license-GPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/gpl-3.0-standalone.html
+   :alt: License: GPL-3
+
+===========
+Website Wow
+===========
+
+Provides animations for your Odoo website using WowJS, which is powered by `animate.css`.
```
yajo: RST uses double backticks for code format.

website_wow/README.rst
```diff
@@ -0,0 +1,62 @@
...
+
+This is created as a replacement to Odoo's `website_animate` module, which uses proprietary
```
yajo: double backticks

website_wow/__manifest__.py
```diff
@@ -0,0 +1,22 @@
...
+# License GPL-3.0 or later (http://www.gnu.org/licenses/gpl.html).
+
+{
+    "name": "Website Wow",
+    "summary": "Provides website animations using WowJS and AnimateJS.",
+    "version": "10.0.1.0.0",
+    "category": "Website",
+    "website": "https://laslabs.com/",
+    "author": "LasLabs, Odoo Community Association (OCA)",
+    "license": "GPL-3",
```
yajo, line 12: Sure?

website_wow/static/src/js/website_wow.editor.js
```diff
@@ -0,0 +1,97 @@
+/* Copyright 2017 LasLabs Inc.
+   License GPL-3.0 or later (http://www.gnu.org/licenses/gpl).
+*/
+
+odoo.define('website_wow.website_wow_editor', function (require) {
```
yajo: `website_wow.editor` seems good enough.

website_wow/static/src/js/website_wow.js
```diff
@@ -0,0 +1,39 @@
...
+   License GPL-3.0 or later (http://www.gnu.org/licenses/gpl).
+*/
+
+odoo.define('website_wow.website_wow', function (require) {
+    'use strict';
+
+    var base = require('web_editor.base');
+    var core = require('web.core');
+
+    var WebsiteWow = {
```
yajo: Could this extend `web.Class`, so we have the common Odoo JS framework boilerplate? Or even better from `web_editor.snippets.animation.Class`?

website_wow/static/src/js/website_wow.editor.js
```diff
@@ -0,0 +1,97 @@
...
+    s_options.registry.o_wow_duration = Mixin.extend({});
+
+    // Delay
+    s_options.registry.o_wow_delay = Mixin.extend({});
+
+    // Offset
+    s_options.registry.o_wow_offset = Mixin.extend({});
+
+    // Iteration
+    s_options.registry.o_wow_iteration = Mixin.extend({});
```
yajo, line 92: Do we really need to have 5 equal instances? Can't a single one do the same with no extra effort?

website_wow/templates/assets.xml
```diff
@@ -0,0 +1,39 @@
...
+            <script src="/website_wow/static/src/js/website_wow.js"
+                    type="application/javascript" />
+        </xpath>
+    </template>
+    <template id="assets_editor"
+              name="WOW Editor Assets"
+              inherit_id="website.assets_editor"
+              >
+        <xpath expr="//link[last()]" position="after">
+            <link href="/website_wow/static/src/less/website_wow.editor.less"
```
yajo: This file seems empty, remove it or fill it.

website_wow/README.rst
```diff
@@ -0,0 +1,62 @@
...
+animations build by Odoo for Odoo. Animate.css is a more widely used framework that is more
+efficient, better maintained, and contains significantly more animations.
+
+.. image:: static/description/screenshot.png?raw=true
+   :alt: Data Sliders
+
+Usage
+=====
+
```
yajo: Fill this section please

website_wow/templates/assets.xml
```diff
@@ -0,0 +1,39 @@
...
+    Copyright 2017 LasLabs Inc.
+    License GPL-3 or later (http://www.gnu.org/licenses/gpl.html).
+-->
+
+<odoo>
+    <template id="assets_frontend"
+              name="WOW Frontend Assets"
+              inherit_id="website.assets_frontend"
+              >
+        <xpath expr="//link[last()]" position="after">
```
yajo: You can do `<xpath expr=".">` if you prefer 😉

website_wow/templates/assets.xml
```diff
@@ -0,0 +1,39 @@
...
+            <link href="/website_wow/static/lib/css/animate.css"
```
yajo: I think you forgot this file.

website_wow/templates/assets.xml
```diff
@@ -0,0 +1,39 @@
...
+                  rel="stylesheet"
+                  type="text/css" />
+        </xpath>
+        <xpath expr="//script[last()]" position="after">
+            <script src="/website_wow/static/lib/js/wow.js"
```
yajo: Missing file too

yajo (changes requested):
Lovely icon! 🤣 🐶 

BTW, Travis is broken, possibly due to https://github.com/OCA/maintainer-quality-tools/pull/465 again.

website_wow/static/src/js/website_wow.editor.js
lasley: Nope - not with the way that the Odoo options selectors work. If I made them all one instance, the options would override each other (duration would override animation name).

Take a look at Odoo SA's website_animation, which does something similar. At least I made them abstract and identical :man_shrugging:

website_wow/templates/assets.xml
lasley: Hmmmm I bet our gitignore skipped things in lib

website_wow/__manifest__.py
lasley: Compatible with the library licensing, which is GPL-3

website_wow/static/src/js/website_wow.js
lasley: I tried this originally, but it will not work with Odoo SA's method of allowing definition of animatable elements in template, because the animations require specific selector. I emulated that method [here](https://github.com/OCA/website/pull/365/files/49144005bd37ab957519aa0e4ccd8fe03728d85d#diff-6d557f05ade49724e2f41b75a4dae43aR30) in order to allow for somewhat cross-compatibility.

There was another issue too, but I can't remember TBH.

website_wow/README.rst
lasley: I was actually going for italics on both of these, because they're titles of things and not really code. IMO code blocks for things like this detract from the statement that is being made due to how much different they are from the text.

website_wow/static/src/js/website_wow.js
yajo: But the selector can have commas, right?

website_wow/templates/assets.xml
```diff
@@ -26,11 +26,6 @@
               name="WOW Editor Assets"
               inherit_id="website.assets_editor"
               >
-        <xpath expr="//link[last()]" position="after">
-            <link href="/website_wow/static/src/less/website_wow.editor.less"
-                  rel="stylesheet"
-                  type="text/less" />
-        </xpath>
         <xpath expr="//script[last()]" position="after">
```
yajo: You can use `<xpath expr=".">` here also if you want.

yajo:
Can't test until you enable sass. You know... https://github.com/OCA/maintainer-quality-tools/pull/465

website_wow/static/src/js/website_wow.js
lasley: Well the issue is that they allow a paradigm of templates changing the selectors that are used. From what I understand of these template variables, they are not accessible in my Javascript, which is what would need to happen for a template to be able to update the targeted selector by using only XML

lasley: The code docs for the section I'm describing:

```
+    How to define an animable element in your theme
 +    ===============================================
 +    <xpath expr="//t[@t-set='o_animate_elements']" position="after">
 +        <t t-set="o_animate_elements" t-value="o_animate_elements + ['.myclass']"/>
 +    </xpath>
```

I need to join the `o_animate_elements` template variable into a selector for use. In order to go with the `web.Class` design, I would need my `selector` class attribute in JS to be based off of this variable in XML. Do you know how to access this XML namespaced variable in my class?

website_wow/README.rst
lasley: (list of things to do - I honestly just kind of started writing this app, so I wasn't really sure where I was going with it until I did)

website_wow/static/src/js/website_wow.js
yajo: Yeah, you can write an inline module, [just like it happened with the CSRF token in v9](https://github.com/odoo/odoo/blob/9.0/addons/web/views/webclient_templates.xml#L306-L310). Then your other module should depend on that one.

lasley: I don't think this helps me here. In the instance of the inline method, it is a different method of variable declaration. 

Additionally, I think it's a slightly different use case - the CSRF token is assigning XML declared within the JS to a JS var. I need a predefined XML variable as a JS variable.

My goal here is to be 100% compatible with the method of `o_animate_elements` declaration that Odoo SA provides in `website_animation`, which will allow for the same animable elements defined in custom themes specifically for `website_animation` will also work with this module without modification.

website_wow/README.rst
yajo: Still missing

yajo (changes requested):
As you can see the wow icon is misaligned.

Also, it seems logical to have all in one menu like:

- Animation
  - Type
  - Duration
  - Offset
  - ...

And maybe [fa-film](http://fontawesome.io/icon/film/) would be a better icon for this.

captura de pantalla de 2017-07-25 09-28-11

I also miss the feature to be able to see the effect just by hovering over the element. It's common in almost all odoo snippet options, and would make it much easier to choose an animation.

lasley:
@Yajo - most comments attended to.

> Also, it seems logical to have all in one menu like:

Unfortunately we're fighting the Odoo editor here. I am not able to get the javascript to work when I nest it. Seems Odoo had the same problem in their's, which also isn't nested - I just can't figure out how to actually show the settings in their module.

image

Another case of fighting the editor. It wants to put a "Disable" there, but there is no disable with the options. I guess maybe we could make it clear the option?

yajo:
Maybe adding a "Default" instead of "disable"?

BTW bots are ❌

lasley:
> BTW bots are ❌

Looks like it's red on 10.0 too & not in this module, so I'd say unrelated. It is in one of my modules though, so I'll handle in another PR

lasley:
Ok so hopefully bots fix with the rebase, or at least give a better error.

lasley:
Welp I give up on testing this. I've been fighting the tour system for a solid 4-5 hours now with no results. Basically the issue is that I cannot get the "Customize" menu to show up after clicking an editable element. The exact same code works outside of the tour, but not within the run block (even when in debugger doing it manually).

website_wow/static/tests/js/website_wow.editor.tour.js
```diff
@@ -0,0 +1,104 @@
...
+        content: 'Go back to edit',
+        trigger: 'a[data-action=edit]',
+        run: function () {
+            $('a[data-action=edit]').click();
+        }
+    }, {
+        content: 'Click an animable element.',
+        trigger: 'main h2:contains("Headline")',
+        run: function () {
+            $('main h2:contains("Headline")').click();
```
lasley: [The problem](https://github.com/OCA/website/pull/365#issuecomment-321942677) is here. I've even tried to get it out of the thread with setTimeout, but no luck.

website_wow/static/tests/js/website_wow.editor.tour.js
```diff
@@ -0,0 +1,104 @@
+/* Copyright 2017 LasLabs Inc.
+   License GPL-3.0 or later (http://www.gnu.org/licenses/gpl).
+*/
+
+odoo.define('website_wow.tour_editor', function (require) {
+    'use strict';
+
+    var tour = require("web_tour.tour");
+    var base = require("web_editor.base");
+    var snippet_editor = require('web_editor.snippet.editor');
```
yajo: Maybe you need to require `website_wow.editor/view` to avoid races in tour?

website_wow/templates/assets.xml
```diff
@@ -0,0 +1,36 @@
...
+        </xpath>
+    </template>
+    <template id="assets_editor"
+              name="WOW Editor Assets"
+              inherit_id="website.assets_editor"
+              >
+        <xpath expr="." position="inside">
+            <script src="/website_wow/static/src/js/website_wow.editor.js"
+                    type="application/javascript" />
+            <script src="/website_wow/static/tests/js/website_wow.editor.tour.js"
```
yajo: Try putting this in assets_frontend.

yajo (changes requested):
Right now the problem is that [PhantomJS does not find the tour declaration](https://travis-ci.org/OCA/website/jobs/266909770#L1694-L1697).

I noticed you are declaring the tour in `assets_editor` and are running the tour in `/` (that's the URL defined in the python test). Try changing any of those.

Another option is that PhantomJS is getting aborted because of some new JS syntax that  it does not recognize. It is very silent in those cases. I checked your code uses old JS so it should be OK, maybe you can check if the wow.js library does too.

Just some ideas... OTOH, it seems `website_blog_category` is failing too...

lasley:
> Right now the problem is that PhantomJS does not find the tour declaration.

Ahh yeah good call, although I've been testing locally & this isn't the big issue unfortunately. I pushed with a debugger statement to allow playing on Runbot:

```javascript
odoo.__DEBUG__.services['web_tour.tour'].run('website_wow_editor')
```

> Another option is that PhantomJS is getting aborted because of some new JS syntax that it does not recognize. It is very silent in those cases. I checked your code uses old JS so it should be OK, maybe you can check if the wow.js library does too.

We're not getting to the wow part yet- I'm just trying to get Phantom to click the block element to make the "Customize" menu for it appear. The click itself doesn't seem to be actually firing the handler while within the tour.

I think the issue stems from the active element selection happening in [`snippet.editor.bind_snipper_click_editor`](https://github.com/odoo/odoo/blob/79477b8d96be47cca5609ce5e4dd84a7e7f64f0a/addons/web_editor/static/src/js/snippets.editor.js#L370). There's a [`setTimeout`](https://github.com/odoo/odoo/blob/79477b8d96be47cca5609ce5e4dd84a7e7f64f0a/addons/web_editor/static/src/js/snippets.editor.js#L380) in there that is just shouting race condition I think.

But then again, I could be totally off. The editor code is dense and doesn't have the best documentation.

The exact same jQuery selection and click code that is failing in debugger while inside of the Phantom tour is working just fine once the tour ends. All editor assets look loaded too from what I can tell.

The click code is:
```javascript
$('main h2:contains("Headline")').click()
```

If you do that right after debugger, the Phantom tour actually continues.

website_wow/templates/assets.xml
```diff
@@ -0,0 +1,37 @@
...
+                    type="application/javascript" />
+        </xpath>
+    </template>
+    <template id="assets_editor"
+              name="WOW Editor Assets"
+              inherit_id="website.assets_editor"
+              >
+        <xpath expr="." position="inside">
+            <script src="/website_wow/static/src/js/website_wow.editor.js"
```
yajo, line 31: This is required by `website_wow.editor.tour.js`, so not much sense to have both in different assets.

website_wow/static/tests/js/website_wow.editor.tour.js
```diff
@@ -0,0 +1,108 @@
...
+    require('website_wow.view');
+
+    tour.register("website_wow_editor", {
+        url: '/page/contactus?debug=assets&enable_editor=1',
+        test: true,
+        wait_for: base.ready(),
+    }, [{
+        content: 'Create an animable element.',
+        trigger: 'div[name=Cover]',
+        run: function () {
```
yajo: Possibly better:

```js
{
  ...,
  trigger: ".oe_snippet[name='Text Block']",
  run: "drag_and_drop"
}
```

That's the way a human would create a snippet.

website_wow/static/tests/js/website_wow.editor.tour.js
```diff
@@ -0,0 +1,108 @@
...
+            $('button[data-action=save]').click();
+            // But our element causes some weird error, so discard (it's still saved somehow...)
+            $('button[data-action=cancel]').click();
+            // Click OK
+            $('span:contains("Ok")').parent().click();
+        }
+    }, {
+        content: 'Go back to edit',
+        trigger: 'a[data-action=edit]',
+        run: function () {
```
yajo: Possibly better: `run: "click"`

website_wow/static/tests/js/website_wow.editor.tour.js
```diff
@@ -0,0 +1,108 @@
...
+            $('a[data-action=edit]').click();
+        }
+    }, {
+        content: 'Click an animable element.',
+        trigger: 'main h2:contains("Headline")',
+        run: function () {
```
yajo: In my by-hand tests, the customize menu appears automatically after drag and dropping, so possibly you can skip this step. Otherwise, maybe better `run: "click"`

website_wow/static/tests/js/website_wow.editor.tour.js
```diff
@@ -0,0 +1,108 @@
...
+
+    var tour = require("web_tour.tour");
+    var base = require("web_editor.base");
+    var snippet_editor = require('web_editor.snippet.editor');
+
+    require('website_wow.editor');
+    require('website_wow.view');
+
+    tour.register("website_wow_editor", {
+        url: '/page/contactus?debug=assets&enable_editor=1',
```
yajo: `debug=assets` should not be required, although it can help debugging these problems...

yajo (commented):
OK, I'm giving you some suggestions that you can check out for making it work, just trying to help you get unlocked 😊 

- Try to imitate a human behavior.
- Better use [action helpers](https://github.com/odoo/odoo/blob/10.0/addons/web_tour/static/src/js/tour_manager.js#L91-L113) wherever possible, they tend to work better and are easier to maintain.
- [PhantomJS still cannot see your tour](https://travis-ci.org/OCA/website/jobs/267385048#L1895-L1899). That's the first thing you need to fix.

See in line comments too.

website_wow/templates/assets.xml
lasley: It seems that in order to require something you also need to have it included in assets. Removing this yields:

```
warning: Some modules could not be started 
Missing dependencies:    ["website_wow.editor"] 
Non loaded modules:      ["website_wow.tour_editor"] 
Debug:                   {website_wow.tour_editor: {…}}
```

lasley:
You're freaking amazing @Yajo, thank you. I think the trick was the use of the `drag_and_drop` in the run, which I totally had no idea about.

Manual tests were 🍏 ; hoping for the same on CI!

website_wow/templates/assets.xml
yajo: I mean that I'd put `website_wow.editor.tour.js` inside `assets_editor` instead of `assets_frontend`, because anyways it will fail in frontend when you're not an editor.

Besides, it could go in demo data to avoid overloading the website.

yajo (changes requested):
Glad to help! 😊 Although it seems bots are still sad 😞

website_wow/demo/assets_demo.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<!--
+    Copyright 2017 LasLabs Inc.
+    License GPL-3 or later (http://www.gnu.org/licenses/gpl.html).
+-->
+
+<odoo>
+    <template id="assets_editor"
```
yajo: Don't duplicate your XMLID! This one should be `assets_editor_demo`

website_wow/demo/assets_demo.xml
```diff
@@ -0,0 +1,20 @@
...
+              name="WOW Editor Assets"
+              inherit_id="website.assets_editor"
+              >
+        <xpath expr="." position="inside">
+            <script src="/website_wow/static/src/js/website_wow.editor.js"
```
yajo: And remove this from here. This goes into the production assets.

yajo (changes requested):

yajo:
Weird... [`website_blog_category` tour is failing](https://travis-ci.org/OCA/website/jobs/275069985#L2768)... I think the addon was yours too, any idea? 🤔

lasley:
To me this looks like the [timeout issue that @sylvain-garancher found](https://github.com/OCA/report-print-send/pull/95#issuecomment-325097584). 

It's been cropping up everywhere randomly, and I bet that one of the [four TransactionCase](https://github.com/OCA/website/search?utf8=%E2%9C%93&q=TransactionCase&type=) tests in this repo is causing it (one of which is `website_blog_category`). Test commit incoming.

lasley:
Well crap, so much for a simple solution. I'm pretty sure this has to be something I'm causing here then. The test failing is more of a cascade if anything, starting with [website_logo](https://travis-ci.org/OCA/website/jobs/275538843#L1468). That typically means something wonky in the request.

I'll take a look soon.

yajo:
Thanks. Usually HttpCases are required when doing any frontend or controller tests. Otherwise your requests will use a different non-testing cursor, and tests will go nuts.

website_wow/static/src/xml/snippet_options.xml
```diff
@@ -0,0 +1,537 @@
...
+            <t t-set="o_animate_elements"
+               t-value="['.o_animable', 'section', 'img', '.fa', '[class^=\'col-md\']', '.btn']" />
+            <t t-set="o_wow_selector"
+               t-value="', '.join([s + ':not(.o_not-animable)' for s in o_animate_elements])" />
+
+            <!-- Animations -->
+            <div data-js="o_wow" t-att-data-selector="o_wow_selector">
+                <li class="dropdown-submenu o_wow_options">
+                    <a href="#">
+                        <i class="fa fa-film">Wow!</i>
```
hughesbm: Placing the menu text outside the `<i>` tag will keep the styling consistent with the rest of the menu. (same thing applies to the rest of your top-level menu titles)

E.g:
```
<a href="#"><i class="fa fa-film" />Wow!</a>
```

website_wow/static/src/xml/snippet_options.xml
```diff
@@ -0,0 +1,537 @@
...
+                        </li>
+                    </ul>
+                </li>
+            </div>
+
+            <!-- Durations -->
+            <div data-js="o_wow_duration" t-att-data-selector="o_wow_selector">
+                  <li class="dropdown-submenu">
+                      <a href="#">
+                          <i class="fa fa-clock">
```
hughesbm: Closing the `<a>` tag here instead of later (line 391) prevents the spacing issue above the first item in the sub-menu.

Also, `fa-clock` doesn't appear to be a valid font-awesome icon, but `fa-clock-o` is.

These things apply to the rest of your top-level menus too.
```
<a href="#"><i class="fa fa-clock-o" />Wow Duration</a>
```

hughesbm (commented):
Your menu fonts and spacing are a little off (unless you're intentionally going for doge weirdness).

Currently:

screen shot 2017-09-28 at 10 39 09 am

After suggested fixes:

screen shot 2017-09-28 at 10 37 54 am

lasley:
Thanks @hughesbm - updated

lasley:
I was hoping that coming back to this with a fresh mind would help, but I have come to the conclusion that PhantomJS or the Odoo tour running is simply incompatible with something. The tour works in all browsers, the issue is simply with the promise that says the tour is ready. This promise is outside of the scope of this module, and has already caused a significant amount of debugging. 

Watch the tour in action with the below on Runbot:

```
odoo.__DEBUG__.services['web_tour.tour'].run('website_wow_tour_editor')
```

I give up, and have removed the import for `test_ui.py` from `tests/__init__.py`. I propose this for merge as is - the module works well and IMO there's no reason to hold it back because Odoo sucks. We don't even lose test coverage because we don't have it in JS.

website_wow/tests/test_ui.py
```diff
@@ -0,0 +1,22 @@
...
+@common.post_install(True)
+class TestUi(HttpCase):
+
+    post_install = True
+    at_install = False
+
+    def test_wow(self):
+        tour_module = 'odoo.__DEBUG__.services["web_tour.tour"]'
+        self.phantom_js(
+            url_path='/',
```
yajo, line 18: Have you tried with `/?enable_editor=1`?

yajo (commented):
Do you know why we don't have a runbot here?

website_wow/tests/test_ui.py
lasley: Yep, and I tried with the `?debug=assets&enable_editor=1`. Basically every combination I can think of, ultimately settling on this because I can't find an example of Odoo SA declaring `url_path` with args in the JS

lasley:
> Do you know why we don't have a runbot here?

Hrmmm seems like it's stuck in pending. I amend/pushed, hopefully that kicks it in gear

.travis.yml
```diff
@@ -21,8 +21,8 @@ env:
   - secure: eNfbQI7Pkp3mjTFQgjrVVldmyE/Yn3+iCMVykhXppC85dJaCFMpjwPUfZa7p38tnnH7jhng2JL2tAXnJ4uOry0B3m9xYR7fNYXg/9uj1988FHvpDhfDFEKcMF8s70mr11HRijj6U8p6WDt/0ygD6aZ0LJzWzxhMFecFavOb6XOw=
   matrix:
   - LINT_CHECK="1"
-  - TESTS="1" ODOO_REPO="odoo/odoo"
-  - TESTS="1" ODOO_REPO="OCA/OCB"
+  - TESTS="1" ODOO_REPO="odoo/odoo" EXCLUDE="website_wow"
+  - TESTS="1" ODOO_REPO="OCA/OCB" EXCLUDE="website_wow"
```
lasley, line 25: Note I think this is a bit cleaner, but I think it means that you'll have to actually install this module on Runbot vs. it being pre-installed. I'm indifferent on whether this commit should stay, and am interested in others' opinions here

simahawk:
anyone willing to move this fwd? Maybe porting it to v11 or v12? @Yajo what's your POV?

yajo:
This is stale. Closing. Whoever wants, get the commits and create a new PR. Thanks!

PR #366, opened 2017-07-20
[FIX] Remove lib from gitignore

lasley:
* Lib folder is standard for web repo static directories, so should not be in gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,7 +13,6 @@ build/
 develop-eggs/
 dist/
 eggs/
-lib/
 lib64/
 parts/
 sdist/


pedrobaeza:
But shouldn't be inside static/src?

lasley:
It's still ignored - Jairo noticed it here https://github.com/OCA/website/pull/365#discussion_r128441927

All I did was remove this line from the gitignore, and git saw the files to commit.

simahawk:
I had the same issue and I had to add the folder w/ `-f` to force it. I tried several path combo in .gitignore but not luck.
Any real reason to keep this ignore?

simahawk (approved):

pedrobaeza (approved):
OK then.

yajo (approved):
I expected this behaviour to happen only on `**/lib/` but it seems git thinks otherwise...

PR #368, opened 2017-07-25
[MIG][website_snippet_marginless_gallery] Migrate to v9

yajo:
- Relicense to LGPL.
- Move tour to demo asset.
- Remove precompiled CSS.
- Update snippet options JS API.
- Update tour's API and definition.

@Tecnativa


website_snippet_marginless_gallery/__openerp__.py
```diff
@@ -2,14 +2,13 @@
 {
     "name": "Marginless Gallery Snippet",
     "summary": "Add a snippet to have a marginless image gallery",
-    "version": "8.0.1.0.1",
+    "version": "9.0.1.0.1",
```
pedrobaeza: 9.0.1.0.0

website_snippet_marginless_gallery/tests/__init__.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# © 2015 Grupo ESOC Ingeniería de Servicios, S.L.U. - Jairo Llopis
-# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+# Copyright 2015-today Jairo Llopis <jairo.llopis@tecnativa.com>
```
pedrobaeza: Use exact year, not today

pedrobaeza (changes requested):
Tests are broken and I can't try on runbot due to `compass installation`

website_snippet_marginless_gallery/tests/__init__.py
yajo: I got inspired by https://github.com/facebook/react/commit/bef45b0b1a98ea9b472ba664d955a039cf2f8068 and the assumption that Facebook has good lawyers.

pedrobaeza: We don't use this rule in OCA (which is abusing), so change it

pedrobaeza:
I still can't try on runbot.

pedrobaeza:
I have been trying on runbot but I'm not able to change the background image of any tile, as "Customize" doesn't appear when selecting a tile.

yajo:
I guess it should be fixed now, let's try when runbot builds.

pedrobaeza:
@Yajo, I'm still not able to customize tile background image.

pedrobaeza (approved):
Now is finally working.

@rafaelbn please review

rafaelbn (approved):
Super 👍 . Tested in runbot

yajo:
Squashed for merge.

PR #383, opened 2017-09-22
[10.0][ADD] website_snippet_barcode: Add module

hughesbm:
# Website Snippet - Barcode

This module provides a barcode snippet to website, allowing the generation and display of various types of barcodes.

Uses the report module's [barcode controller route](https://github.com/odoo/odoo/blob/b7c7591bfba41eba5be43ec490d6da9a2c7368fe/addons/report/controllers/main.py#L54) to generate barcodes w/ several options.

## Usage

1. Edit a web page
2. Find Barcode in the Blocks sidebar, in the "Inner Content" section.
3. Drag the Barcode snippet to the desired location
4. Adjust options as needed
5. Save the page

### Barcode Options

**Barcode Type**

* All barcode types available in the report module are available in this snippet.
* Certain barcode types are not compatible with certain values, but this will be apparent during selection.
* *Default: QR*

**Barcode Value**

* This is the value represented by the barcode. Accepts the current page URI or a custom string value.
* *Default: URI of current page*

**Aspect Ratio**

* This setting adjusts the width-to-height ratio of the barcode image.
* *Default: 1:1 (square)*

**Human Readable**

* This setting controls the display of barcode values in human-readable format.
* **None**: Human-readable value is not displayed.
* **Text**: Display the human-readable value as a caption below the
  image.
* **Image**: *Not supported by all barcode types.* Include the human-readable
  value in the generated image.
* *Default: None*


hughesbm:
@lasley review please 😄

website_snippet_barcode/templates/snippet_template.xml
```diff
@@ -0,0 +1,65 @@
...
+
+                <li data-select_humanreadable="true">
+                    <a href="#"><i class="fa fa-user"/>Human Readable</a>
+                </li>
+            </div>
+        </xpath>
+    </template>
+
+    <template id="barcode_snippet" name="Barcode">
+        <div class="s_barcode col-md-12">
```
lasley: Take a look at the current "Image Floating" snippet. We're going for something like that, vs the full-screen block that we have here.

```xml
    <div class="o_image_floating o_margin_l pull-right s_image_floating">
        <div class="o_container">
            <img class="img img-rounded img-responsive" src="/web/image/website.s_image_floating_default_image" alt="Odoo CMS- Sample image floating" data-original-title="" title="">
        </div>
        <div class="o_footer">
            <small class="text-muted o_default_snippet_text">A great way to catch your reader's attention is to tell a story. Everything you consider writing can be told as a story.</small>
        </div>
    </div>
```

The top part would be the barcode, bottom would be the barcode numbering/text if the option is on.

website_snippet_barcode/static/src/js/website_snippet_barcode.js
```diff
@@ -0,0 +1,56 @@
...
+                             '&humanreadable=' + Number(options.humanreadable === 'true');
+
+            this.$target.find('img').attr('src', barcodeSrc);
+            _.each(options, $.proxy(this._setDataAttribute, this));
+            // Set height to ensure proper editor overlay sizing
+            this.$target.height(height);
+        },
+
+        _getOptions: function ($target) {
+            return Object.keys(this.defaults).reduce(function (obj, key) {
```
lasley: Take a look at [`_.reduce`](http://underscorejs.org/#reduce)

website_snippet_barcode/static/src/js/website_snippet_barcode.js
```diff
@@ -0,0 +1,56 @@
...
+        start: function () {
+            this.defaults.width = this.$target.width();
+            this.renderBarcode();
+            return this._super();
+        },
+
+        renderBarcode: function () {
+            var options = _.defaults(this._getOptions(this.$target), this.defaults);
+            var height = Math.round(options.width / options.aspectratio);
+            var barcodeSrc = '/report/barcode?type=' + options.type +
```
lasley: I think template strings (template literals in ES6) would be better here:

```javascript
var barcodeSrc = `/report/barcode`
    + `?type=${ options.type }`
    + `&value=${ options.value }`
    + `&width=${ options.width }`
    + `&height=${ options.height }`
    + `&humanreadable=${ options.humanreadable }`;
```

Which would equal:
```
"/report/barcode?type=QR&value=https%3A%2F%2Fdeveloper.mozilla.org%2Fen-US%2Fdocs%2FWeb%2FJavaScript%2FReference%2FTemplate_literals&width=500&height=undefined&humanreadable=false"
```


Given default options

website_snippet_barcode/static/src/js/website_snippet_barcode.js
```diff
@@ -0,0 +1,56 @@
...
+                             '&value=' + options.value +
+                             '&width=' + options.width +
+                             '&height=' + height +
+                             '&humanreadable=' + Number(options.humanreadable === 'true');
+
+            this.$target.find('img').attr('src', barcodeSrc);
```
lasley: We should give the barcode image a special class and use that for the selector. It will also help if someone wants to custom style the barcode

website_snippet_barcode/static/src/js/website_snippet_barcode.js
```diff
@@ -0,0 +1,56 @@
...
+            _.each(options, $.proxy(this._setDataAttribute, this));
+            // Set height to ensure proper editor overlay sizing
+            this.$target.height(height);
```
lasley: Hmmmm I feel like this might be sweeping something under a rug. After we switch the xml structure as indicated below, let's see if this is still needed.

website_snippet_barcode/static/tests/js/website_snippet_barcode.js
```diff
@@ -0,0 +1,19 @@
+/* Copyright 2017 LasLabs Inc.
+ * License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html). */
+
+odoo.define_section('website_snippet_barcode', ['module_name.ExportedObject'], function(test) {
+    "use strict";
+
+    test('It should demonstrate a PhantomJS test for web (backend)',
```
lasley: This is for backend test; you'll need a website tour instead

lasley (changes requested):

website_snippet_barcode/templates/snippet_template.xml
hughesbm: Works for me. A couple related questions:

* If I add the footer, should I remove the option of generating the barcode image with the human-readable portion? Or make it an option (in-image vs footer text)?

* I added the option to use it as a block (e.g. replacing one of the image/text blocks in the "Three Columns" snippet) because the "Image Floating" placement/sizing options seemed overly restrictive. Should I keep that or no? It does making resizing a little more complicated (as you noted in another comment).

lasley: We would probably just optionally show/hide the footer. I think the option is good.

Regarding block - I'm not sure how useful a block barcode will be. I've never personally seen a barcode spanning the width of an entire page, and I'm not sure what elements we could combine it with that would fit all use cases.

What restrictions were you noticing with floating image?

website_snippet_barcode/static/src/js/website_snippet_barcode.js
lasley: Bah nvm Phantom probably won't support this

website_snippet_barcode/templates/snippet_template.xml
hughesbm: A full-width barcode would be pretty silly, definitely. I was placing it in the more structured snippets like three columns, references, features, and a couple others. Not sure how often that's likely to be an appropriate place for a barcode, though.

It's possible that the floating image placement just seemed restrictive when working with a bare-bones demo page without much content. I'm happy to remove the block option and have it behave like a floating image, if it works well in real-life use.

website_snippet_barcode/static/src/js/website_snippet_barcode.js
hughesbm: 😭

website_snippet_barcode/static/src/js/website_snippet_barcode.js
hughesbm: This is fixed in the new commit-- adding a placeholder image src to the tag in the template accomplishes the same thing (preventing the initial overlay from having a height of 1px).

website_snippet_barcode/static/tests/js/website_snippet_barcode.js
hughesbm: Thanks. I'll add the tour next.

website_snippet_barcode/static/src/js/website_snippet_barcode.js
```diff
@@ -0,0 +1,64 @@
...
+
+    var animation = require('web_editor.snippets.animation');
+
+    animation.registry.barcode = animation.Class.extend({
+        selector: '.s_barcode',
+
+        defaults: {
+            type: 'QR',
+            value: encodeURIComponent(window.location.href.split('#')[0]),
+            aspectratio: '1',
```
lasley: Oops I should have noticed this first time around. Should use camelCase for JS:
```
aspectRatio: ...,
humanReadable: ...,
```

lasley (approved):
One comment. Test failures look to be unrelated

hughesbm:
Travis is red due to an issue affecting some other PRs as well, likely related to https://github.com/OCA/website/pull/365#issuecomment-329564620

website_snippet_barcode/static/src/js/website_snippet_barcode.js
hughesbm: Oh. Derp. Thanks for catching that.

gonzaloruzafa (approved):
Great module. Some glitches on some aspect ratio selection, that just happened once and I could not replicate. +1

chienandalu (approved):
Cool one

PR #390, opened 2017-10-06
[8.0] [FIX] website_portal*: fix link to monksoftware website

LeartS:
Unfortunately we forgot to renew our domain/brand wearemonk-dot-com and the domain has been squatted by a less than reputable agent (pornography); so we are going through every one of our contributions that have that link somewhere, and replacing it with our real website.

### DIFF
--- a/website_portal/__openerp__.py
+++ b/website_portal/__openerp__.py
@@ -7,7 +7,7 @@
     'license': 'LGPL-3',
     'author': 'Odoo SA, MONK Software, Antiun Ingeniería S.L.',
     'website': (
-        'https://www.odoo.com/, http://www.wearemonk.com, '
+        'https://www.odoo.com/, http://www.monksoftware.it, '
         'https://github.com/OCA/website'
     ),
     'depends': [
--- a/website_portal/__openerp__.py
+++ b/website_portal/__openerp__.py
@@ -5,10 +5,16 @@
     'summary': 'Account Management Frontend for your Customers',
     'version': '8.0.1.0.0',
     'license': 'LGPL-3',
-    'author': 'Odoo SA, MONK Software, Antiun Ingeniería S.L.',
+    'author': (
+        'Odoo SA'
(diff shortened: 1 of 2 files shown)


pedrobaeza (approved):

website_portal/__openerp__.py
```diff
@@ -7,7 +7,7 @@
     'license': 'LGPL-3',
     'author': 'Odoo SA, MONK Software, Antiun Ingeniería S.L.',
```
pedrobaeza: Please include OCA here as usual.

pedrobaeza (changes requested):

website_portal/__openerp__.py
```diff
@@ -5,10 +5,16 @@
...
+    'author': (
+        'Odoo SA'
+        ', MONK Software'
+        ', Antiun Ingeniería S.L.'
+        ', Odoo Community Association (OCA)'
+    ),
     'website': (
-        'https://www.odoo.com/, http://www.monksoftware.it, '
-        'https://github.com/OCA/website'
+        'https://www.odoo.com/'
```
pedrobaeza: Sorry for not noticing before, but this URL should be unique and parseable or pypi upload will fail. We are right now putting the repository URL

pedrobaeza (changes requested):

website_portal/__openerp__.py
LeartS: Sorry, I'm not clear on this. What do you mean by 'unique'? We can't have the same URL in multiple addons? Or it only counts for the first URL?

pedrobaeza: You have to leave only one URL, not 3.

pedrobaeza:
Are you going to change this or should I close it?

simahawk:
no answer since quite some time. 
I'm closing this, feel free to re-open if needed ;)
Thanks

PR #391, opened 2017-10-12
add missing space before caret

zeroheure:
Add space missing between company name and caret. It doesn't hurt to have 2 consecutive space in case there is already one (like between "Privacy Policy" and "Legal Advice"), because html hide them.

PR #393, opened 2017-10-13
[imp] website_legal: show Company Registry if valued

zeroheure:
Add field company_registry. 
Please note : curently that field is not checked in Odoo and can contain any other legal information. Useful to provide required legal mention in various country.

### DIFF
--- a/website_legal_page/__manifest__.py
+++ b/website_legal_page/__manifest__.py
@@ -7,7 +7,7 @@
     'name': "Website Legal Page",
     'description': 'Add legal information, such as privacy policy',
     'category': 'Website',
-    'version': '10.0.1.1.0',
+    'version': '10.0.2.0.0',
     'depends': [
         'website',
     ],
--- a/website_legal_page/__manifest__.py
+++ b/website_legal_page/__manifest__.py
@@ -7,7 +7,7 @@
     'name': "Website Legal Page",
     'description': 'Add legal information, such as privacy policy',
     'category': 'Website',
-    'version': '10.0.2.0.0',
+    'version': '10.0.1.3.0',
     'depends': [
         'website',
     ],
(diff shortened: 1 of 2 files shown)


yajo:
I'm not sure all customers will want this displayed by default...

zeroheure:
In France it  is mandatory. How can I add this as a l10n files ?

website_legal_page/__manifest__.py
```diff
@@ -7,7 +7,7 @@
     'name': "Website Legal Page",
     'description': 'Add legal information, such as privacy policy',
     'category': 'Website',
-    'version': '10.0.1.1.0',
```
simahawk, line 10: shouldn't this be "2" instead of "3"?

website_legal_page/views/website_legal.xml
```diff
@@ -48,6 +48,11 @@
               </p>
             </div>
             <div>
+              <t t-if="res_company.company_registry">
```
simahawk, line 51: I agree w/ @Yajo: should be conditional. Option: add a flag on website panel "Show company registry".

simahawk (changes requested):

website_legal_page/views/website_legal.xml
yajo: Maybe in the res.company form view, along with the `company_registry` field, a checkbox: "Display in website legal advice page"

zeroheure:
Version is "3" because of previous PR #391

I wonder... as there is different legal requirements per country, shouldn't this module be moved in each l10n project ? For example the /OCA/l10n-france/l10n_fr_siret add some fields in company form. These fields must be printed on the legal page.

yajo:
I think you answered yourself: if the requirements are for a single country, yes. If they are present in many, not. 😃

simahawk:
@zeroheure so, do you still need this?
Closing but feel free to re-open or submit another PR if needed ;)
Thanks!

Issue #400, opened 2017-12-06
Website Snippet - Data Slider, how to limit qty products?


joseponzo:
Hi
Please help. I have 30 products in the store and the 30 loads the slider. I have data_limit in 10, I change it to 5 and it's the same load all the products published. I need to configure something more?.
environment: odoo 10 enterprise, ubuntu 16, nginx. ssl.
Ragards

PR #401, opened 2017-12-07
11.0 website snippet data slider

smurf-U:
Migrate Module and Fix https://github.com/OCA/website/issues/400 issues


pedrobaeza:
Please squash together "OCA Transbot..." commits following https://github.com/OCA/maintainer-tools/wiki/Merge-commits-in-pull-requests

website_snippet_data_slider/README.rst
```diff
@@ -0,0 +1,104 @@
...
+* `data_title` - Text to use as main snippet header
+* `data_title_tag` - HTML element type to use for title element (such as `h1`)
+* `data_title_class` - Class to use for the title element
+* `data_uri_field` - Field to use for the link
+* `data_container_width` - Width of the outer container, default 90%
+* `prevArrow` & `nextArrow` - HTML or jQuery selector of slider arrows
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/186/10.0
```
simahawk: /11.0

website_snippet_data_slider/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
```
simahawk: remove encoding on py files

website_snippet_data_slider/__manifest__.py
```diff
@@ -0,0 +1,27 @@
...
+
+{
+    "name": "Website Snippet - Data Slider",
+    "summary":
+        "Abstract data slider for use on website."
+        "  Primary use (and default implementation) is product slider.",
+    "version": "11.0.1.0.0",
+    "category": "Website",
+    "website": "https://laslabs.com/",
+    "author": "LasLabs, Odoo Community Association (OCA), Kaushal Prajapati",
```
simahawk: OCA should stay at the end, always. Plus, we don't put 'migrators' inside authors AFAIK

website_snippet_data_slider/controllers/main.py
```diff
@@ -0,0 +1,39 @@
...
+        res = []
+        _logger.debug(kwargs)
+        domain = json.loads(kwargs.get('domain', "[]"))
+        fields = json.loads(kwargs.get('fields', "[]"))
+        limit = kwargs.get('limit', None)
+        if limit:
+            limit = int(limit)
+        for rec_id in request.env[model].search(domain, limit=limit):
+            res.append({
+                k: getattr(rec_id, k, None) for k in fields
```
simahawk: don't use `getattr` access the field as a key: `record[k]` (rec_id is indeed a record)

website_snippet_data_slider/static/src/js/data_slider.js
```diff
@@ -0,0 +1,134 @@
...
+        },
+        
+        // It loops parses JSON records and calls _handleRecord on each
+        handleRecords: function(records) {
+            _.each(JSON.parse(records), $.proxy(this._handleRecord, this)); 
+        },
+        
+        // Accepts a record object and appends to slick
+        _handleRecord: function(record) {
+            var $img = $('<img>');
```
simahawk: I'd add a TODO here to replace this element creation w/ pure JS: much more performant.

simahawk (changes requested):

website_snippet_data_slider/controllers/main.py
smurf-U: we can't access data record using an index. ex:`record[k]` is not possible. And in this scenario `getattr` is useful.

simahawk: `k` is not and index, is a field name :wink: 
in any case if an index is there, then is broken, because you cannot slice a single record (eg: record[0]).
Accessing data w/ getattr is not even safe. Check this https://www.odoo.com/slides/slide/10-rules-for-safer-code-398 (slide 41)

PR #402, opened 2017-12-15
[10.0][ADD] website_form_builder: Exactly what the title says

yajo:
This is WIP until tours pass, but you can start playing with it!


pedrobaeza:
@lasley I think you will love this

lasley:
Oh snap!

website_form_builder/README.rst
```diff
@@ -0,0 +1,106 @@
...
+   * Choose the field where to store custom fields data in *Field for custom
+     form data*. If you leave this one empty and the model is a mail thread,
+     a new message will be appended with that custom data.
+
+#. In the *Fields* tab, there's a new column called *Blacklisted in web forms*.
+   It's a security feature that forbids form submitters to write to those
+   fields. When you create a new website form, all its model fields are
+   automatically whitelisted for the sake of improving the UX. If you want to
+   have higher control, come back here after creating the form and blacklist
+   any fields you want.
```
lasley: Hmmm this is opposite of the operation of the existing `website_form`, which is a whitelist. More thoughts probably after reading the code, this is mainly a reminder for me to update with more thoughts. If I don't, remind me to update with more thoughts!

website_form_builder/README.rst
```diff
@@ -0,0 +1,106 @@
...
+   :target: https://runbot.odoo-community.org/runbot/186/10.0
+
+Known issues / Roadmap
+======================
+
+* These type of fields will not appear, they are forbidden since they make no
+  sense in this module's context, or a correct implementation would be adding
+  not much value while adding lots of complexity:
+
+  * ``id``
```
pedrobaeza, line 70: What about create_date, write_date, create_uid, write_uid? They should be forbidden also I think.

lasley, line 70: IMO we should prevent all the magic fields

website_form_builder/README.rst
```diff
@@ -0,0 +1,106 @@
...
+  * ``one2many``
+  * ``reference``
+  * ``serialized``
+
+* You should include https://github.com/odoo/odoo/pull/21628 in your
+  installation to get a better UX when a user has already sent a form and
+  cannot resend it.
+
+* To edit any ``<label>`` text, you need to click twice. Review the problem
```
pedrobaeza, line 85: Only in Firefox?

website_form_builder/static/src/js/snippets.js
```diff
@@ -0,0 +1,460 @@
...
+    var options = require('web_editor.snippets.options');
+    var _t = core._t;
+    var widgets = require("website_form_builder.widgets");
+    var data = require("web.data");
+
+    core.qweb.add_template(
+        "/website_form_builder/static/src/xml/snippets.xml"
+    );
+
+    var _fields_asked = {},
```
lasley: IMO we should use JS naming conventions & camelCase the vars/methods

website_form_builder/static/src/js/snippets.js
```diff
@@ -0,0 +1,460 @@
...
+        },
+
+        /**
+         * Inject a new field from the model into the form.
+         *
+         * @param {name: field_name, field: field_definition} info
+         */
+        add_model_field: function (info) {
+            var relational_data = [],
+                template = _.str.sprintf(
```
lasley: TIL `_.str`

website_form_builder/static/src/js/snippets.js
```diff
@@ -0,0 +1,460 @@
...
+        },
+
+        /**
+         * Selectable options for relational fields.
+         */
+        relational_options: function (field) {
+            var domain = [], context = base.get_context();
+            // Domain might contain un-evaluable literals
+            try {
+                domain = new data.CompoundDomain(field.domain || []).eval()
```
lasley: Missing `;` * 3 other instances below

website_form_builder/static/src/js/widgets.js
```diff
@@ -0,0 +1,166 @@
...
+             * Store needed field information.
+             *
+             * @param {DOMElement} $field DOM element of the field
+             * that wants a new default value.
+             */
+            init: function (parent, options, field) {
+                this.field_html = $(field).html();
+                options = $.extend({}, {
+                    title: _t("Set field's default value"),
+                    size: "small",
```
lasley: trailing comma no bueno in JS

lasley (approved):
Freaking awesome @Yajo. Haven't played with it functionally yet, but excellent code. A few minor things inline

website_form_builder/README.rst
pedrobaeza: I prefer this approach for not having to manually add each of the fields we wish to add or supply glue modules for each possible model.

website_form_builder/README.rst
```diff
@@ -0,0 +1,106 @@
...
+
+To configure this module, you need to:
+
+#. Have *Administration / Settings* privileges.
+#. Go to *Settings > Activate developer mode*.
+#. Go to *Settings > Technical > Database Structure > Models*.
+#. Search for the model you want to manage website form access for.
+#. When you find it, it will have a *Website Forms* section where you can:
+
+   * Allow the model to get forms, by checking *Allowed to use in forms*.
```
pedrobaeza, line 33: We need to provide a built-in list of allowed models without depending on the corresponding modules (crm.lead if present, hr.recruitment if present...). We can use a similar approach to the one used in website_multi_theme.

lasley:
Hmmm but modules that are naive of forms, which inherit a model that is be exposed by a form, 
typically expect that the data is coming from an authenticated source. An example is prescription or doctor verification. This data can now be submitted by users that are not authenticated. 

-Dave Lasley

website_form_builder/README.rst
lasley: (moved inline)

Hmmm but modules that are naive of forms, which inherit a model that is be exposed by a form, typically expect that the data is coming from an authenticated source. An example is prescription or doctor verification. This data can now be submitted by users that are not authenticated.

pedrobaeza: Models are opt-in, fields are opt-out

lasley: In the example I was mentioning, the validation field is on the partner. I see a lot of potential uses for opting the partner in, which would opt in those fields too.

I'm actually kind of confused though. This black list is circumventing the pre-existing whitelist of `website_form`? Why is the whitelist also being used [in the demo data](https://github.com/OCA/website/pull/402/files#diff-3919e194d50de1418256b036abcbda13R13)?

website_form_builder/README.rst
yajo: Chrome worked without patch out of the box. Besides, this bug affects EE `website_form_editor`, just try in the enterprise runbot.

website_form_builder/README.rst
yajo: Well, the point here is that, in an awesomely rare exception, `website_form` is completely undocumented (🤣). I tried to reuse the existing engine as much as possible, and it actually makes sense. Let me explain what it does:

- `ir.model` has [some fields to whitelist the model to be used in website forms](https://github.com/odoo/odoo/blob/cfe0ebcc59318a931e0f2592283ac933ee842ba3/addons/website_form/models/models.py#L19-L21).
- [`ir.model.fields` has a `website_form_blacklisted` boolean that defaults to `True`](https://github.com/odoo/odoo/blob/cfe0ebcc59318a931e0f2592283ac933ee842ba3/addons/website_form/models/models.py#L107), so all model fields are blacklisted by default.
- [When a model is not whiltelisted, the controller aborts](https://github.com/odoo/odoo/blob/cfe0ebcc59318a931e0f2592283ac933ee842ba3/addons/website_form/controllers/main.py#L24). No models are whitelisted by default in `website_form` addon.
- [When a field is blacklisted, but the model is whitelisted, the field is considered *extra data* (like a custom field that does not exist)](https://github.com/odoo/odoo/blob/cfe0ebcc59318a931e0f2592283ac933ee842ba3/addons/website_form/controllers/main.py#L134-L144).
- The `formbuilder_whitelist` method is a [public method only available to website designers](https://github.com/odoo/odoo/blob/cfe0ebcc59318a931e0f2592283ac933ee842ba3/addons/website_form/models/models.py#L93), which means that those users should be able to decide which fields they want to make writeable. The most UX-friendly way of expressing such desire is when you add a field to a website form, so [when you save the form, any model fields found there are whitelisted automatically](https://github.com/OCA/website/pull/402/files#diff-95cb4eb71cc38363ac6f31057c94ad32R185).
- [Core addons use such method to enable specific models](https://github.com/odoo/odoo/search?l=XML&q=formbuilder_whitelist&type=&utf8=%E2%9C%93), so when you install those addons you automatically get a website-form-enabled model ready to use out of the box.
- I use such method in demo data to mimic the behavior of such addons and have some data to run the tour (which is ❌ right now).
- The main purpose of `website_form` is to allow unauthenticated users to create records that they should have no access to, such as creating a `crm.lead`. The way to verify that the form came from a trusted source we have the CSRF token, provided out of the box. If we want to use the engine to provide a more granular authentication system, such as allowing only portal users to submit to certain models, then I think that belongs to another addon, independent of this one (it would extend the `website_form` engine, no matter if to use it with EE builder or this one).

website_form_builder/README.rst
yajo: [That happens automatically](https://github.com/odoo/odoo/blob/052323fae4b8f37428e844a80cb807c9eefbc47c/addons/website_form/models/models.py#L57), I was just trying to avoid having to fill the docs gap from `website_form`, but I think I'll have to do it after all 😆

website_form_builder/static/src/js/snippets.js
yajo: I know those common JS conventions, but in Odoo it seems they decided to match the python ones. Just give a glance at https://www.odoo.com/documentation/11.0/reference/javascript.html and you'll see. So, I prefer to stick with those.

website_form_builder/static/src/js/widgets.js
yajo: I'd say it's safe to use it nowadays. 🤔 See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas#Browser_compatibility

website_form_builder/README.rst
yajo: As I just mentioned above in https://github.com/OCA/website/pull/402#issuecomment-352096802, addons that depend on `website_form` already provide those, so not needed IMO.

yajo:
I think all your concerns are solved now, let me focus on the tour. However functional reviews can start already 😊

yajo:
Ok, it went ✔️ locally, let's see in Travis. The tour is not as exhaustive as I'd wish, but it's a good start IMHO.

tarteo (changes requested):
@Yajo Very useful module!

I followed the steps provided by the readme using model project and task but I got stuck when I tried to whitelist a field. The error I got was: "Properties of base fields cannot be altered in this manner! Please modify them through Python code, preferably through a custom addon!".

It also would be nice to be able to edit "Blacklisted in web forms" from the Model form. Now it's only available through the Field form (Settings -> Technical -> Database Structure -> Fields).

yajo:
AFAIK the setting is available in the model form too:

captura de pantalla de 2018-01-10 10-40-28

> I followed the steps provided by the readme using model project and task but I got stuck when I tried to whitelist a field. The error I got was: "Properties of base fields cannot be altered in this manner! Please modify them through Python code, preferably through a custom addon!".

Could you provide the steps please?

tarteo:
@Yajo 

- Activate developer mode
- Go to Technical -> Database Structure -> Models
- Check "Allowed to use in forms"
- Click a field in the Fields tab
- A modal form opens with the field form
- "Blacklisted in web forms" is not there
- Go to Technical -> Database Structure -> Fields
- Search for Project
- Click the name field
- "Blacklisted in web forms" is there
- Uncheck "Blacklisted in web forms"
- Click Save
- A message shows: "Properties of base fields cannot be altered in this manner! Please modify them through Python code, preferably through a custom addon!"

yajo:
It seems [you can't alter fields properties through UI unless it's a manually-created model](https://github.com/odoo/odoo/blob/2bedd640db14642441c568c2d0b36441f4ba554a/odoo/addons/base/ir/ir_model.py#L649-L652). I'm not sure I want to change that... Should I then just change the docs?

tarteo:
@Yajo haha, that would be funny! So the module will only work for manual created models through the UI. That's np, it's still a very useful module. Maybe you can add a disclaimer to the readme.

yajo:
Docs fixed

tarteo (approved):
:+1:

yajo:
It seems travis will randomly crash until https://github.com/odoo/odoo/pull/22330 is merged. Not this addon's fault in any case.

yajo:
Setting as WIP again to fix the hidden fields UI

yajo:
Done. Now, instead of having the previously ugly and unusable UI for hidden data, now the user can simply set the field as hidden and use it as usual. Since hidden fields are expected to have a value, when hiding a field without it, the user will be asked for it automatically.

Please @rafaelbn review

rafaelbn:
Hi @Yajo 👍 

Please try the form I've created in http://3315250-402-88cc0c.runbot1.odoo-community.org/#scrollTop=0

Form says "An error has occured, the form has not been sent."

```
-----------------------------27635325922190
Content-Disposition: form-data; name="name"

Mismo asunto para todos
-----------------------------27635325922190
Content-Disposition: form-data; name="campaign_id"

2
-----------------------------27635325922190
Content-Disposition: form-data; name="partner_name"

Bodeguita
-----------------------------27635325922190
Content-Disposition: form-data; name="email_from"

a@a
-----------------------------27635325922190
Content-Disposition: form-data; name="phone"

+34 123 456 7891
-----------------------------27635325922190
Content-Disposition: form-data; name="mobile"

+34666888444
-----------------------------27635325922190
Content-Disposition: form-data; name="street"

street
-----------------------------27635325922190
Content-Disposition: form-data; name="state_id"

28
-----------------------------27635325922190
Content-Disposition: form-data; name="zip"

28045
-----------------------------27635325922190
Content-Disposition: form-data; name="description"

Notas largas
-----------------------------27635325922190
Content-Disposition: form-data; name="tag_ids"

6,7
-----------------------------27635325922190
Content-Disposition: form-data; name="Custom Date field"

01/30/2018
-----------------------------27635325922190
Content-Disposition: form-data; name="Custom Short text field"

Corto
-----------------------------27635325922190
Content-Disposition: form-data; name="Custom Long text field"

Largo
-----------------------------27635325922190
Content-Disposition: form-data; name="Custom Yes or not field"

Custom Yes or not field
-----------------------------27635325922190
Content-Disposition: form-data; name="Custom File upload field[0]"; filename="2018-01-22_14-13-36.png"
Content-Type: image/png

PNG


```

Answer says: false

yajo:
I added the possibility to add multiple model fields at once, but I didn't have time yet to check that bug @rafaelbn

rafaelbn:
Thanks @Yajo ! after reviewing the BUG it is 👍 in my side!

yajo:
I can reproduce the bug if I try to add a file upload field when the model has no `mail.thread` features. I'm not sure if that's a bug or how to fix that, honestly, but I will add a note in known issues for now.

yajo:
Done now @rafaelbn

rafaelbn:
@Yajo stil fails in model crm.lead (for ex.) basic one 😄

yajo:
Could you please try in the empty database? It seems to be conflicting with some other addon, maybe one that validates input or something like that. Your very same form works for me locally.

rafaelbn:
That's right @Yajo  I tested in production in customer and it works. 👍 😄

rafaelbn (approved):
Great one!

yajo:
Big milestone today! 🚀 🎉 😊

PR #404, opened 2017-12-21
[11.0][MIG] website_analytics_piwik

astirpe:
Porting of module website_analytics_piwik to V11


simahawk (approved):
just code review: LGTM

simahawk:
I'd squash translation commits all in one and - if time allows - prefix some commits (like "fix coding style" and such) w/ the module name :wink:

website_analytics_piwik/views/templates.xml
```diff
@@ -1,9 +1,9 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <odoo>
 
     <template id="layout" inherit_id="website.layout">
         <xpath expr="//script[@id='tracking_code']" position="after">
-            <t t-if="website and website.piwik_analytics_id and website.piwik_analytics_host">
+            <t t-if="website and website.has_piwik_analytics and website.piwik_analytics_host">
```
hmrodrigues: Maybe also check if the piwik_analytics_id is set, since I as a user can check the bool but not specify the tracking ID. Or just set the piwik_analytics_id (and host) as required when has_piwik_anlytics is true on the config view.

website_analytics_piwik/models/website.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
hmrodrigues: Why have you removed the encoding?

astirpe: Because utf-8 is default encoding for Python 3

hmrodrigues (approved):

astirpe:
Translation commits squashed together

astirpe:
Can we merge this one?

yung-wang (approved):
Tested on our piwik server with odoo v11 and it works. Please merge this PR.

hbrunn (changes requested):
could you add a migration script that sets the flag for existing installations which have an ID set?

astirpe:
@hbrunn Done!

astirpe:
Squashed

hbrunn (approved):

pedrobaeza:
Merging

hbrunn:
so I kept this tab open so wait for runbot come back positive...

PR #354, opened 2017-06-15
[ADD][website_multi_theme] Allow different themes by website

yajo:
This a somewhat hacky addon that adds support for setting a different theme by website.

It includes a workaround for https://github.com/odoo/odoo/pull/17635 and https://github.com/odoo/odoo/pull/17970.

It bundles lazy support for theme_bootswatch and a bunch of popular private themes found in the app store.

@Tecnativa

PR #410, opened 2018-01-31
[FIX] skip handling non-theme or non extenstion templates

yelizariev:
Small update that can be considered independently from #409.

Technical details are in the code comments. Simple example of the error is theme with snippets. Once you set multi-theme for several website, you get error after initiating Website Editor  
```

  File "/mnt/odoo-source/odoo/addons/base/ir/ir_qweb/qweb.py", line 682, in _compile_directives
    return compile_handler(el, options)
  File "/mnt/odoo-source/addons/web_editor/models/ir_qweb.py", line 49, in _compile_directive_snippet
    name = self.env['ir.ui.view'].search([('key', '=', el.attrib.get('t-call'))]).display_name
  File "/mnt/odoo-source/odoo/fields.py", line 942, in __get__
    record.ensure_one()
  File "/mnt/odoo-source/odoo/models.py", line 4265, in ensure_one
    raise ValueError("Expected singleton: %s" % self)
ValueError: Expected singleton: ir.ui.view(1339, 1543)
```


yelizariev:
@Yajo do you have objections against skipping views without ``inherirt_id`` value?

yajo:
🤔 I feel this should be an error on the Qweb engine itself...

If I'm not wrong, no matter if the view has or not a parent, the problem is that Qweb is not narrowing the search to views enabled only for this website. I remember that when I published #354, I had to patch Odoo in https://github.com/odoo/odoo/pull/17635. Most likely now the problem is that [the search QWeb is doing again doesn't exclude views from other websites](https://github.com/odoo/odoo/blob/d295b4360db93a0d3333153098e623e573ac099b/addons/web_editor/models/ir_qweb.py#L47) (the link is from v10, so I guess you should patch it there so Odoo forwards the patch to v11).

If you agree, please close this PR and open one to Odoo and ping me there. Thanks!

yelizariev:
It seems unlikely to have a necessity to have two different views with the same name (``key``), but with different ``website_id`` value.

Also, this update in ``website_multi_theme`` should speed it up a little bit (though it's not so important, because it affects configuration stage only)

yajo:
> It seems unlikely to have a necessity to have two different views with the same name (key), but with different website_id value.

On the contrary, that's exactly the most basic feature for `website_multi_theme` to work!

The assumption that a standalone view doesn't need to be website-enabled could be good in most cases, but I'm not sure it will work in all of them... I strongly recommend the bug fix in Odoo/OCB. 😅

yelizariev:
I made a patch: https://github.com/yelizariev/odoo/commit/52de9ae545506c8601e727c0cc80b52b4cce3707

But I still don't understand why someone needs different copies of snippets for different websites. The websites can share the same single copy of snippet.

The same about views that don't inherit anything.

For those reasons, I don't want to propose it to odoo, because I cannot strongly explain why it's needed

yajo:
At least I'd try to PR them.

It's maybe not needed for their product, but it affects addon developed on their framework. It's indeed a framework failure, and that's the place where it should be fixed.

The reasoning is simple: depending on the website, it could be resolving to one or another result, and that should be honored.

yelizariev:
Ok, I think I know when it would be needed:

> Allow duplicate snippets to have different versions for different websites

The PR: https://github.com/odoo/odoo/pull/24429

yelizariev:
There is still the question what to do with ``website_multi_theme`` because it's unlikely that odoo will merge the updates to stable branch (it's not a fix, but rather improvement or even new feature)

pedrobaeza:
You should patch here the function (I think you can do it via a context + overriding search on the model)

yelizariev:
ok, that could be a solution

yelizariev:
Implemented here https://github.com/OCA/website/pull/453

PR #420, opened 2018-02-07
[11.0][MIG] website snippet anchor

tarteo:


website_snippet_anchor/README.rst
```diff
@@ -0,0 +1,82 @@
...
+#. Edit any web page.
+#. Select some text.
+#. Press *Link* in the editor toolbar.
+#. Choose a page in the opened dialog.
+#. Set an anchor there too (without ``#`` symbol).
+#. Press *Save*.
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/186/10.0
```
simahawk: /11.0

website_snippet_anchor/__manifest__.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+# Copyright 2015, 2017 Jairo Llopis <jairo.llopis@tecnativa.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+{
+    "name": "Set Snippet's Anchor",
+    "summary": "Allow to reach a concrete section in the page",
+    "version": "11.0.1.0.0",
+    "category": "Website",
+    "website": "https://www.tecnativa.com",
```
simahawk: https://github.com/OCA/website

simahawk (changes requested):
just small issues. Code LGTM.

tarteo:
@simahawk Thanks, I've applied your comments.

rgarnau:
LGTM :+1:

website_snippet_anchor/static/src/js/link_dialog.js
```diff
@@ -50,6 +46,7 @@ odoo.define("website_snippet_anchor.link_dialog", function (require) {
             var url = this.data.url || "",
                 url_parts = url.split("#", 2);
             if (url_parts.length > 1) {
+                console.log(url_parts);
```
yajo: :fire:

website_snippet_anchor/templates/snippets.xml
```diff
@@ -4,13 +4,11 @@
...
-<template id="snippet_options" inherit_id="web_editor.snippet_options">
-    <xpath expr=".">
+<template id="snippet_options" inherit_id="website.snippets">
+    <xpath expr="//div[@id='snippet_options']" position="inside">
         <div data-js="anchor"
              data-selector="section, .carousel">
-            <li data-anchor_ask=""
-                data-only="click"
-                title="Use anchors to go directly to a section of your page, appending #anchor to the URL">
+            <li title="Use anchors to go directly to a section of your page, appending #anchor to the URL">
```
yajo: I prefer the previous way, since it just follows the standard for snippet options. Please restore it if possible.

yajo (changes requested):
Functionally tested and it works.
Some migration remarks though.

website_snippet_anchor/static/src/js/anchor_option.js
```diff
@@ -5,20 +5,29 @@
...
-            if (type != "click") {
-                return $.Deferred().reject();
+
+        start: function() {
+            var res = this._super.apply(this, arguments);
+            return res;
+        },
+
+        anchor_ask: function(type) {
+            if(type !== false) {
```
Yajo: If I'm not wrong, this should be `if (type !== "click")`, because the 1st argument that gets in is always the type of event, which could be `click`, `reset` or `hover`. We want to react only on click.

website_snippet_anchor/static/src/js/anchor_option.js
```diff
@@ -5,20 +5,29 @@
...
+                return;
```
Yajo: Since the other option is to return a `Deferred`, the normal thing here would be to `return $.Deferred().reject();`

website_snippet_anchor/templates/snippets.xml
```diff
@@ -4,12 +4,11 @@
...
 <odoo>
 
-<template id="snippet_options" inherit_id="web_editor.snippet_options">
-    <xpath expr=".">
+<template id="snippet_options" inherit_id="website.snippets">
+    <xpath expr="//div[@id='snippet_options']" position="inside">
         <div data-js="anchor"
              data-selector="section, .carousel">
             <li data-anchor_ask=""
-                data-only="click"
```
Yajo: You should leave this line if you want to avoid `hover` events. However, there are still the `reset` ones... I think this is a bug in Odoo, not sure if still happening in v11 TBH.

website_snippet_anchor/templates/snippets.xml
```diff
@@ -0,0 +1,22 @@
...
+<!-- Copyright 2015, 2017 Jairo Llopis <jairo.llopis@tecnativa.com>
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html). -->
+
+<odoo>
+
+<template id="snippet_options" inherit_id="website.snippets">
+    <xpath expr="//div[@id='snippet_options']" position="inside">
+        <div data-js="anchor"
+             data-selector="section, .carousel">
+            <li data-anchor_ask=""
```
yajo, line 11: There was a `data-only="click"` here before. Leave it as it was if you want to avoid hover events.

website_snippet_anchor/static/src/js/anchor_option.js
```diff
@@ -0,0 +1,80 @@
...
+        utils = require("website.utils"),
+        sprintf = _.str.sprintf,
+        _t = core._t;
+
+    // Option to have anchors in snippets
+    options.registry.anchor = options.Class.extend({
+
+        anchor_ask: function(type) {
+            if(type !== false) {
+                return;
```
yajo: Since the good result is a deferred, it makes more sense to `return $.Deferred().reject()` when failing. Leave as it was please.

website_snippet_anchor/static/src/js/anchor_option.js
```diff
@@ -0,0 +1,80 @@
...
+        options = require("web_editor.snippets.options"),
+        utils = require("website.utils"),
+        sprintf = _.str.sprintf,
+        _t = core._t;
+
+    // Option to have anchors in snippets
+    options.registry.anchor = options.Class.extend({
+
+        anchor_ask: function(type) {
+            if(type !== false) {
```
yajo: type is the event type. You want to react only in `click` event, so leave it as it was: `if (type !== "event")`

yajo (changes requested):
I think you're changing things because of taste before understanding how the snippet options api works... :thinking:

website_snippet_anchor/templates/snippets.xml
tarteo: No, this attribute is not implemented anymore. I'll  use the new data-no-preview="true" instead.

yajo: Oh great, I didn't know that :wink:

website_snippet_anchor/static/src/js/anchor_option.js
tarteo: No, when on click type is in 11.0 false. I'll remove it when I apply the attribute data-no-preview

tarteo:
@Yajo I know how the options for snippets work... Do not try to assume things next time.

yajo:
Yikes, then it was me that didn't know the api changes on v11! Thanks, didn't want to bother you :innocent:

yajo (approved):
Perfect. Please squash migration commits to merge :blush:

simahawk (approved):

yajo (approved):

Issue #415, opened 2018-02-02
Travis failure in 11.0 due to website_legal_page


pedrobaeza:
@Yajo @nadiaafa this module is provoking now a failing in Travis due to changes in upstream. Can you check?

simahawk:
@pedrobaeza tests are green, nevertheless this hack should be removed https://github.com/OCA/website/blob/11.0/website_legal_page/tests/test_controller.py#L14 as the bug has been fixed by odoo. /cc @nadiaafa

pedrobaeza:
Tests are not green. You can see it here: https://travis-ci.org/OCA/website

PR #422, opened 2018-02-09
[11.0][ADD] website_adv_image_optimization

tarteo:
This module allows the user to manually apply compression and resize options on an image.


website_adv_image_optimization/controllers/main.py
```diff
@@ -0,0 +1,55 @@
...
+
+class MainController(http.Controller):
+    @http.route('/website_adv_image_optimization/attachment/optimize',
+                type='http', auth='user', methods=['POST'])
+    def optimize(self, attachment_id, quality=80,
+                 width=False, height=False, **kwargs):
+        attachment_obj = request.env['ir.attachment']
+        attachment = attachment_obj.search([('id', '=', attachment_id)])
+
+        image = Image.open(io.BytesIO(base64.b64decode(attachment.datas)))
```
simahawk: can't you use some tools from here https://github.com/odoo/odoo/blob/11.0/odoo/tools/image.py ?

website_adv_image_optimization/controllers/main.py
```diff
@@ -0,0 +1,55 @@
...
+        })
+        return json.dumps(True)
+
+    @http.route('/website_adv_image_optimization/attachment/dimensions',
+                type='http', auth='user', methods=['POST'])
+    def dimensions(self, attachment_id, **kwargs):
+        attachment_obj = request.env['ir.attachment']
+        attachment = attachment_obj.search([('id', '=', attachment_id)])
+
+        image = Image.open(io.BytesIO(base64.b64decode(attachment.datas)))
```
simahawk: As far as I see you use this only to get the image size to be used in `set_defaults`: can't you simply get the size via JS?

In any case you are leaving opened file objects around ;)

simahawk (changes requested):
some questions.
Then would be nice to include some screenshots in README.

website_adv_image_optimization/controllers/main.py
tarteo: I can try

website_adv_image_optimization/controllers/main.py
tarteo: Not really, because those do 1 operation at the time it will be less efficient. This way i can reuse some variables.

simahawk: ok, then 3 recommendations: 
1. add a comment stating it (for me, for your future self and the ones that will ask the same question ;))
2. move resizing to a separated method 
3. close IO objects after getting the value

rgarnau (approved):
Tested on runbot :+1:

tarteo:
The travis error is related to https://github.com/OCA/website/issues/415

tarteo:
@simahawk Is this what you meant?

- I added a comment
- Moved resizing and changing jpeg quality to separated methods
- Closed Image object after all values are extracted

website_adv_image_optimization/static/src/js/website_adv_image_optimization.js
```diff
@@ -0,0 +1,107 @@
...
+            var done = $.Deferred();
+            var img = new Image();
+            img.onload = function() {
+                done.resolve([this.width, this.height]);
+            }
+            img.src = this.attachment_src;
+            return done.promise();
+        },
+        apply: function() {
+        console.log(this);
```
simahawk: debug bit lost? :)

website_adv_image_optimization/tests/test_controller.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class FakeRequest:
+    def __init__(self, env):
+        self.env = env
+        self.endpoint = FakeEndPoint()
+
+
+class TestController(TransactionCase):
+    def test_optimize(self):
```
simahawk: In the controller file you should import request:
```python
from odoo.http import request
```
then:
```python
import mock
[...]
    @mock.patch('dotted.path.to.your.controller.request')
    def test_optimize(self, mocked_request):
        mocked_request.env = self.env
        attachment = self.env.ref('website.business_conference')
        size = attachment.file_size
        ctrl = MainController()
        ctrl.optimize(attachment.id, 70, 300, 300)
        attachment = self.env.ref('website.business_conference')
        new_size = attachment.file_size
        self.assertFalse(new_size == size)
```

simahawk (changes requested):
looks good. I'd change the way you patched request in tests to something more common and reliable :wink:

tarteo:
@simahawk Ah yes! Thanks!

Fuchur777:
@simahawk changes done to your liking? Can it pass your review standards?

simahawk (approved):

simahawk:
let's remember to squash commits before merge :wink:

tarteo:
@simahawk I squashed the commits :)

PR #423, opened 2018-02-09
[11.0][ADD] website_media_size

tarteo:
This module shows the size of the media in the media selector (only for media that was uploaded).

screenshot


website_media_size/static/src/js/website_media_size.js
```diff
@@ -0,0 +1,42 @@
...
+
+    MediaDialog.include({
+        xmlDependencies: MediaDialog.prototype.xmlDependencies.concat(
+            ['/website_media_size/static/src/xml/website_media_size.xml']
+        ),
+        start: function() {
+            var res = this._super.apply(this, arguments);
+            this.include_file_size(this.imageDialog);
+            this.include_file_size(this.documentDialog);
+             // FIXME
```
simahawk: ?

simahawk (changes requested):

website_media_size/static/src/js/website_media_size.js
tarteo: The this.search(); is a weird fix to update the size values. I couldn't find a proper solution. There must be a better way to fix it.
In the start of super it already executes .search(). I can probably just remove the FIXME because I don't think it's actually an issue.

rgarnau (approved):
LGTM :+1:

website_media_size/static/src/css/website_media_size.css
```diff
@@ -0,0 +1,12 @@
+/* Copyright 2018 Onestein
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). */
+
+.o_existing_attachment_cell .o_existing_attachment_size {
+    position: absolute;
+    padding: 3px;
+    background-color: rgba(0, 0, 0, 0.8);
```
yajo, line 7: You can use "opacity"

website_media_size/static/src/css/website_media_size.css
```diff
@@ -0,0 +1,12 @@
...
+    color: #FFF;
+    border-radius: 0 0 0 2px;
+    cursor: pointer;
+    font-size: 80%;
```
yajo, line 11: It's way better if you use a less file and use Bootstrap variables for such properties, or even better to use a normal bootstrap class such as: https://getbootstrap.com/docs/3.3/components/#labels

website_media_size/static/src/js/website_media_size.js
```diff
@@ -0,0 +1,41 @@
+/* Copyright 2018 Onestein
+ * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). */
+
+odoo.define('website_media_size', function(require) {
+    var MediaDialog = require('web_editor.widget').MediaDialog;
```
Yajo: "use strict"

website_media_size/static/src/js/website_media_size.js
```diff
@@ -0,0 +1,41 @@
...
+    var weContext = require("web_editor.context");
+
+    MediaDialog.include({
+        xmlDependencies: MediaDialog.prototype.xmlDependencies.concat(
+            ['/website_media_size/static/src/xml/website_media_size.xml']
+        ),
+        start: function() {
+            var res = this._super.apply(this, arguments);
+            this.include_file_size(this.imageDialog);
+            this.include_file_size(this.documentDialog);
```
yajo: Wouldn't it be better (less monkey-y 🙈) to patch each of these 2 classes' `_rpc` method and add the required field to `arguments[0].kwargs.fields` if the model is `ir.attachment`?

yajo (changes requested):

website_media_size/static/src/js/website_media_size.js
tarteo: That's a great idea! Why I didn't taught of that. That's genius. :smile: I'll apply it asap.

website_media_size/static/src/css/website_media_size.css
tarteo: I need the background to be black.

tarteo:
@Yajo I applied your suggestion on Widget class because ImageDialog is not accessible. I do a check on template so it is fine.
I also replaced the screenshot because of the possibly copyrighted picture of the monkey. I replaced it with a non copyrighted picture of a monkey.

website_media_size/static/src/css/website_media_size.css
tarteo: Yes, but I would be changing the label's style to match the other buttons.

website_media_size/static/src/js/website_media_size.js
```diff
@@ -2,40 +2,32 @@
  * License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). */
 
 odoo.define('website_media_size', function(require) {
+"use strict";
```
yajo: Indent this line please :innocent:

website_media_size/static/src/js/website_media_size.js
```diff
@@ -2,40 +2,32 @@
...
-                        fields: ['name', 'mimetype', 'checksum', 'url', 'type', 'file_size'],
-                        order: [{name: 'id', asc: false}],
-                        context: weContext.get(),
-                    }
-                }).then(obj.proxy('fetched_existing'));
-            };
+        }
+    });
+
+    // ImageDialog of the web_editor.widget module is not available
```
yajo: I took the freedom to open https://github.com/odoo/odoo/pull/23180, I'm pretty sure our friend @ged-odoo will merge that soon. :wink:

website_media_size/static/src/css/website_media_size.css
yajo: ... Only if the foreground is white :stuck_out_tongue_winking_eye:

website_media_size/static/src/css/website_media_size.css
yajo: Well, the point is that if you leverage Bootstrap, you make your thing look good no matter the theme; if you don't, i.e. a dark theme would render this ugly.

Non-blocking in any case IMHO

website_media_size/static/src/js/website_media_size.js
tarteo: Thanks!!

website_media_size/static/src/css/website_media_size.css
tarteo: Yes, I understand it.

website_media_size/static/src/css/website_media_size.css
tarteo: A black background will create contrast so it will be easy to read. Opacity will also apply it on the text making it less readable in all cases.

Fuchur777:
@simahawk @Yajo does it now LGTY? Can you amend your review?

website_media_size/static/src/js/website_media_size.js
```diff
@@ -0,0 +1,33 @@
...
+            ['/website_media_size/static/src/xml/website_media_size.xml']
+        ),
+        start: function() {
+            var res = this._super.apply(this, arguments);
+            this.search();
+            return res;
+        }
+    });
+
+    // ImageDialog of the web_editor.widget module is not available
```
simahawk: @tarteo https://github.com/odoo/odoo/pull/23180 got merged. Do you still need this?

tarteo: No, had no time to change it yet.

simahawk (approved):

astirpe (approved):

website_media_size/static/src/js/website_media_size.js
```diff
@@ -0,0 +1,31 @@
...
+    var MediaDialog = web_editor_widget.MediaDialog;
+    var ImageDialog = web_editor_widget.ImageDialog;
+
+    MediaDialog.include({
+        xmlDependencies: MediaDialog.prototype.xmlDependencies.concat(
+            ['/website_media_size/static/src/xml/website_media_size.xml']
+        ),
+        start: function() {
+            var res = this._super.apply(this, arguments);
+            this.search();
```
yajo: Hmm why this?

Also why patching different methods on different classes? I'm not saying it's wrong, just weird at 1st sight... :thinking:

yajo (changes requested):

website_media_size/static/src/js/website_media_size.js
tarteo: You're right this is not needed anymore. I removed it and moved the xml dependency to ImageDialog.

yajo (approved):
Yikes, no runbot to test, but code looks OK.

PR #427, opened 2018-02-21
[10.0][FIX] website_form_recaptcha: reset the recaptcha when the server send a error

RobinetDenisAcsone:
When the form is sent, and a error occur at the creation of the object (by constraint for example) the recaptcha isn't reset but the recaptcha key is already checked by google.
The second time the form is sent, the key isn't changed, and google will reject it.

This pr reset the recaptcha to obtain a new key from google when the user recheck it.

### DIFF
--- a/website_form_recaptcha/__manifest__.py
+++ b/website_form_recaptcha/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Website Form - ReCaptcha",
     "summary": 'Provides a ReCaptcha field for Website Forms',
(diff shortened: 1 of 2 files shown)


yajo (approved):

pedrobaeza:
Please update module version number and send the CLA

RobinetDenisAcsone:
Version updated, and the CLA should be good since yesterday evening.

pedrobaeza (approved):

andreparames:
LGTM :+1:

PR #430, opened 2018-03-07
[FIX] point inherit to copied views

yelizariev:
Otherwise, theme_customize doesn't work when we have a chain of views like in
``theme_clean``:

`theme_clean.option_bg_shade_light2` inherits `theme_clean.less`
`theme_clean.less` inherits `website.assets_frontend`

without this update, `website_multi_theme` copies
`theme_clean.option_bg_shade_light2` but keeps `inherit_id` the same (i.e. to
`theme_clean.less`, which is not used, because it has a copy)

### DIFF
--- a/website_multi_theme/models/website.py
+++ b/website_multi_theme/models/website.py
@@ -48,6 +48,10 @@ class Website(models.Model):
             self._multi_theme_activate()
         return result
 
+    def _find_duplicate_view_for_website(self, origin_view, website):
+        xmlid = VIEW_KEY % (website.id, origin_view.id)
+        return self.env.ref(xmlid, raise_if_not_found=False)
+
     def _duplicate_view_for_website(self, pattern, xmlid, override_key):
         """Duplicate a view pattern and enable it only for current website.
 
@@ -169,6 +173,25 @@ class Website(models.Model):
                         copied_view.inherit_id = custom_layout
                         data.attrib["inherit_id"] = custom_layout.key
                     copied_view.arch = etree.tostring(data)
+                else:
+                    theme_module_name \
+                        = website.multi_theme_id.converted_theme_addon
+
+                    parent_view = copied_view.inherit_id
+                    parent_view_module = parent_view.model_data_id.module
+                    copied_parent = None
+
+                    if parent_view_module == theme_module_name:
+                        # it inherits view of the same module,
+                        # which might be copied
+                        copied_parent = self._find_duplicate_view_for_website(
+                            parent_view, website
+                        )
+
+                    if copied_parent:
+                        copied_view.inherit_id = copied_parent
+                        data.attrib["inherit_id"] = copied_parent.key
+
                 custom_views |= copied_view
             # Delete any custom view that should exist no more
             (website.multi_theme_view_ids - custom_views).unlink()
--- a/website_multi_theme/__manifest__.py
+++ b/website_multi_theme/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Website Multi Theme",
     "summary": "Support different theme per website",
-    "version": "10.0.1.0.1",
+    "version": "10.0.1.0.2",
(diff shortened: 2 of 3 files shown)


yelizariev:
@Yajo :smiley:

ilmir-k (approved):
:+1:

yajo (approved):
Good catch!

yajo:
Travis has :x:

yelizariev:
it's fixed and squashed
https://github.com/yelizariev/website/commit/2f87813674e42748dee2ab970f4bcc2907865f98

Also, I've added separate commit to fix jslints

yelizariev:
@pedrobaeza, could you take a look?

website_multi_theme/models/website.py
```diff
@@ -170,6 +175,21 @@ def _multi_theme_activate(self):
                         copied_view.inherit_id = custom_layout
                         data.attrib["inherit_id"] = custom_layout.key
                     copied_view.arch = etree.tostring(data)
+                elif (copied_view.inherit_id and
```
pedrobaeza: This is very hard to read. Can you somehow split this to be more readable (an `else`, then prepare variables and make the comparison)?

website_multi_theme/models/website.py
```diff
@@ -48,6 +48,10 @@ def write(self, vals):
             self._multi_theme_activate()
         return result
 
+    def _find_duplicate_view_for_website(self, origin_view, website):
+        xmlid = VIEW_KEY % (website.id, origin_view.id)
+        return self.env.ref(xmlid, raise_if_not_found=False)
```
Yajo: Are you sure of this? Shouldn't you search the view by the combination of `key` and `website_id` in `ir.ui.view`?

I think this needs tests :thinking:

website_multi_theme/models/website.py
yelizariev: This should be better https://github.com/yelizariev/website/commit/52eff0a3d660b8776712582ce054036cd500215e

website_multi_theme/models/website.py
```diff
@@ -120,6 +124,10 @@ def _multi_theme_activate(self):
         # Patterns that will be duplicated to enable multi themes
         assets_pattern = self.env.ref("website_multi_theme.assets_pattern")
         layout_pattern = self.env.ref("website_multi_theme.layout_pattern")
+
+        def view2module(view):
+            return view and view.model_data_id.module
```
yajo: I'm not sure if you know that Odoo lets you browse empty recordsets, returning a falsey value for all fields.

Thus, you can use a variable instead of a method if you want, just below line 141 you can write:

```python
copied_view_module = copied_view.model_data_id.module
```

It will be `False` if `len(copied_view)==0`.

pedrobaeza:
Please rebase after merging the other patch

website_multi_theme/models/website.py
yelizariev: Yes, I know. For some reason, I thought that it could be False in the middle of the chain, so I decided just to make the extra check. But it's not needed here. You are right

yelizariev:
Rebased.
Also, I made refactoring and changed only last number in the version (to 10.0.1.0.2) as this update doesn't require module updating

ilmir-k:
@pedrobaeza @Yajo could you review the updates please. Thanks

yajo (approved):
Code looks good. It would be nice to have some testing on it.

pedrobaeza:
Can you add those tests?

yelizariev:
Not soon...
Also, the code uses ``website.multi_theme_id.converted_theme_addon`` which is empty for demo data. So I'm not sure how to make CI tests for it

yajo:
I prefer to have the fix than not having tests for it. It can be a good future improvement, but I'm not blocking just for that. Thanks @yelizariev!

PR #433, opened 2018-03-09
[FIX] website_multi_theme: Proper checking of special views inheritance

yajo:
While reviewing #430 I saw this bug, so here's the quick fix.

Some of these blocks were never hit, so here I'm fixing the false negatives by using recordset operations.

@Tecnativa


pedrobaeza (approved):
Fast-tracking this

yelizariev:
@Yajo Shall it be applied for 11.0?

yelizariev:
https://github.com/OCA/website/pull/435

PR #436, opened 2018-03-16
[FIX] "Customize Theme" for theme_anelusia didn't' work

yelizariev:
original method theme_customize_get returns view.xml_id, which are
'website_multi_theme.auto_view_ID_WEBSITE', while client works with original
IDs.

It leads to problem with, for example, theme_anelusia, because it has more than
one setting in its wizard. Without this commit it saves only one setting and
disable all other

### DIFF
--- a/website_multi_theme/__manifest__.py
+++ b/website_multi_theme/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "Website Multi Theme",
     "summary": "Support different theme per website",
-    "version": "11.0.1.0.0",
+    "version": "11.0.1.0.1",
     "category": "Website",
     "website": "https://www.tecnativa.com",
     "author": "Tecnativa, Odoo Community Association (OCA)",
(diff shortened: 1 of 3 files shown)


yelizariev:
@Yajo one more  fix :wink:

Railia (approved):
I installed this updates, created a database without demo data. Then I installed Theme Anelusia eCommerce with dependencies.  Now everything works as it should.

pedrobaeza (changes requested):
The override doesn't call super, breaking inheritability, so you have to find a way to keep both things (your change and the inheritability). Change also version number

yelizariev:
Updated: https://github.com/yelizariev/website/commit/ad6ffac09381c2b90579ac9a07431c4a79c1fdc0

yelizariev:
I thought sacrifice inheritability for the following reason:

* unlikely that someone needs to inherit this specific method
* it doubles the number of browsing \ searching

Both arguments are weak, though

yajo (approved):
Good catch!

Railia:
Seems it works

Railia (approved):
Seems it works

pedrobaeza (approved):

PR #441, opened 2018-03-23
[10.0] website_multi_theme: Backport for theme_customize_get fix

yelizariev:

PR #442, opened 2018-03-23
10.0 website multi theme fix data is not defined

yelizariev:
Depends on #441 

Fixes mistake in #430 

So, it's just another evidence that tests are useful thing, but I still have a [doubt](https://github.com/OCA/website/pull/430#issuecomment-372824306) how to do it in this case.

### DIFF
--- a/website_multi_theme/__manifest__.py
+++ b/website_multi_theme/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Website Multi Theme",
     "summary": "Support different theme per website",
-    "version": "10.0.1.0.3",
+    "version": "10.0.1.0.4",
     "category": "Website",
     "website": "https://www.tecnativa.com",
     "author": "Tecnativa, Odoo Community Association (OCA)",
--- a/website_multi_theme/models/website.py
+++ b/website_multi_theme/models/website.py
@@ -164,15 +164,12 @@ class Website(models.Model):
                     False
                 )
                 # Applied views must inherit from custom assets or layout
+                new_parent = None
                 if copied_view.inherit_id & main_views:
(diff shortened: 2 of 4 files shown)


Railia (changes requested):
When I set up the theme "Clean" for the 2nd website, using the "Reload theme" button, I get an error message

Railia (approved):
The previous check was incorrect. It was a problem with the database. After reinstalling the database, the module is working fine!

yajo (approved):
Keep in mind that, although the design of this addon is mostly OK, it's got some design flaws and is very *hacky*. It abuses many low-level things inside Odoo to make multi themes possible. With such properties, if you don't test these bugs, it's very likely that we'll have them in future versions...

I won't lock it for not having tests, but friend please think on that. :wink: 

About how to do them, basically you just have to do like always:

1. Create some data that hits the corner case.
2. Check that the method works.

- In some cases you might need to use an `HttpCase`
- In such cases you'll need to apply https://github.com/odoo/odoo/issues/12237#issuecomment-254781169 to acquire/release the same cursor that the http connections will use
- You might also need to use [mock](https://docs.python.org/3/library/unittest.mock.html) under some circumstances.

Oh, and rebase please!

pedrobaeza:
Please solve conflicts. Are you going to try the tests?

yelizariev:
I'll try to do some test. Probably just put fake data in ``converted_theme_addon`` field or try to find another solution

yelizariev:
There is no coverage tool anymore?

I added some test. They are only for coverage without real checking

yelizariev:
Ok, codecov is working and shows that the lines are covered

pedrobaeza:
OK, thank you very much.

PR #443, opened 2018-04-11
[10.0][IMP][website_multi_theme] make customize_show views part of multi-theme

yelizariev:
The PR allows making customization independently per website
multi-customization


website_multi_theme/data/themes_default.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2018 Ivan Yelizariev <https://it-projects.info/team/yelizariev>
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html). -->
+<odoo>
+    <!-- Default theme. It's basically a workaround for "Multi-customization"
+         feature which requires theme be set -->
```
yelizariev, line 6: I don't see other solution. At least until we make refactoring discussed #409

website_multi_theme/models/ir_ui_view.py
```diff
@@ -22,3 +24,29 @@ class IrUiView(models.Model):
...
              "the single website theme that owns it to multi website mode.",
     )
+
+    # TODO: add fields below to form view
+    origin_view_id = fields.Many2one(
+        "ir.ui.view",
+        string="Copied from",
+        help="View from where this one was copied for multi-website"
+    )
+    copy_ids = fields.One2many(
```
yelizariev: Those two fields are left after initial draft of this update, but I think it could useful later

website_multi_theme/models/website.py
```diff
@@ -82,18 +85,15 @@ def _duplicate_view_for_website(self, pattern, xmlid, override_key):
...
             "arch_fs": False,
-            "customize_show": False,
             "key": key,
             "multi_theme_generated": True,
-            "name": u"{} (Multi-website {} for {})".format(
-                pattern.name,
-                xmlid,
-                self.display_name,
-            ),
+            "name": '%s (Multi-Website)' % pattern.name,
```
yelizariev: It's renamed, because this name is used in Customize menu (grayed text)

website_multi_theme/models/website.py
```diff
@@ -82,18 +85,15 @@ def _duplicate_view_for_website(self, pattern, xmlid, override_key):
...
             "website_id": self.id,
+            "origin_view_id": pattern.id,
```
yelizariev: It's kind of replacement to removed data in name

website_multi_theme/models/website.py
```diff
@@ -178,12 +178,12 @@ def _multi_theme_activate(self):
...
-                        # which might be copied
-                        copied_parent = self._find_duplicate_view_for_website(
-                            parent_view, website
-                        )
+                    # check if parent was copied, so we need inherit that
+                    # instead of original parent, which is deactivated and not
+                    # used
+                    copied_parent = self._find_duplicate_view_for_website(
+                        parent_view, website
+                    )
```
yelizariev, line 188: It's not directly part of this update, though there is a theoretical situation when several Customized views inherit each other, so they need to use a copied view too. But I need to make real multi-footer in my module ``website_multi_company`` to apply some extension to make ``website.footer_custom`` copyable too

website_multi_theme/models/website_theme.py
```diff
@@ -36,20 +36,48 @@ class WebsiteTheme(models.Model):
...
+            if one.converted_theme_addon:
+                # Get all views owned by the converted theme addon
+                refs += self.env["ir.model.data"].search([
+                    ("module", "=", one.converted_theme_addon),
+                    ("model", "=", "ir.ui.view"),
+                ])
+
+            if refs or one.asset_ids:
+                # add common_refs only for installed themes
+                refs += common_refs
```
yelizariev: This is the key idea of the update

website_multi_theme/templates/patterns.xml
```diff
@@ -6,7 +6,7 @@
     <template
         id="layout_pattern"
         inherit_id="website.layout"
-        name="Template for replacing assets_frontend"
+        name="Main layout"
```
yelizariev, line 9: it's renamed because it's used in Customize menu

yelizariev (commented):
Some explanation of the updates

website_multi_theme/models/website.py
```diff
@@ -82,18 +85,15 @@ def _duplicate_view_for_website(self, pattern, xmlid, override_key):
             return result
         # Copy patterns only for current website
         key = xmlid if override_key else pattern.key
+
         result = pattern.copy({
             "active": pattern.was_active,
             "arch_fs": False,
-            "customize_show": False,
```
yelizariev: @Yajo is there any reason why this was set to False? Do you remember the case when it's required?

website_multi_theme/README.rst
```diff
@@ -117,7 +117,7 @@ Known issues / Roadmap
 * Theme picker should include some kind of thumbnail if possible.
 * If you install any of the supported themes after installing this addon, you
   will have to press *Reload* in the website config wizard to make it notice
-  the change.
+  the change. The same is true when you install any module with ``customize_show="True"``
```
yelizariev, line 120: Tests are not passed in travis due to this new requirement: *Reload* must be done after installing  website_blog_category https://github.com/OCA/website/blob/10.0/website_blog_category/templates/website_blog_template.xml#L25

What can we do here? @Yajo @pedrobaeza

website_multi_theme/models/website.py
```diff
@@ -85,15 +88,10 @@ def _duplicate_view_for_website(self, pattern, xmlid, override_key):
...
```
yelizariev: *Copied message to don't miss it:*

@Yajo is there any reason why this was set to False? Do you remember the case when it's required?

website_multi_theme/models/website.py
yelizariev: Maybe, the name can be changed to ``'%s (Website #%s)' % (pattern.name,  self.id)`` , because **Multi-website** sounds like **the same per different websites**

yelizariev: I don't want to use ``self.display_name``, because website could be renamed

yelizariev:
When you review PR yourself

dog-walking-itself-1

website_multi_theme/models/website.py
yajo: Can't remember, honestly :roll_eyes:

website_multi_theme/models/website.py
```diff
@@ -171,19 +169,15 @@ def _multi_theme_activate(self):
                     elif copied_view.inherit_id & main_layout:
                         new_parent = custom_layout
                 else:
-                    theme_module_name \
-                        = website.multi_theme_id.converted_theme_addon
-
                     parent_view = copied_view.inherit_id
-                    parent_view_module = parent_view.model_data_id.module
                     copied_parent = None
```
yajo: This line makes no sense now, isn't it?

website_multi_theme/models/ir_ui_view.py
```diff
@@ -13,12 +14,36 @@
 class IrUiView(models.Model):
     _inherit = 'ir.ui.view'
 
-    multi_theme_generated = fields.Boolean(
```
yajo, line 16: You need a migration script to fill `origin_view_id` from preexisting views that had this boolean set to true.

website_multi_theme/models/website.py
```diff
@@ -25,11 +25,14 @@ class Website(models.Model):
...
+        require=True,
+        default=lambda self: self.env.ref('website_multi_theme.theme_default',
+                                          raise_if_not_found=False)
     )
     multi_theme_view_ids = fields.One2many(
         comodel_name="ir.ui.view",
         inverse_name="website_id",
-        domain=[("multi_theme_generated", "=", True),
+        domain=[("origin_view_id", "!=", False),
                 "|", ("active", "=", True), ("active", "=", False)],
```
yajo, line 36: Maybe we should replace this by `context={"active_test": True}`

website_multi_theme/models/website_theme.py
yajo: In general I recommend to use `|=`, to avoid duplicates in the recordsets.

website_multi_theme/models/website_theme.py
```diff
@@ -36,20 +36,48 @@ class WebsiteTheme(models.Model):
...
         Asset = self.env["website.theme.asset"]
-        for one in self.filtered("converted_theme_addon"):
-            # Get all views owned by the converted theme addon
-            refs = self.env["ir.model.data"].search([
-                ("module", "=", one.converted_theme_addon),
-                ("model", "=", "ir.ui.view"),
-            ])
-            existing = frozenset(one.mapped("asset_ids.name"))
+
+        common_refs = self.env["ir.model.data"]
```
yajo: This line is not needed; the next sentence will return an empty recordset when nothing is found.

website_multi_theme/models/website_theme.py
```diff
@@ -87,6 +115,12 @@ class WebsiteThemeAsset(models.Model):
              "website, and disabled otherwise. Usually used to load assets.",
     )
 
+    auto = fields.Boolean(
```
yajo, line 124: Move fields above, just like [explained here](https://github.com/OCA/maintainer-tools/blob/master/CONTRIBUTING.md#models).

website_multi_theme/views/ir_ui_view.xml
```diff
@@ -0,0 +1,26 @@
...
+        <field name="model">ir.ui.view</field>
+        <field name="inherit_id" ref="base.view_view_form"/>
+        <field name="arch" type="xml">
+
+            <xpath expr="//notebook" position="inside">
+                <page string="Multi-website" attrs="{'invisible': [('copy_ids', '=', [])]}">
+                    <separator string="Copies"/>
+                    <field name="copy_ids"/>
+                </page>
+                <page string="Multi-website" attrs="{'invisible': [('origin_view_id', '=', False)]}">
```
yajo: 2 pages called the same?

Better 1 page and 2 groups.

Remember to add a `name="multiwebsite"`, to enable possible inheritance.

website_multi_theme/wizards/website_config_settings_view.xml
```diff
@@ -15,7 +15,7 @@
             <field name="module_website_form_editor" position="after">
                 <label for="multi_theme_id"/>
                 <div>
-                    <field name="multi_theme_id" class="oe_inline"/>
+                    <field name="multi_theme_id" class="oe_inline" required="1"/>
```
yajo: Why not require it in the model?

website_multi_theme/README.rst
```diff
@@ -137,6 +137,7 @@ Contributors
 * Rafael Blasco <rafael.blasco@tecnativa.com>
 * Antonio Espinosa <antonio.espinosa@tecnativa.com>
 * Jairo Llopis <jairo.llopis@tecnativa.com>
+* Ivan Yelizariev <https://it-projects.info/team/yelizariev>
```
yajo, line 140: I can't believe you weren't here before!

yajo (changes requested):
OK, let me know if I understand what you are doing here:

1. Now all websites have a theme, even if it must be the default one. I like this approach, it reduces variability :blush:

1. Converted assets inherit the `customize_show` attribute independently.

1. All views with `customize_show=True` are considered multiwebsite views. I'm not sure if we should instead duplicate all views belonging to addons that depend on `website`; after all, a user usually can customize any view in a website... :thinking: 

Some comments below. Thanks!

website_multi_theme/models/website.py
yelizariev: I cannot find an example of using such trick. Are you sure that context is used when we read the field in python via ``.multi_theme_view_ids`` and other cases? I'd prefer to keep as it's now

website_multi_theme/models/website_theme.py
yelizariev: The idea was to allow in possible next updates don't change git blame much.

For example,

Version1

common_refs = self.env["ir.model.data"]
common_refs |= A

Version 2

common_refs = self.env["ir.model.data"]
common_refs |= A
common_refs |= B

Version 3 

common_refs = self.env["ir.model.data"]
common_refs |= A1
common_refs |= B
common_refs |= C

Version 5


common_refs = self.env["ir.model.data"]
common_refs |= B
common_refs |= C

website_multi_theme/models/website_theme.py
yelizariev: Do you mean above ``_sql_constraints``? It's actually not explained in the guidelines. It mentions only constrain methods. While ``_sql_constraints`` seems to be part of 

    Private attributes (_name, _description, _inherit, ...)

website_multi_theme/views/ir_ui_view.xml
yelizariev: The idea is don't pass attributes twice in each node here:

                    <separator string="Copies"/>
                    <field name="copy_ids"/>

yelizariev: I can add different names, if it's the concern

    name="multiwebsite_origin_view"
    name="multiwebsite_copied_view"

website_multi_theme/wizards/website_config_settings_view.xml
yelizariev: Because we have default website, which doesn't have multi_theme_id on installation and we cannot set it on installation, because *Default theme* is created after loading python files

yelizariev:
Rebased and added some fixes. Few comments from @Yajo's review are subject to discuss. I'll squash commits once we close all concerns or one can do it on merging (just fist commit message has to be keeped: ``[IMP] make customize_show views part of multi-theme``)

yelizariev:
About this idea:

> All views with customize_show=True are considered multiwebsite views. I'm not sure if we should instead duplicate all views belonging to addons that depend on website; after all, a user usually can customize any view in a website... :thinking:

The possible problem here is that those changes will be destroyed after switching a website to new theme. So, it's better to share them, than letting user face data lost

website_multi_theme/models/ir_ui_view.py
yelizariev: I'll do that

yelizariev:
I'll do this later

- [x] a migration script to fill origin_view_id from preexisting views that had this boolean set to true.

rafaelbn:
Thank you @yelizariev for this great contribution! :smile: :clap: :clap: :clap:

yelizariev:
I added migration scripts and squashed the PR

Previous history:
https://github.com/yelizariev/website/commits/10.0-website_multi_company-multi-customize-before-squashing

yelizariev:
And another commit  https://github.com/yelizariev/website/commit/1984fbf5566e7c21dc1e7942cfaf704badc57d4c
(squashed)

website_multi_theme/models/website.py
yajo: OK no problem, it was just an idea

website_multi_theme/models/website_theme.py
yajo: Yikes! forget this. I think I didn't expand the diff in between :laughing:

website_multi_theme/wizards/website_config_settings_view.xml
yajo: It makes sense. Could you add a note here then please? It's not obvious...

yelizariev:
One more update from @Yajo's review  https://github.com/yelizariev/website/commit/0fd0ad3809aaab98eb58085f4836d8f268387964

and there is still [the question](https://github.com/OCA/website/pull/443#discussion_r180923312) about travis tests. Can we do something about it or just left it as is

website_multi_theme/README.rst
yajo: What about calling reload from the last demo data file read?

yelizariev: Do you mean demo file of **website_blog_category**?

yelizariev: I'd prefer to Reload after creating new ``ir.ui.view`` record with ``customize_show`` (via extending ``create`` method). What do you think, @pedrobaeza ?

ilmir-k:
@yelizariev 
During the tests I found out wrong behavior. When we use ``Default theme`` or ``Multiwebsite Demo Theme`` for websites everything is ok, but if we install and set **Bootswatch Theme** we miss the options for ``Customize``

missed_options

yelizariev:
I think the error was introduced in https://github.com/OCA/website/commit/a4ed42fe08278cff4ec618e785c64dedd35f9d5c#diff-93d51900ef6c23af9b0a13e1278dfad5R45 by @Yajo 

How to reproduce without this updates:

* Install *Bootswatch Theme*
* Click reload twice
* RESULT: *Multi-theme Bootswatch Theme* is disappeared from the list

yelizariev:
@Yajo could you check and fix?

website_multi_theme/README.rst
yajo: That would make sense, yes.

I was talking about demo data file included in this addon. Or even maybe a normal data file.

yajo:
See #447

yelizariev:
Rebased
```

ivann@ivann-360:/mnt/files/odoo/odoo-10.0/website$ git fetch OCA  pull/443/head
De https://github.com/OCA/website
 * branch            refs/pull/443/head -> FETCH_HEAD
ivann@ivann-360:/mnt/files/odoo/odoo-10.0/website$ git diff FETCH_HEAD..HEAD
diff --git a/website_multi_theme/models/website_theme.py b/website_multi_theme/models/website_theme.py
index 45f3ea4..a0190f2 100644
--- a/website_multi_theme/models/website_theme.py
+++ b/website_multi_theme/models/website_theme.py
@@ -65,10 +65,11 @@ class WebsiteTheme(models.Model):
                 # add common_refs only for installed themes
                 refs |= common_refs
 
-            views = self.env["ir.ui.view"].search([
-                ("id", "in", refs.mapped("res_id")),
-                ("type", "=", "qweb"),
-            ])
+            views = self.env["ir.ui.view"].with_context(active_test=False) \
+                .search([
+                    ("id", "in", refs.mapped("res_id")),
+                    ("type", "=", "qweb"),
+                ])
             existing = frozenset(
                 one
                 .mapped("asset_ids")

```

yelizariev:
Update for the last question: https://github.com/yelizariev/website/commit/3b61c5d086ba50a9bd914c4750ec4410f0e4699d

It took more time than I [thought](https://github.com/OCA/website/pull/443#discussion_r182094341)

@Yajo your review is very needed

yelizariev:
Other updates

LINT: https://github.com/yelizariev/website/commit/52b962fa313c9d207d2116b629ec1f97332b37ce
CI fix:  https://github.com/yelizariev/website/commit/6c5a852892339988162b6db4d6182b5c71be936a

rafaelbn:
Please @Yajo could you review here? Thank you!

website_multi_theme/models/ir_ui_view.py
```diff
@@ -13,12 +15,36 @@
...
-        name="Generated from multi theme module",
-        readonly=True,
-        help="Was this view autogenerated by website_multi_theme?",
-    )
     was_active = fields.Boolean(
+        readonly=True,
         help="Indicates if the view was originally active before converting "
              "the single website theme that owns it to multi website mode.",
     )
+
```
yajo: I'd remove this white line

website_multi_theme/models/website.py
```diff
@@ -125,15 +126,21 @@ def _multi_theme_activate(self):
...
-                _logger.info(
-                    "Deleting multi website theme views for %s: %s",
-                    website.display_name,
-                    website.multi_theme_view_ids,
+                default_theme = self.env.ref(
+                    'website_multi_theme.theme_default'
                 )
-                website.multi_theme_view_ids.unlink()
-                continue
+                if not default_theme:
```
yajo: If I'm not wrong, this `if` will never pass because the above call to `ref()` doesn't have a `raise_if_not_found=False`.

OTOH I'd use the try-except-else option instead.

website_multi_theme/models/website_theme.py
```diff
@@ -36,25 +37,54 @@ class WebsiteTheme(models.Model):
...
         """Generate assets for converted themes"""
         Asset = self.env["website.theme.asset"]
-        for one in self.filtered("converted_theme_addon"):
-            # Get all views owned by the converted theme addon
-            refs = self.env["ir.model.data"].search([
-                ("module", "=", one.converted_theme_addon),
-                ("model", "=", "ir.ui.view"),
-            ])
+
+        common_refs = self.env["ir.model.data"]
```
yajo, line 41: You can delete this, the search below will return an empty recordset on his own if no results are found.

yajo (changes requested):
Amazing work, now just minor details

website_multi_theme/models/website_theme.py
yelizariev: My explanation is here: https://github.com/OCA/website/pull/443#discussion_r181154724
You probably miss it. What do you think about it?

yelizariev:
Tiny updates https://github.com/yelizariev/website/commit/e028b325a1329bad03542f29b814c151b2ca4c48

yajo (approved):

yelizariev:
@pedrobaeza ?

ilmir-k (approved):
:+1:

website_multi_theme/__manifest__.py
```diff
@@ -1,12 +1,13 @@
...
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 
 {
     "name": "Website Multi Theme",
     "summary": "Support different theme per website",
-    "version": "10.0.1.1.1",
+    "version": "10.0.1.2.0",
     "category": "Website",
     "website": "https://www.tecnativa.com",
     "author": "Tecnativa, Odoo Community Association (OCA)",
```
pedrobaeza: @yelizariev your company deserves to be here

website_multi_theme/models/ir_model_data.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class IrModelData(models.Model):
+    _inherit = 'ir.model.data'
+
+    @api.model
+    def create(self, vals):
+        res = super(IrModelData, self).create(vals)
+        if vals.get('model') == 'ir.ui.view' \
+           and not self.env.context.get('duplicate_view_for_website'):
```
pedrobaeza: Incorrect indendation and line split

website_multi_theme/models/ir_ui_view.py
```diff
@@ -13,12 +15,35 @@
...
+        string="Copied from",
+        readonly=True,
+        help="View from where this one was copied for multi-website"
+    )
+    copy_ids = fields.One2many(
+        "ir.ui.view",
+        "origin_view_id",
+        string="Copies",
+        readonly=True,
+        help="Duplicates of this view"
```
pedrobaeza, line 34: I would prefer a more explicit name: `multitheme_copy_ids` and a longer explanation in help.

website_multi_theme/models/ir_model_data.py
```diff
@@ -0,0 +1,37 @@
...
+            view = self.env['ir.ui.view'].browse(vals.get('res_id'))
+            if view.customize_show:
+                # Reload theme to make new customize_show available right after
```
pedrobaeza: Put this explanation in the method docstring instead of here, as it's the purpose of the override.

website_multi_theme/models/ir_ui_view.py
```diff
@@ -13,12 +15,35 @@
...
+    )
+
+    @api.model
+    def _customize_template_get_views(self, key, full=False, bundles=False):
+        """This method is used to prepare items
+           in 'Customize' menu of website Editor"""
+        views = super(IrUiView, self)._customize_template_get_views(
+            key, full=full, bundles=bundles
+        )
+
```
pedrobaeza: These empty lines don't add anything useful and PEP8 discourages its use. Better to strip them.

website_multi_theme/models/website.py
```diff
@@ -25,11 +26,14 @@ class Website(models.Model):
         comodel_name='website.theme',
         domain=[("asset_ids.view_id", "!=", False)],
         help="Multiwebsite-compatible theme for this website",
+        require=True,
```
pedrobaeza: Please don't add require fields in inherited classes, as this will make that tests that create records from this model fail when this dependent module is not loaded. Instead, put the require on the view.

website_multi_theme/models/website_theme.py
```diff
@@ -36,25 +37,54 @@ class WebsiteTheme(models.Model):
...
+
+        # add views with customize_show menu, so we can activate them per
+        # website independently
+        common_refs |= self.env['ir.ui.view']\
+                           .with_context(active_test=False)\
+                           .search([
+                               ('website_id', '=', False),
+                               ('customize_show', '=', True),
+                           ]).mapped('model_data_id')
+
```
pedrobaeza: Don't add too much empty lines...

pedrobaeza (changes requested):

website_multi_theme/models/website.py
yelizariev: I believe it's true for existing fields only, while here we create a new field.

pedrobaeza: No, it's the same. The sequence is the following in MQT:

- This module is installed along with all the testable ones and its dependencies.
- Then the SQL NOT NULL constrained is added.
- On the second pass, modules are being tested while being loaded.
- Any module adding a website will have the not null error until this module is loaded.

website_multi_theme/models/ir_model_data.py
yelizariev: Multi-line if-block is my weak side :disappointed:

pedrobaeza: The best is to enclose in parenthesis and always indent second line 2 levels more:
```python
if (vals.get('model') == 'ir.ui.view' and
        not self.env.context.get('duplicate_view_for_website')):
```

yelizariev:
New updates: https://github.com/yelizariev/website/commit/3a24f70032d6db872b795257eb91b698d9dde16d

website_multi_theme/models/website.py
yelizariev: I got it. You are right. 
It seems to be an odoo framework issue. Is there a way to fix it in odoo core?

yelizariev:
And update about ``required=True`` https://github.com/yelizariev/website/commit/b70f910cbcda7a8f02fd9fcbf10db013317bd470

website_multi_theme/models/ir_model_data.py
yelizariev: I'd prefer to put logical operator at the beginning of the line

yelizariev: Oh, it's W503 requirement. OK

yelizariev: I'm not the only one who thinks so.

W503 is ignored by default in latest version of pycodestyle.

https://github.com/OCA/maintainer-quality-tools/issues/545

yelizariev: I said wrong. W504 is ignored by default

yelizariev: But people say that W504 is more logical

pedrobaeza: Well, we can discuss about that in the issue, but right now it's W503 what is used.

yelizariev:
I suppose this is the last question here:

    website_multi_theme/models/ir_model_data.py:33:17: W503 line break before binary operator

I made an issue in MQT https://github.com/OCA/maintainer-quality-tools/issues/545

pedrobaeza:
But that's because you haven't put the `and` in the previous line.

yelizariev:
I know, but I'd prefer to put ``and`` at the beginning of the line. 
why

pedrobaeza:
Yeah, but if you want this merged before discussing about the change, you have to fix it. If not, we have to wait until the discussion is solved.

yelizariev:
I'd prefer to merge it now.
(Travis is green now)

pedrobaeza (approved):

rafaelbn:
Thanks @yelizariev ! :clap: :clap:

PR #444, opened 2018-04-12
[FIX] website_multi_theme: Manage only Qweb views

yajo:
Current implementation disabled all backend views from a theme addon when converting it to multiwebsite mode.

This is now fixed, and includes a migration script to make things work again if you hit this situation in your database.

@Tecnativa

### DIFF
--- a/website_multi_theme/models/website_theme.py
+++ b/website_multi_theme/models/website_theme.py
@@ -42,8 +42,12 @@ class WebsiteTheme(models.Model):
                 ("module", "=", one.converted_theme_addon),
                 ("model", "=", "ir.ui.view"),
             ])
+            views = self.env["ir.ui.view"].search([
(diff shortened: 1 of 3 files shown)


website_multi_theme/models/website_theme.py
```diff
@@ -42,8 +42,12 @@ def _convert_assets(self):
                 ("module", "=", one.converted_theme_addon),
                 ("model", "=", "ir.ui.view"),
             ])
+            views = self.env["ir.ui.view"].search([
+                ("id", "in", refs.mapped("res_id")),
```
pedrobaeza: `refs.mapped("res_id").ids`

pedrobaeza: OK, forget it. I have seen it...

pedrobaeza (approved):

yelizariev (approved):

PR #413, opened 2018-02-01
10.0 port website canonical url to Odoo 10

zeroheure:
I discovered Renka port after working on this. Anyway this one is complete. Readme updated also.

PR #449, opened 2018-04-20
[MIG][10.0] website_canonical_url

simahawk:
replaces https://github.com/OCA/website/pull/413


simahawk:
@zeroheure pls check that your stuff have been preserved

zeroheure:
@simahawk ok for me
The repo miss a static/description folder and an icon file

simahawk:
@zeroheure not required AFAIK

simahawk:
@zeroheure @Yajo reviews? :) Code is exactly the same of #413

simahawk:
it actually changed: templates were duplicated, I kept `templates` folder.

zeroheure (changes requested):
template/layout.xml must contains domain otherwise it is not canonical. See my file https://github.com/zeroheure/website-1/blob/10.0-add-module-website-canonical-url/website_canonical_url/views/template.xml

yajo (approved):
Code OK, just review :x: in bots

zeroheure:
It is better to use absolute URLs rather than relative ones. See these two good explanations about this by moz.org SEO specialists :
(short) https://moz.com/community/q/absolute-vs-relative-canonical-links
(long) https://moz.com/blog/relative-vs-absolute-urls-whiteboard-friday

pedrobaeza:
@zeroheure that's something out of scope of this migration. Any way, Odoo converts any relative URL in code to absolute to the outside.

zeroheure:
@pedrobaeza using absolute URLs is very easy, just use the template.xml file from my PR. Furthermore, the test files contains code I've added to test absolute URLs. Why put this code and the "relative" template.xml together in the PR ?
And no, Odoo will not convert the URL in this tag (try, you will see).

Canonical URL are useful for SEO, it doesn't make sense to not tell the search engin which domain, subdomain and protocol to index (e.g. https://www.domain rather than http://do-main). 
Forgive me if I'm a bit rude, but I've worked hard with the french SEO master Olivier Andrieu (abondance.com) to enhance Odoo for SEO, and I know very well this subject — I have already a few modules ready to publish (GPL) on https://github.com/zeroheure/odoo-seo and a draft of detailed bug reports with suggested changes to Odoo. Please, consider adding absolute URLs in canonical tags.

pedrobaeza:
I don't get how can you get absolute URLs with your template... Any way, I'm not a reviewer of this PR because it's out of my scope. I let reviewers to judge.

zeroheure:
because it request the url_root

simahawk:
regarding tests: I have to get back, no time now, probably in next days.
Regarding implementation and absolute url: to me it sounds weird and I planned to double check that.
I'd say we should produce the final **absolute url** and do it in the method.
Nevertheless, if it worked and still work, I'd keep it like this and change behavior on v11 migration (in  my TODO list).

website_canonical_url/models/website.py
```diff
@@ -0,0 +1,35 @@
...
+                url_parts[2] = lang + url_parts[2]
+                canonical_url = urlunparse(url_parts)
+        # Special case for rerouted requests to root path
+        try:
+            if (canonical_url.startswith("/page/") and
+                    self.menu_id.child_id[0].url == canonical_url):
+                canonical_url = "/"
+        except:  # pragma: no cover
+            pass
+        return canonical_url
```
yajo: This method should return the protocol and hostname too.

What I'm not sure about is if it should get them from the request or from `web.base.url` config param. I think it should be from the param, since that's something you can configure, but then it could be different by website.

Possibly the best solution would be to add a new field to the `website` model, called `canonical_domain`, where one could define the full protocol+hostname for all canonical urls that come from this website.

Then, here, we'd just `return urlparse.urljoin(website.canonical_domain, canonical_url)`

yajo (changes requested):
Indeed absolute URLs are best for canonical purposes.

Just check this excerpt from [the related Google support page](https://support.google.com/webmasters/answer/139066?hl=en):

> **Use absolute paths rather than relative paths** with the `rel="canonical"` link element.
>
> Use this structure:  `https://www.example.com/dresses/green/greendresss.html`
Not this structure: `/dresses/green/greendress.html`

I think that if the world's biggest search engine gives such hint, it should be heard.

OTOH, @zeroheure's patch is wrong. See comment below.

website_canonical_url/models/website.py
simahawk: Totally agree on the abs url. Regarding where to take the domain:
the doc you linked clearly states that we should have 1 single URL no matter the subdomain you might use (like m.mydomain.com for mobile). So, to me the best solution seems to use `web.base.url` (but I might change idea below :smile: )

> but then it could be different by website.

@Yajo I don't get you here: that param is one and only one for one DB thus is going to be the same. So unless you want to serve contents from different instances/DBs the param is going to be always the same. The issue might come when you have multiple websites on the same DB that are not related to each other and don't share the same domain. In this case, yes, probably having the canonical domain on the website object is a good idea.

Am I missing anything?

yajo: > The issue might come when you have multiple websites on the same DB that are not related to each other and don't share the same domain.

You nailed it. `website_multi_theme` addon comes to my mind. It would be affected with this problem and it would be incompatible with `website_canonical_url` then.

Another good thing of having it per website is that several websites could share the same canonical domain, which is actually why the recommendation recommends such thing.

`web.base.url` could be a fallback value if the canonical domain is undefined.

simahawk: Ok, then I'll go for new field + fallback to param. Not sure about the protocol tho, since web.base.url usually does not specify it. For this we might want to stick to the request protocol which should be ok if you setup your web server properly. No?

yajo: Hmmmm AFAICR `web.base.url` always includes the protocol, and the new canonical_domain field should be constrained to force including it always I think.

zeroheure: I came to the same conclusion : it would be better with a new field. But I thought it wouldn't be accepted. I found the solution to call the canonical domain in the template file, because it would be easier to adapt the template to each website.

pedrobaeza:
Can this be finished then with a green Travis status?

simahawk:
I'll get back on it in a while

pedrobaeza:
Ok

website_canonical_url/models/website.py
```diff
@@ -0,0 +1,35 @@
...
+    def get_canonical_url(self, req=None):
+        canonical_url = None
+        if req is None:  # pragma: no cover
+            req = request
+        if req and req.httprequest.path:
+            lang = self.env.lang
+            if lang == request.website.default_lang_code:
+                canonical_url = req.httprequest.path
+            else:
+                url_parts = urlparse(req.httprequest.path)
```
simahawk: @Yajo @zeroheure I'm refactoring and writing tests but I don't get some things here and current implementations seems broken to me:

1. language code is added always when is not default, but is already in the path (and even better is cleaned up in `urlparse(req.httprequest.path).path`) 
2. querystring params are not stripped out which is kind of wrong AFAIK

Am I wrong?

simahawk: to me it should be something like:
```python
parsed = urlparse(req.httprequest.path)
canonical_url = parsed.path
if lang != req.website.default_lang_code:
    for lang, __ in self.get_languages():
        if canonical_url.startswith(lang):
            canonical_url = canonical_url[len(lang):]
            break
```
kind of....

simahawk:
@zeroheure @Yajo done. Fixed, refactored, added configuration and added more tests.
I also bumped 10.0.1.1.0 as the are improvements here.

zeroheure:
@simahawk. Great! I suggest to change README from 

> You may want to check the domain in Settings / Technical / System Parameters / web.base.url

to

> You may want to check the domain in Settings / Technical / System Parameters / web.base.url and set a prefered domain in Website Admin / Settings / Canonical Domain

Also, access rights for files and folders are curently limited to owner.

website_canonical_url/README.rst
```diff
@@ -0,0 +1,61 @@
...
+Website Canonical URL
+=====================
+
+In Website Builder, eCommerce, Blog, etc. Odoo add parameters in URLs, like category,
+page or sorting. This is a bad thing for SEO because it creates DUST (Duplicate URL,
+Same Text) and Duplicate Content. That is to say, multiple URLs that leads to the same
+page search engine's index.
+
+This module doesn't prevent Odoo to generate such URLs, but it helps to reduce the
+search engine's index size by adding canonical URLs for each page :
```
rafaelbn, line 15: How can I test this please?

I go to runbot in Odoo without this module and:
- http://324224-11-0-99d287.runbot9.odoo.com/shop/category/laptops-11
I go to this runbot with this module and:
- http://3324803-449-32d4d7.runbot2.odoo-community.org/shop/category/laptops-11

So how can I test what you say here?
Thank you!

rafaelbn (commented):
I'm not an expert in this matter, I just would like to know how to prove that this module makes :smile:

website_canonical_url/README.rst
zeroheure: Look for a "link rel=canonical" tag in the html source (in the head part).
You're right, the README could be more explanatory

website_canonical_url/README.rst
```diff
@@ -0,0 +1,61 @@
...
+It insert a HTML tag in the html header that contains a `canonical URL <https://support.google.com/webmasters/answer/139066>`_
+for the current page, no matter what query it gets, and `rel=next and rel=prev links
+<https://webmasters.googleblog.com/2011/09/pagination-with-relnext-and-relprev.html>`_
+wherever a pager is found.
```
zeroheure: add "You will find these tags in the html source code, under the head section."

rafaelbn (approved):
Thanks @simahawk !

Thanks @zeroheure for explanation :smile:

rafaelbn:
Still pending some comments from @zeroheure  but in my side is :+1:

simahawk:
@zeroheure README improved

website_canonical_url/README.rst
```diff
@@ -13,18 +13,43 @@ page search engine's index.
...
+-------------
+
+Canonical URL is absolute. The domain name by default matches
+
+`Settings / Technical / System Parameters / web.base.url`
+
+This might not be enough to make sure that you have always one and only one URL
+to access your resources.
+
+You can force the domain by setting `Canonical domain` field into website settings`.
```
yajo: Same as above.

Besides, remove the quote at the end.

website_canonical_url/README.rst
```diff
@@ -13,18 +13,43 @@ page search engine's index.
...
-wherever a pager is found.
+whenever a pager is found.
+
+
+Configuration
+-------------
+
+Canonical URL is absolute. The domain name by default matches
+
+`Settings / Technical / System Parameters / web.base.url`
```
yajo: Do not use backticks because those have special meanings in RST. Use `"` or `'` instead to avoid confusing the engine please :blush:

website_canonical_url/README.rst
```diff
@@ -13,18 +13,43 @@ page search engine's index.
...
+to access your resources.
+
+You can force the domain by setting `Canonical domain` field into website settings`.
+
+
+How to verify
+-------------
+
+To check this on your website inspect the source of a page: 
+you'll find canonical and prev/next `<link />` elements inside `<head />`.
```
yajo: Again, these backquotes should be double for RST to indicate a code chunk.

website_canonical_url/README.rst
```diff
@@ -13,18 +13,43 @@ page search engine's index.
...
+
+On a demo site (like Runbot) you can go to `/page/website_canonical_url.canonical_demo`
```
yajo: You know...

website_canonical_url/README.rst
```diff
@@ -13,18 +13,43 @@ page search engine's index.
...
 
-The canonical URLs are absolute ; the domain name is the one you've setup for your company.
-You may want to check the domain in Settings / Technical / System Parameters / web.base.url
 
 .. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
    :alt: Try me on Runbot
    :target:  https://runbot.odoo-community.org/runbot/186/10.0
 
+
+
```
yajo: :fire:

website_canonical_url/demo/pages.xml
```diff
@@ -1,14 +1,17 @@
...
+     Copyright 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 <odoo>
 
     <template id="canonical_demo"
               page="True"
               name="Canonical URL and pager demo">
+        <!-- Add a fake pager to test the rel=prev/next links -->
+        <t t-set="pager"
+           t-value="website.pager(request.httprequest.path, 100, 2, url_args=request.httprequest.args)"/>
```
yajo: Does this change obey any hidden or changed logic that I'm not aware of? :thinking: AFAIK, if you define a `t-set` inside a `t-call`, you're defining those variables only for the `t-call`

website_canonical_url/models/res_config.py
```diff
@@ -0,0 +1,17 @@
...
+
+from odoo import models, fields
+
+
+class WebsiteConfigSettings(models.TransientModel):
+
+    _inherit = 'website.config.settings'
+
+    canonical_domain = fields.Char(
+        string='Canonical Domain',
```
yajo: No need to repeat param in a related field.

website_canonical_url/models/res_config.py
```diff
@@ -0,0 +1,17 @@
...
+        related='website_id.canonical_domain',
+        help='Canonical domain is used to build unique canonical URLs '
```
yajo: Same here.

website_canonical_url/models/website.py
```diff
@@ -1,35 +1,63 @@
...
 
+    canonical_domain = fields.Char(
+        help='Canonical domain is used to build unique canonical URLs '
+             'to make SEO happy.'
+    )
+
+    @api.constrains('canonical_domain')
+    def _check_canonical_domain(self):
+        domain = self.canonical_domain
+        if domain and not domain.startswith('http'):
```
yajo: What if your website is `httpstuff.com`?

I'd make sure also there's no ending `/`

website_canonical_url/models/website.py
```diff
@@ -1,35 +1,63 @@
...
-            if lang == request.website.default_lang_code:
-                canonical_url = req.httprequest.path
-            else:
-                url_parts = urlparse(req.httprequest.path)
-                url_parts = list(url_parts)
-                # change the path of the url
-                url_parts[2] = lang + url_parts[2]
-                canonical_url = urlunparse(url_parts)
+        req = req or request
+        lang = getattr(req, 'lang', None) or self.env.lang
```
yajo: Is there any real chance of `req.lang` not being the same as `self.env.lang`? :thinking:

website_canonical_url/models/website.py
```diff
@@ -1,35 +1,63 @@
...
+        parsed = urlparse(req.httprequest.path)
+        canonical_url = parsed.path
+        if lang != req.website.default_lang_code:
+            for avail_lang, __ in self.get_languages():
+                avail_lang = avail_lang.encode('utf-8')
+                if canonical_url.startswith('/' + avail_lang):
+                    canonical_url = canonical_url[len(avail_lang) + 1:]
```
yajo: I think the logic here is wrong.

If you're on the `it_IT` version of the website, the canonical URL for `http://example.com/it_IT/events?show_all=true` should be `http://example.com/it_IT/events`.

If we canonicalize only the default language version, search engines would never index other languages. The canonical tag **must** include the lang if it's not the default.

For you to understand better, the metatag used to indicate a page translations is [hreflang](https://support.google.com/webmasters/answer/189077), and Odoo ships it by default.

website_canonical_url/tests/test_rendering.py
```diff
@@ -0,0 +1,57 @@
...
+from odoo.modules.module import get_resource_path
+from odoo import tools
+from lxml.html import document_fromstring
+
+
+def load_xml(cr, module, filepath, kind='demo'):
+    tools.convert_file(
+        cr, module,
+        get_resource_path(module, filepath),
+        {}, mode='init', noupdate=False, kind=kind)
```
yajo: This one is a very interesting pattern that could help resolving the eternal declare-in-test-case vs. declare-in-demo-data debate (ref. https://github.com/OCA/maintainer-tools/pull/213).

IMHO it should appear in OCA docs.

OTOH, I'm not sure we need a separate method that just calls another method and is used just once...

website_canonical_url/tests/test_rendering.py
```diff
@@ -0,0 +1,57 @@
...
+
+class UICase(HttpCase):
+    def setUp(self):
+        super(UICase, self).setUp()
+        self._reload_page()
+        self.website = self.env['website'].browse(1)
+        self.domain = self.website._get_canonical_domain()
+        self.path = "/page/website_canonical_url.canonical_demo"
+        self.url_absolute = self.domain + self.path
+        self.qstring = "?ultimate_answer=42"
```
yajo: :rofl:

website_canonical_url/tests/test_rendering.py
```diff
@@ -0,0 +1,57 @@
...
+        self.url_full = "%s%s" % (self.url_absolute, self.qstring)
+        self.url_data = self.url_open(self.url_full)
+        self.doc = document_fromstring(self.url_data.read())
+
+    def _reload_page(self):
+        # if you run tests more than once (locally for instance)
+        # sp if you update the page we make sure is reloaded.
+        self.env.ref('website_canonical_url.canonical_demo').unlink()
+        load_xml(self.cr, 'website_canonical_url', 'demo/pages.xml')
```
yajo: AFAIK in these kind of tests, you need to get a special cursor:

```python
with self.cursor() as cr:
  load_xml(cr, 'website_canonical_url', 'demo/pages.xml')
```

website_canonical_url/tests/test_rendering.py
```diff
@@ -0,0 +1,57 @@
+# -*- coding: utf-8 -*-
```
yajo: Could you avoid renaming this file? Thanks

website_canonical_url/models/website.py
```diff
@@ -1,35 +1,63 @@
...
-                url_parts[2] = lang + url_parts[2]
-                canonical_url = urlunparse(url_parts)
+        req = req or request
+        lang = getattr(req, 'lang', None) or self.env.lang
+        parsed = urlparse(req.httprequest.path)
+        canonical_url = parsed.path
+        if lang != req.website.default_lang_code:
+            for avail_lang, __ in self.get_languages():
+                avail_lang = avail_lang.encode('utf-8')
+                if canonical_url.startswith('/' + avail_lang):
```
yajo: Better `"/%s/" % avail_lang`

yajo (changes requested):

website_canonical_url/tests/test_rendering.py
simahawk: `test_rendering` sounded more clear then `test_tags` but I don't mind that much :wink:

website_canonical_url/demo/pages.xml
simahawk: no, you are right. Has been moved out but should get back inside. Just forgot it.

website_canonical_url/models/website.py
simahawk: Well, getting it from the request is more correct IMO, also because I don't know - honestly - if they match always, especially in tests.
I'm trying do be as defensive as I can and this is not hurting anybody I think :)

website_canonical_url/models/website.py
simahawk: @Yajo seems legit but then I don't get WTH we are dealing w/ languages wheres we could just strip extra stuff from the path. 

The old implementation here https://github.com/OCA/website/blob/9.0/website_canonical_url/models/website.py#L20
was doing nothing for the main lang and forcing lang into path for other langs.
BUT  if you get a request for `/it_IT` that piece is already in the path (w/out checking if the lang was already in the path BTW).

So, to sum it up... assuming that `it_IT` is the main lang, we should:

1. `foo.com/it_IT?foo=baz` -> canonical URL must be `foo.com` BUT this is not needed because Odoo already redirects to root URL for main lang. So, all in all: nothing to do other than stripping out extra params;

2. `foo.com/es_ES?foo=baz` -> canonical URL must be `foo.com/es_ES` BUT odoo already redirects you to the proper lang path AFAIK, so there's nothing to do here again other then stripping args.

Am I missing any strange behavior? Can we have secondary lang URLs w/out lang in the path?

simahawk: ah, also the old implementation (like current) seems broken in the special case handling for `/page/homepage` check since is not considering the language and set the url always to `/`.

website_canonical_url/models/website.py
simahawk: fixed

website_canonical_url/models/res_config.py
simahawk: fixed

simahawk:
@Yajo should be better now :)

website_canonical_url/models/website.py
```diff
@@ -1,35 +1,75 @@
...
+        canonical_url = parsed.path
+        lang = (
+            getattr(req, 'lang', None) or
+            self.env.lang or
+            req.website.default_lang_code
+        )
+        lang_path = ''
+        if lang != req.website.default_lang_code:
+            # eg: language is `it_IT` but main lang is `en_US`
+            lang_path = '/%s/' % lang
```
yajo: OK, I'm remembering the _why_ behind previous implementation. This is gonna simplify all the algorithm.

The main point is that **the request you will use will never start by the lang code**. If you read [here](https://github.com/odoo/odoo/blob/79bbd1dcb8c24d4be26807494d8d670faff9d7e3/addons/website/models/ir_http.py#L198-L202), you'll see that it gets rerouted, so your code must never care about the path starting or not by the lang, and you can just use `request.lang`.

So, this would be mostly the correct algorithm:

```python
canonical_url = req.httprequest.path
if canonical_url == "/page/homepage":
    canonical_url = "/"
if request.lang != request.website.default_lang_code:
    canonical_url = u"/{}{}".format(request.lang, canonical_url)
return canonical_url
```

website_canonical_url/tests/test_url.py
```diff
@@ -0,0 +1,121 @@
...
+            self.website._get_canonical_relative_url(req=req),
+            '/'
+        )
+
+    def test_canonical_relative_url_home_special_case_lang_no_match(self):
+        req = FakeRequest(self.website, '/en_US/page/homepage?foo=baz')
+        self.website.default_lang_id = self.env.ref('base.lang_it')
+        self.assertEqual(
+            self.website._get_canonical_relative_url(req=req),
+            '/en_US'
```
yajo: This should be `'/en_US/`, and then you should make the test pass.

If I'm not wrong, with the above algorithm I told you, it should be enough.

website_canonical_url/tests/test_url.py
```diff
@@ -0,0 +1,121 @@
...
+            'https://oh.yeah/'
+        )
+
+    def test_canonical_url_home_special_case_no_match(self):
+        self.website.canonical_domain = 'https://oh.yeah'
+        req = FakeRequest(
+            self.website, '/it_IT/page/homepage?foo=baz', lang='it_IT')
+        self.assertEqual(
+            self.website.get_canonical_url(req=req),
+            'https://oh.yeah/it_IT'
```
yajo: Same here: `'https://oh.yeah/it_IT/'`

yajo (changes requested):

website_canonical_url/views/website_views.xml
```diff
@@ -0,0 +1,26 @@
...
+    </field>
+  </record>
+
+  <record id="view_website_config_settings" model="ir.ui.view">
+    <field name="name">website settings canonical form</field>
+    <field name="model">website.config.settings</field>
+    <field name="inherit_id" ref="website.view_website_config_settings" />
+    <field name="arch" type="xml">
+      <field name="website_name" position="after">
+        <field name="canonical_domain" placeholder="https://mydomain.com" />
```
yajo: Oh and please, be consistent with indentation: 4 spaces.

website_canonical_url/models/website.py
simahawk: @Yajo actually is the other way around: it's rerouted [only if default lang is used](https://github.com/odoo/odoo/blob/79bbd1dcb8c24d4be26807494d8d670faff9d7e3/addons/website/models/ir_http.py#L181):
> if lang in url but not the displayed or default language --> change or remove
> or no lang in url, and lang to dispay not the default language --> add lang

so you definitely get the request w/ the lang in the path. And, this is ok and we don't have to check if the lang is there or not because is going to be there always, when needed.

The only exception to handle is the root menu item but we cannot handle /page/homepage only.
Anyway, I got the point so I'm going to adapt it.

simahawk:
@Yajo fixed

website_canonical_url/models/website.py
```diff
@@ -0,0 +1,69 @@
...
+        parsed = urlparse(req.httprequest.path)
+        # here we get the full path and qstring args stripped out already
+        canonical_url = parsed.path
+        # build lang path if request lang does not match default lang
+        lang_path = ''
+        if req.lang != req.website.default_lang_code:
+            # eg: language is `it_IT` but main lang is `en_US`
+            lang_path = '/%s/' % req.lang
+        # handle special case for rerouted requests to root path
+        if self._is_root_page(canonical_url, lang_path=lang_path):
```
yajo: So, after all, leave this implementation as it was, which was already working fine (except for the missing hostname) and taking into account redirections and reroutings:

https://github.com/OCA/website/blob/6ec9b1598a9ad6f92186f338562fbdbd336d3439/website_canonical_url/models/website.py#L14-L35

If you really want to improve it, I'd better concentrate in the root page detection algorithm, which should be exactly the same as upstream if we don't want to hit some corner cases, but nothing else.

yajo (changes requested):
Nope, not fixed:

captura de pantalla de 2018-05-11 12-51-32

There are 2 different concepts I think you are mixing:

1. [Redirection](https://github.com/odoo/odoo/blob/79bbd1dcb8c24d4be26807494d8d670faff9d7e3/addons/website/models/ir_http.py#L195), done when the user expects a specific, non-default language, or when the user specifies the default language but it should be stripped. It's a very well known HTTP 30X redirection.
2. [Rerouting](https://github.com/odoo/odoo/blob/79bbd1dcb8c24d4be26807494d8d670faff9d7e3/addons/website/models/ir_http.py#L202). It's an Odoo internal system that allows the developer to code controllers without taking care about the language. The developer will always find the lang in `request.lang`, no matter if it is the default or not, and will find the lang-less path in `request.httprequest.path`.

Both things are automated, and if I'm not wrong, it means that https://github.com/OCA/website/pull/449#discussion_r185716668 is still true (except for the part of the 1st menu item, as you said correctly :wink:).

simahawk:
@Yajo dammit, now I got it... thanks for the details :sweat_smile: 
I fixed it but I'd like to add one more test since my tests were passing because of the hardcorded full path in the fake request.

Regarding 1st menu item retrieval: if I'm not wrong, here's where they handle it https://github.com/odoo/odoo/blob/79bbd1dcb8c24d4be26807494d8d670faff9d7e3/addons/website/controllers/main.py#L64

website_canonical_url/models/website.py
```diff
@@ -1,35 +1,73 @@
...
+        canonical_url = parsed.path
+        # build lang path if request lang does not match default lang
+        lang_path = ''
+        if req.lang != req.website.default_lang_code:
+            # eg: language is `it_IT` but main lang is `en_US`
+            lang_path = '/%s/' % req.lang
+        # requests might be redirected or rerouted,
+        # when this happens here we don't have the lang in the path anymore
+        if lang_path and not canonical_url.startswith(lang_path):
+            canonical_url = urljoin(lang_path, canonical_url.lstrip('/'))
```
yajo: Oh come on, not again... :man_facepalming: 

Please, friend, I really don't wanna be rude, really :sweat_smile: :v: , but requests *always* get rerouted. Don't complicate things, it's extremely simple:

```py
if "/" in getattr(req, "rerouting", []) or req.httprequest.path == self.menu_id.child_id[:1].url:  # <-- This can go into `_is_root_page()` if you want, and you can use the more complex algorithm you linked before also if you want
    return "/"
if req.lang != req.website.default_lang_code:
    return "/%s%s" % (req.lang, req.httprequest.path)
return req.httprequest.path  # This already strips out hostname and query string
```

Please KISS.

simahawk: Oh come on, not again: is this "KISSy" solution more clear? I don't think so :stuck_out_tongue: 

What you have more in my solution is:

1. parsing the path -> I can remove it if that's stripped already [done]
2. check if the lang is there in the path, which I consider more safe anyway (WOW, what a non KISSy thing!!).
Also, building the lang path IMHO is better when you have to test the menu item: users can copy paste URLs w/ the lang in the URL (yes, they can do it :sweat_smile:) so in that case my check is more defensive
3. I moved the check for the root menu to make things more readable and hackable in case

yajo: > check if the lang is there in the path, which I consider more safe anyway (WOW, what a non KISSy thing!!).

Yes, it's not KISSy. Simply because the lang will **never** be in the path.

> users can copy paste URLs w/ the lang in the URL

Imagine this case:

- Website default lang is en_US
- Website alt lang is es_ES
- User sets the 1st menu item URL to `/es_ES/`
- Somebody clicks in the main link

The result of this is an endless 302 redirection loop.

So, if users copy a lang-specific link in a menu item, the canonical url will only be one of the problems it will produce. There will be more. That's simply a **not supported** scenario.

BTW, while editing a website, you always edit in the default language, because other ones only have translation mode, so it's not much likely that will happen.

So, basically, any scenario that justifies this code is simply impossible. Remove it please.

pedrobaeza:
@Yajo please finish your review here

yajo (approved):

yajo:
@simahawk Can you please squash migration-related commits to merge? Thanks! :blush:

pedrobaeza:
Rebased and merging when CIs finish

zeroheure:
Great ! thanks to all !

PR #454, opened 2018-04-25
[IMP] support for sub-themes

yelizariev:
closes #409

### DIFF
--- a/website_multi_theme/README.rst
+++ b/website_multi_theme/README.rst
@@ -8,6 +8,37 @@ Website Multi Theme
 
 Allow the website admin to set a different theme for each website.
 
+The *theme* might be not just a theme-module, but any set of themes and even
+particular views from any module (e.g. view ``website.custom_footer`` from
+``website`` module). It also means, that *theme* is not just a styling, but
+a content as well.
+
+How it works
+============
+
+Core idea is as following
+
+* Find views created by *theme-module* and mark them as *multi-views* (``website.theme.asset``) additionally to one specified manually via XML (see `demo/themes.xml <demo/themes.xml>`_ as an example). The method `_convert_assets <models/website_theme.py>`_ is responsible for it.
+
+* Set ``active`` to ``False`` for *multi-views*. See method `_find_and_deactivate_views <models/website_theme.py>`_.
+
+* Apply *Multi-theme* (record in new model ``website.theme``) to the specific
+  website. See method `_multi_theme_activate <models/website.py>`_
+
+  * Make some magic with technical views ``website.assets_frontend`` and ``website.layout``.
+
+    * Duplicate *patterns* from `templates/patterns.xml <templates/patterns.xml>`_
+    * In ``layout_pattern`` replace ``{theme_view}`` placeholder to a duplicate
+      of ``assets_pattern``
+    * Corresponding duplicated *pattern* will be used as a new value for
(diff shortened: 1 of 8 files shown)


yelizariev:
No time for relaxing @Yajo @pedrobaeza

website_multi_theme/models/website.py
```diff
@@ -183,23 +187,26 @@ def _multi_theme_activate(self):
...
-                    data = etree.fromstring(copied_view.arch)
-                    data.attrib["inherit_id"] = new_parent.key
-                    copied_view.arch = etree.tostring(data)
+                    copied_view._replace_parent(new_parent)
 
                 custom_views |= copied_view
             # Delete any custom view that should exist no more
-            (website.multi_theme_view_ids - custom_views).unlink()
+            views_to_remove = website.multi_theme_view_ids - custom_views
+            # Removed views can be copied parent for someone.
```
yajo: "Removed views could be a copied parent for others"

website_multi_theme/data/themes_default.xml
```diff
@@ -8,14 +8,8 @@
...
-    <template id="theme_default_dummy" name="Dummy asset for Default Theme">
-        <!-- It's empty, but we have to have one, because otherwise multi-theme
-             magic is not applied for themes without any views -->
-    </template>
-
-
-    <record model="website.theme.asset" id="theme_default_asset">
-        <field name="name">website_multi_theme.theme_default_dummy</field>
+    <record model="website.theme.asset" id="theme_default_asset_footer_custom">
+        <field name="name">website.footer_custom</field>
```
yajo, line 12: Shouldn't we include the header by default also?

yajo (commented):
Ooook... I see the code and I think I sorta understand it somehow... but TBH I need you to draw here the big picture on what, why and how this. Could you please? :innocent:

website_multi_theme/data/themes_default.xml
yelizariev: Which header do you mean? Could you say xmlid of the view

yelizariev:
So, the big picture is:

we copy for each website not just views of one theme (module), but views of other themes ("dependencies"). Those themes are not necessarily technical dependencies of the main module. Moreover, a theme might be not a module at all (e.g. "Default Theme").

This updates along with previous (merged) updates also allow converting not just a theme to multi-theme view, but a specific view to multi-website view (as it's done with ``website.custom_footer``). I'll add this to README

yelizariev:
I've updated the README with a hope that it brings more developers that understand what is going on here :smiley_cat:

website_multi_theme/data/themes_default.xml
yajo: You're right, it's not a header. it was `website.layout`.

yajo (approved):
OK I think I got it now.

website_multi_theme/README.rst
```diff
@@ -8,6 +8,37 @@ Website Multi Theme
 
 Allow the website admin to set a different theme for each website.
 
+The *theme* might be not just a theme-module, but any set of themes and even
+particual views from any module (e.g. view ``website.custom_footer`` from
```
pedrobaeza: s/particual/particular

pedrobaeza (approved):
Little nitpicking

yelizariev:
Spelling is fixed. The PR is green

PR #457, opened 2018-05-01
[10.0][FIX][website_multi_theme] button method cannot be api.model

yelizariev:
While the method doesn't use recordset, it must be decorated with ``api.multi``, otherwise we get error on clicking the button:

```
  File "/mnt/odoo-source/addons/web/controllers/main.py", line 896, in call_button
    action = self._call_kw(model, method, args, {})
  File "/mnt/odoo-source/addons/web/controllers/main.py", line 884, in _call_kw
    return call_kw(request.env[model], method, args, kwargs)
  File "/mnt/odoo-source/odoo/api.py", line 687, in call_kw
    return call_kw_model(method, model, args, kwargs)
  File "/mnt/odoo-source/odoo/api.py", line 670, in call_kw_model
    recs = self.with_context(context or {})
  File "/mnt/odoo-source/odoo/models.py", line 4911, in with_context
    context = dict(args[0] if args else self._context, **kwargs)
TypeError: cannot convert dictionary update sequence element #0 to a sequence
```

### DIFF
--- a/website_multi_theme/data/themes_default.xml
+++ b/website_multi_theme/data/themes_default.xml
@@ -16,5 +16,6 @@
     <!-- This method prepare assets and applies default theme for all websites -->
     <function model="website.config.settings"
               name="multi_theme_reload"
+              eval="([],)"
               />
 </odoo>
--- a/website_multi_theme/models/website.py
+++ b/website_multi_theme/models/website.py
@@ -202,9 +202,13 @@ class Website(models.Model):
             views_to_remove = website.multi_theme_view_ids - custom_views
             # Removed views could be a copied parent for others
             # So, replace to original parent first
-            for view in self.env['ir.ui.view'].search([
-                    ('inherit_id', 'in', views_to_remove.ids)
-            ]):
+            views_to_replace_parent = \
+                self.env['ir.ui.view']\
+                    .with_context(active_test=False)\
+                    .search([
+                        ('inherit_id', 'in', views_to_remove.ids)
+                    ])
+            for view in views_to_replace_parent:
                 view._replace_parent(view.inherit_id.origin_view_id)
             views_to_remove.unlink()
             _logger.info(
--- a/website_multi_theme/wizards/website_config_settings.py
+++ b/website_multi_theme/wizards/website_config_settings.py
@@ -4,7 +4,7 @@
 
 import logging
 
-from odoo import fields, models, api
+from odoo import fields, models
 
 _logger = logging.getLogger(__name__)
 
@@ -14,7 +14,6 @@ class WebsiteConfigSettings(models.TransientModel):
 
     multi_theme_id = fields.Many2one(related="website_id.multi_theme_id")
 
-    @api.model
     def multi_theme_reload(self):
         """Update multiwebsite themes when loading a new wizard."""
         _logger.info("Reloading available multi-website themes")
--- a/website_multi_theme/__manifest__.py
+++ b/website_multi_theme/__manifest__.py
@@ -7,7 +7,7 @@
 {
     "name": "Website Multi Theme",
     "summary": "Support different theme per website",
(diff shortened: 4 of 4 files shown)


website_multi_theme/data/themes_default.xml
```diff
@@ -15,6 +15,6 @@
 
     <!-- This method prepare assets and applies default theme for all websites -->
     <function model="website.config.settings"
-              name="multi_theme_reload"
+              name="_multi_theme_reload"
```
pedrobaeza: Can't this call an @api.multi method?

website_multi_theme/models/website.py
```diff
@@ -202,9 +202,11 @@ def _multi_theme_activate(self):
             views_to_remove = website.multi_theme_view_ids - custom_views
             # Removed views could be a copied parent for others
             # So, replace to original parent first
-            for view in self.env['ir.ui.view'].search([
-                    ('inherit_id', 'in', views_to_remove.ids)
-            ]):
+            for view in self.env['ir.ui.view']\
```
pedrobaeza: Better for readability now to store in a variable the search result and then use it in the for.

website_multi_theme/data/themes_default.xml
yelizariev: It raise error. At least in the call without args. It would probably work with ``eval="([],)"``

pedrobaeza: Then I think it's better to try for simplifying the code

website_multi_theme/models/website.py
```diff
@@ -202,9 +202,13 @@ def _multi_theme_activate(self):
             views_to_remove = website.multi_theme_view_ids - custom_views
             # Removed views could be a copied parent for others
             # So, replace to original parent first
-            for view in self.env['ir.ui.view'].search([
-                    ('inherit_id', 'in', views_to_remove.ids)
-            ]):
+            views_to_replace_parent = \
+                self.env['ir.ui.view']\
+                    .with_context(active_test=False)\
```
yelizariev, line 207: While it may confuse when to use and when don't use ``active_test`` in this module. But in this case is easier: we must replace parent for all views because otherwise, we get constraint error (deleted record is still referenced), Postgres constraint doesn't care about ``active`` field value

website_multi_theme/data/themes_default.xml
```diff
@@ -16,5 +16,6 @@
     <!-- This method prepare assets and applies default theme for all websites -->
     <function model="website.config.settings"
               name="multi_theme_reload"
+              eval="([],)"
```
yelizariev, line 19: it's required for calling ``@api.multi`` method

yelizariev (commented):
@Yajo take a look on this PR please

website_multi_theme/data/themes_default.xml
yelizariev: It works

website_multi_theme/wizards/website_config_settings.py
```diff
@@ -14,7 +14,7 @@ class WebsiteConfigSettings(models.TransientModel):
 
     multi_theme_id = fields.Many2one(related="website_id.multi_theme_id")
 
-    @api.model
+    @api.multi
```
yajo: I'd better remove the decorator. Since Odoo v10, no decorator means multi.

yajo (approved):

website_multi_theme/wizards/website_config_settings.py
yelizariev: I guessed that it's so, but kept it for the sake of explicitivity

yelizariev:
Can this be merged? There is still more work to do and merging this allows us to test the module with website_multi_company

pedrobaeza (approved):

PR #395, opened 2017-10-30
 [MIG] website_form_recaptcha: Migrate to 11.0

bouvyd:

PR #458, opened 2018-05-07
[MIG] 11.0 mig website form recaptcha

hmrodrigues:


hmrodrigues:
@pedrobaeza @Yajo I've fixed the travis issues
Codecov fails because the module doesn't exist on the target branch right?

pedrobaeza (approved):

yajo (approved):

yajo:
I'll take the freedom to merge, since this comes from #395 and the same code had approvals there.

PR #461, opened 2018-05-15
[10.0][IMP] website_cookie_notice: use specific cookie "accepted_cookie"

wpichler:


website_cookie_notice/__manifest__.py
```diff
@@ -9,6 +9,7 @@
     'category': 'Website',
     'author': "Agile Business Group, "
               "Tecnativa, "
+              "Callino"
```
rafaelbn: I'm very thankful for you contribution and improvement of this module, but I'm not sure if is fair to add you as an author after 2 lines of code. Usually contributor it is.

rafaelbn (commented):
Waiting for runbot to test it. Travis is :red_circle:  Thanks @wpichler :smile:

wpichler:
Sorry - i thought this is necessary...

tarteo:
Hi @wpichler can you explain why this addition is needed, please?

wpichler:
there has been some rare cases where the normal way did not work - so setting this extra cookie does not hurt

tarteo (approved):

rafaelbn:
@Yajo could you please review this code changes?

website_cookie_notice/templates/website.xml
```diff
@@ -10,7 +10,7 @@
 
 <template id="message" name="Cookies notice">
     <div t-if="request and
-        not request.session.get('accepted_cookies')"
+        not request.session.get('accepted_cookies') and not request.httprequest.cookies.get('accepted_cookies')"
```
yajo: Use only the new method to check cookie acceptance. Remove the old one please.

yajo (changes requested):
I like very much the idea, and IMHO this is much better implementation.

The problem with the previous one is that, when a session is destroyed (it is expired in the server, or user logs out), the user is asked yet again to allow cookies.

With this patch, the cookie acceptance is stored in a separate cookie, unrelated to user's session, so user shouldn't care about the cookie message anymore, no matter how many times he logs in or out.

(BTW, you should be the one explaining your work, not me, @wpichler :wink: )

But if we are going to use this new method, then please remove the old one. No need to have both.

wpichler:
Sorry for not explaining - i'll try to explain better from now on...

Removing the old method is no problem - this would include removing the controller function...

While thinking about that - i'd had another idea to improve it...

* Add a new field "cookies_accepted" - Bool default false to res.partner
* When the cookie gets accepted - then call the controller function only if the current session is a non public session. 
* The controller function then does set the field cookies_accepted on the corresponding partner record
* On Template we do check for this field if it is already set or not

So registered users will get complete away from the cookie notice - also on different browsers / devices...

What do you think ?

yajo:
Hmm I think that better not... :thinking:  I might accept cookies in my computer but not in my phone, for instance. I think the best choice is just the specific cookie as you were doing here.



> Removing the old method is no problem - this would include removing the controller function...

Please do! I :heart: code removal :yum:

wpichler:
Thats a valid point - so i've removed code now - and felling happy with it ;-)

simahawk (approved):
LG, thanks for the cleanup

yajo (approved):
Thanks! I rebuilt Travis; let's see if it goes :heavy_check_mark: now.

simahawk:
Tested on runbot, build is green: merging!
@wpichler thanks!

pedrobaeza:
Isn't this cookie notice deprecated since GDPR? If not, this should be forward-ported to v11

wpichler:
No - it has become more important since gdpr

pedrobaeza:
Well, I don't think so if you don't need any explicit consent. And for that, you need to modify current message. @Yajo, you are our GDPR expert. Can you talk about this?

wpichler:
I think it is needed - take a look here: https://eugdprcompliant.com/cookies-consent-gdpr/

wpichler:
Also - this is something that the website owner has to decide - we do only provide the tools for it

pedrobaeza:
> I think it is needed - take a look here: https://eugdprcompliant.com/cookies-consent-gdpr/

I remember that there was a clause that if the cookie is put for making legitimately work the website (not for tracking), it's not needed

> Also - this is something that the website owner has to decide - we do only provide the tools for it

Of course, but promoting something that is not useful at all is not the best way of optimizing flows. If we provide this module without more explanations on new legal topic, people will install it without knowing the real need.

wpichler:
Providing a explanation on a legal topic is in my opinion dangerous - if we are not correct with our interpretation of the topic - then we have given a wrong explanation.

yajo:
This is part of [GDPR](https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX:32016R0679), Article 6:

> Lawfulness of processing
> 1.   Processing shall be lawful only if and to the extent that at least one of the following applies:

> (a) the data subject has given consent to the processing of his or her personal data for one or more specific purposes;

> (b) processing is necessary for the performance of a contract to which the data subject is party or in order to take steps at the request of the data subject prior to entering into a contract;

So, if you want to use option B, you can skip this addon; but if you want to use option A, you can use it.

OCA is not your lawyer, we just share code here. If more than 1 user in the world wants option A, it makes sense to share this addon here. However, cnsult your lawyer to know whether to use it or not.

PR #462, opened 2018-05-21
[MIG][11.0] website_seo_redirection

jcarlosmontoya:
Migrate website_seo_redirection module 10.0 to 11.0


simahawk:
@jcarlosmontoya I'm not sure we need this on v11: redirection tool is already there. It's hidden: you have to enable debug mode to see it in website config. Any improvements we can cherry pick from here?

jcarlosmontoya:
thank you @simahawk , but i'm using for blog, y does not work, 
url_from: /blog/blog1/myblogexample
url_to: /myblogexample

Am i doing somethig wrong?

simahawk:
@jcarlosmontoya seems so. Can you reproduce it on runbot and file and issue  at https://github.com/odoo/odoo/issues ?

tarteo:
It's better to use standard model website.redirect and only use this module to render the Atom feed (website_blog) and sitemap.xml correctly. The standard functionality of Odoo works fine except for that.

simahawk:
@jcarlosmontoya ping :)

website_seo_redirection/__init__.py
```diff
@@ -0,0 +1,5 @@
+# -*- coding: utf-8 -*-
+# © 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
```
nikul-serpentcs, line 2: @jcarlosmontoya Remove Copyright from all `__init__.py` file

website_seo_redirection/__manifest__.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+# © 2016 Jairo Llopis <jairo.llopis@tecnativa.com>
+# © 2018 Juan Carlos Montoya <jcarlosmontoyach@gmail.com>
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+{
+    "name": "Website SEO Redirection",
+    "summary": "Redirect any controller to the URL of your dreams",
+    "version": "11.0.1.0.0",
+    "category": "Website",
+    "website": "https://tecnativa.com/",
```
nikul-serpentcs, line 10: @jcarlosmontoya Change URL as per [this](https://github.com/OCA/maintainer-tools/blob/master/template/module/__manifest__.py#L10)

nikul-serpentcs (commented):
Small Change

pedrobaeza:
Is this still needed with the new Odoo core features? I have migrated without this module without losing the features my customers were using.

rafaelbn:
Hi @jcarlosmontoya , hi all !

As you have commented this is already in the core



So IMHO opinion we shouldn't migrate this module.

@Yajo as original contributor what do you think?

Thanks

yajo:
As other have explained above, this addon is no more needed in v11.

Besides, apart of the advanced page that @rafaelbn showed above, there is a nice UI for setting redirections as seen here:

captura de pantalla de 2018-10-08 09-22-15

eLBati:
@simahawk @tarteo @Yajo @rafaelbn 
I think odoo core does not cover every use case covered by this module.
For example:

> #. Go to *Website Admin > SEO Redirections > Create*.
#. Set ``/blog/our-news-1/post/amazing-post-23`` as *Original URL*.
#. Set ``/amazing`` as *Destination URL*.
#. Enable *Relocate controller*.

> Now navigate to any of both URLs, and you will get to the blog post, but with
the URL you wanted.

Am I wrong?

yajo:
Possibly there's some case that's not covered, but it should be probably analyzed if it's really needed, and reimplemented as an extension to the new upstream system in case it is.

PR #468, opened 2018-05-31
[11.0][ADD] website_img_dimensions

tarteo:
Simple module that shows the width and height (including original width and height) of images on the website when hovered upon.

Can e.g. be useful when you're optimizing your website.


pedrobaeza:
Please avoid plurals on the module name. An screenshot would be good to have it.

tarteo:
@pedrobaeza Thanks for your review. I've changed the name to website_img_dimension. And added a screenshot to the rst.

website_img_dimension/static/src/js/website_img_dimension.js
```diff
@@ -0,0 +1,40 @@
+odoo.define('website_img_dimension', function(require) {
```
simahawk: website_img_dimension.tooltip

website_img_dimension/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,5 @@
+Simple module that shows the width and height (including original width and height) of images on the website
```
simahawk: can you elaborate on the use case?
AFAIU it works only when the editor is loaded: you might want to mention this ;)

website_img_dimension/static/src/js/website_img_dimension.js
```diff
@@ -0,0 +1,40 @@
...
+    "use strict";
+    var Class = require('web.Class');
+    var web_editor_base = require('web_editor.base');
+
+    var TooltipManager = Class.extend({
+        init: function() {
+            var self = this;
+            web_editor_base.ready().done(function() {
```
simahawk: shouldn't you wrap everything inside this?

simahawk (changes requested):

tarteo:
@simahawk Thanks for the review :)! I improved the description and changed the name of the js module.

website_img_dimension/static/src/js/website_img_dimension.js
tarteo: If I wrap the thing in it I wouldn't be able to expose it.

PR #474, opened 2018-06-05
[IMP] Set the footer links to be customizeable

wpichler:
with default active=False. Many installations do have a custom footer, so install will fail on it.

### DIFF
--- a/website_legal_page/README.rst
+++ b/website_legal_page/README.rst
@@ -15,7 +15,9 @@ This module was written to provide common legal pages needed in any website:
 Usage
 =====
 
-Website editor can change easily any text of these pages using website builder
+Website editor can change easily any text of these pages using website builder.
+After installation you have to enable the links in the footer using the customize menu.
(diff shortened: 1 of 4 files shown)


pedrobaeza:
I don't see this change as tolerable in a stable version, as now all the people with this module installed will lose its footers.

wpichler:
What if we set it into a `<data noupdate="1">` node ? So existing installations won't be affected

pedrobaeza:
But I don't get the problem. Inheritance has nothing to do with custom footer. If you have custom footer, it simply replaces the full structure, so no error.

wpichler:
I agree - if the customization was done the right way - then it is no problem. But the world out there is not as perfect as we want it - so there are many installations out there where it will fail - because some "expert" has done the footer modification...

pedrobaeza:
Then you can see that this is not acceptable. The same happens with other modules in Odoo core: if you modify a base view with website editor that changes something that other modules inherit later, then there are chances that this will fail.

The solution is to change the noupdate flag of that view, update the base module (in this case, `website`), and then install.

Closing this.

PR #475, opened 2018-06-05
[TRA] Added german translation

wpichler:


pedrobaeza:
Please use Transifex for these things: https://odoo-community.org/page/translate

wpichler:
Thanks

PR #477, opened 2018-06-06
[11.0][ADD] website_multi

tarteo:
This module makes it easier to manage multiple websites in one database.

Features
--------

* Adds configuration menu for websites
* Improves the list view of websites with extra fields
* Easy switch between websites
* Change the website of pages in the page properties dialog

The module still needs more features. For our customer the implemented features were adequate.

### DIFF
--- a/website_multi/menuitems.xml
+++ b/website_multi/menuitems.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2018 Onestein
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+    <menuitem id="menu_configure_multi_website"
+        parent="website.menu_website_global_configuration"
+        action="action_configure_multi_website"/>
+
+    <menuitem id="menu_configure_menu_website"
+        parent="website.menu_website_global_configuration"
+        action="action_configure_menu_website"/>
+</odoo>
--- a/website_multi/models/website.py
+++ b/website_multi/models/website.py
@@ -7,7 +7,8 @@ from odoo import models, fields
 class Website(models.Model):
     _inherit = 'website'
 
-    # Replaces the web.base.url system parameter to generate e.g. the sitemap.xml and rss feed properly
+    # Replaces the web.base.url system parameter to generate e.g.
+    # the sitemap.xml and rss feed properly
     base_url = fields.Char(
         string='Base Url'
     )
--- a/website_multi/models/website.py
+++ b/website_multi/models/website.py
@@ -1,0 +1,13 @@
+# Copyright 2018 Onestein
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields, api
+
+
+class Website(models.Model):
+    _inherit = 'website'
+
+    # Replaces the web.base.url system parameter
+    base_url = fields.Char(
+        string='Base Url'
+    )
--- a/website_multi/models/website.py
+++ b/website_multi/models/website.py
@@ -7,7 +7,7 @@ from odoo import models, fields
 class Website(models.Model):
     _inherit = 'website'
 
-    # Replaces the web.base.url system parameter
+    # Replaces the web.base.url system parameter to generate e.g. the sitemap.xml and rss feed properly
     base_url = fields.Char(
         string='Base Url'
     )
--- a/website_multi/models/website.py
+++ b/website_multi/models/website.py
@@ -1,7 +1,7 @@
 # Copyright 2018 Onestein
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import models, fields, api
+from odoo import models, fields
 
 
 class Website(models.Model):
--- a/website_multi/static/src/js/website_multi.editor.js
+++ b/website_multi/static/src/js/website_multi.editor.js
@@ -58,7 +58,7 @@ odoo.define('website_multi.editor', function(require) {
         save: function() {
             var website_ids = [];
             this.$('#website_ids input[type="checkbox"]:checked').each(function() {
-                website_ids.push(parseInt($(this).attr('data-id')));
+                website_ids.push(parseInt($(this).attr('data-id'), 10));
             });
 
             this._rpc({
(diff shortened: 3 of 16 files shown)


website_multi/menuitems.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2018 Onestein
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+    <menuitem id="menu_configure_multi_website"
+        parent="website.menu_website_global_configuration"
+        action="action_configure_multi_website"/>
+
+    <menuitem id="menu_configure_menu_website"
```
simahawk, line 10: same id same action?

website_multi/models/website.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2018 Onestein
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields
+
+
+class Website(models.Model):
+    _inherit = 'website'
+
+    # Replaces the web.base.url system parameter
```
simahawk: can you explain?

website_multi/static/src/js/website_multi.editor.js
```diff
@@ -0,0 +1,101 @@
...
+            });
+
+            return this._super.apply(this, arguments);
+        }
+    });
+
+    var SwitchWebsiteMenu = WebsiteNavbarActionWidget.extend({
+        start: function () {
+            var $current_website = this.$('.current-website');
+            $current_website.html(location.hostname);
```
simahawk: location coming from window?

website_multi/static/src/js/website_multi.editor.js
```diff
@@ -0,0 +1,101 @@
...
+
+            var $dropdown_menu = this.$('.dropdown-menu');
+            rpc.query({
+                model: 'website',
+                method: 'search_read',
+                fields: ['domain']
+            }).then(function (websites) {
+                _.each(websites, function(website) {
+                    var url = location.port ? 'http://' + website.domain + ':' + location.port : 'http://' + website.domain;
```
simahawk: why location info is used for each website?

simahawk (changes requested):

website_multi/models/website.py
tarteo: For example to use in generating the sitemap.xml

website_multi/static/src/js/website_multi.editor.js
tarteo: Yes.

website_multi/static/src/js/website_multi.editor.js
tarteo: I'll change it. I can use base_url here it must been a leftover before I had the field.

website_multi/menuitems.xml
tarteo: Can you explain this comment a little more, please?

website_multi/models/website.py
simahawk: but why replace? Shouldn't you reuse it instead or use it as a default?
Is it just for the switcher?

website_multi/menuitems.xml
simahawk: ouch, my eyes got tricked by multi/menu switch... nevermind! :sweat_smile:

website_multi/models/website.py
tarteo: For now the switcher only uses this field. But I have another module website_multi_blog (no PR yet) that uses this field to generate the RSS for multiple website properly. I can still use the web.base.url property as default in that module.

yajo (commented):
I'm just wondering if this duplicates some features from https://www.odoo.com/apps/modules/11.0/website_multi_theme/ :thinking: 

Did you know about it?

tarteo:
@Yajo Thanks for the reaction. Yes I know about the module and some are duplicate. If you want I can make some small changes to website_multi_theme in this PR e.g. the menu item 'Websites' is duplicate. website_multi_theme will then depend on this module. What do you think? 
We can later extend the multi website functionality by adding options for the footer (automatic / custom) etc to be website specific. In a live stream of Odoo I heard they are also working on multi website functionalities.

yajo:
> If you want I can make some small changes to website_multi_theme in this PR e.g. the menu item 'Websites' is duplicate. website_multi_theme will then depend on this module. What do you think?

OK, it would be nice. @yelizariev will be interested in this too.



> We can later extend the multi website functionality by adding options for the footer (automatic / custom) etc to be website specific.

Right now, the website-specific views is, basically, the main feature for `website_multi_theme`. Just for you to know it. :blush: 

> In a live stream of Odoo I heard they are also working on multi website functionalities.

True, but I wouldn't count on it for odoo 12.

Thanks!

yelizariev:
``website_multi_theme`` is useful even when you use the same theme for all websites. I.e. is not just "multi-theme" feature. 

> If you want I can make some small changes to website_multi_theme in this PR e.g. the menu item 'Websites' is duplicate. website_multi_theme will then depend on this module. What do you think?

I'd prefer to see all updates directly in ``website_multi_theme``. I don't think that multi-website is complete without multi-theme (multi-view) features. While making them separate force us to test ``website_multi_theme`` separately on every update of suggested ``website_multi`` module.

@tarteo be aware of our multi-website modules to don't reinvent the same features 
https://github.com/it-projects-llc/website-addons/tree/10.0/website_multi_company
https://github.com/it-projects-llc/website-addons/tree/10.0/website_multi_company_blog
https://github.com/it-projects-llc/website-addons/tree/10.0/website_multi_company_portal
https://github.com/it-projects-llc/website-addons/tree/10.0/website_multi_company_sale

tarteo:
@Yajo Thanks! didn't know the website_multi_theme module has such great features :astonished:  .@yelizariev What do you suggest? I can put the functionalities that I've added in this module into the website_multi_theme module that includes:

- Improve the list view of websites with extra fields (website_ids)
- Easy switch between websites
- Change the website of pages in the page properties dialog

The improvements are mostly nice controls for the user. This module was implemented at a customer that wanted a website with multiple domains. You can see the result on www.ubports.com and www.ubuntu-touch.io. it's the same database using the same theme but with different content.

tarteo:
@yelizariev I've also created a module website_multi_blog which takes care of the sitemap.xml and proper base_url. I can suggest this with a PR extending the module website_multi_company_blog. Thank you a lot for the references :+1:

tarteo:
https://github.com/OCA/website/pull/488

PR #478, opened 2018-06-08
[IMP] website_cookie_notice: append cookie notice inside header

bishalgit:
If the cookie_notice is used with [this theme](https://github.com/OCA/website-themes/tree/11.0/website_theme_flexible) with sticky enabled then content of cookie gets hidden inside the fixed navbar and pushes the underlying navbar down. This commit fixes the issue.

### DIFF
--- a/website_cookie_notice/templates/website.xml
+++ b/website_cookie_notice/templates/website.xml
@@ -25,7 +25,7 @@
 </template>
 
 <template id="cookiebanner" inherit_id="website.layout">
-    <xpath expr="//header" position="before">
+    <xpath expr="//header" position="inside">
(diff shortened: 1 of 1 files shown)


pedrobaeza:
I'm not sure if putting inside respect all the format. @eLBati what do you think?

yajo:
This change is possibly gonna kill many layouts out there.

eLBati:
Even in odoo standard the result is not what expected:

image


@bishalgit I think the solution is to create an auto_install module that depends on both ```website_cookie_notice``` and ```website_theme_flexible``` and fixes the issue

PR #484, opened 2018-06-18
[FIX] subsections in README

sbidoul:
@simahawk 

Subsections are not allowed in README fragments. If you really want one, only one level is supported, using tilde as underline.

Also be careful with images: an image path relative to a fragment will not work relative to the generated README.rst in the root directory.

### DIFF
--- a/website_addthis/readme/CONFIGURE.rst
+++ b/website_addthis/readme/CONFIGURE.rst
@@ -3,7 +3,7 @@
 #. set your "Public ID" (find it in your profile on addthis.com)
 
 Advanced Settings
------------------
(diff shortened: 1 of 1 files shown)


simahawk:
Tests are broken by website_multi_theme. Fast tracking this.

simahawk:
@sbidoul tnx for the info. Regarding images: proper place is?

sbidoul:
> Regarding images: proper place is?

The only good solution that works everywhere (apps.odoo.com, pypi, etc) is to use some publicly accessible URI. There is an example in mis-builder. That's the least crappy solution we have for now.

-sbi

PR #488, opened 2018-06-20
[11.0][ADD] website_multi_theme user controls

tarteo:
* Easy switch between websites
* Change the website of pages in the page properties dialog
* Add website_ids to website.page tree


yelizariev:
@Railia review this please

Railia (approved):

website_multi_theme/__manifest__.py
```diff
@@ -9,7 +9,8 @@
     "version": "11.0.1.4.0",
```
astirpe: You should update the version.

website_multi_theme/models/website.py
```diff
@@ -36,6 +36,10 @@ class Website(models.Model):
         help="Views generated by the multiwebsite theme just for this website",
     )
 
+    base_url = fields.Char(
+        string='Base Url'
```
astirpe: No need to specify the string. Simply writing `base_url = fields.Char()` will behave exactly the same.

website_multi_theme/views/website_page_view.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2018 Onestein
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+    <record model="ir.ui.view" id="website_pages_tree_view">
+        <field name="name">website_pages_tree_view</field>
```
astirpe: No need to specify the name here since you are inheriting from another view.

website_multi_theme/views/website_view.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2018 Onestein
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+
+<odoo>
+    <record model="ir.ui.view" id="view_website_form">
+        <field name="name">website.form</field>
```
astirpe: No need to specify the name here since you are inheriting from another view.

astirpe (changes requested):

astirpe (approved):

astirpe:
Could you resolve the conflict?

tarteo:
Can this be merged, please?

pedrobaeza:
Can you expand README with the explanation of the new features?

pedrobaeza:
Seeing the readme of the new features, I feel this is out of the scope of the module, as this module is not for multi-websites, but for multiple themes in multi-websites. @yelizariev what do you think?

tarteo:
@pedrobaeza see https://github.com/OCA/website/pull/477#issuecomment-397344330

pedrobaeza:
Agh, it's a pity not seeing that on that moment. I think just the contrary, and more in v12 when you will have full multi-website support in core, where these tools can be useful extracted from all of this, but OK, I'm not going to block the PR now in this advanced state.

website_multi_theme/static/src/xml/switch_website_menu.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2018 Onestein
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+
+<templates>
+
+    <t t-name="website_multi_theme.SwitchWebsiteMenu">
+        <li><a href="#" t-att-data-href="website.base_url or ''"><t t-esc="website.domain"/></a></li>
```
yelizariev, line 8: Why don't use ``website.domain`` here? @tarteo

yelizariev (commented):
I didn't check it carefully in time, but there is another issue about this PR

website_multi_theme/static/src/xml/switch_website_menu.xml
tarteo: I created a new field because field `domain` doesn't contain the protocol (http / https). See discussion here: https://github.com/OCA/website/pull/477#discussion_r194661289

yelizariev: You can use ``href="//youwebsite.example"`` -- it will use the same protocol

tarteo: Yes but from a http website to https or the other way around can cause problems if the webserver is not correctly configured.

yelizariev: True, but too specific case IMO.

Issue #487, opened 2018-06-20
Random failures on travis tests with website_multi_theme


yajo:
It was reported in https://github.com/OCA/website/pull/450#issuecomment-387215721, but it seems helpful to have a place where to track this issue. 

<details><summary>Relevant logs</summary>

```
2018-06-19 18:58:25,406 4809 INFO openerp_test odoo.modules.module: odoo.addons.website_multi_theme.tests.test_website_editor running tests.
2018-06-19 18:58:25,406 4809 INFO openerp_test odoo.addons.website_multi_theme.tests.test_website_editor: test_snippets (odoo.addons.website_multi_theme.tests.test_website_editor.WebsiteEditor)
2018-06-19 18:58:25,407 4809 INFO openerp_test odoo.addons.website_multi_theme.tests.test_website_editor: ` Check that there is no error after duplicating snippets.
2018-06-19 18:58:25,836 4809 INFO openerp_test odoo.addons.website_multi_theme.models.website: Updated multi website theme views for Website localhost: ir.ui.view(519, 520, 537, 515, 517, 516, 518, 522, 521)
2018-06-19 18:58:26,494 4809 INFO openerp_test odoo.addons.website_multi_theme.models.website: Updated multi website theme views for Website 0.0.0.0: ir.ui.view(527, 528, 542, 523, 525, 524, 526, 530, 529)
2018-06-19 18:58:26,494 4809 INFO openerp_test odoo.addons.website_multi_theme.wizards.website_config_settings: Reloading available multi-website themes
2018-06-19 18:58:26,793 4809 INFO openerp_test odoo.addons.website_multi_theme.models.website: Updated multi website theme views for Website localhost: ir.ui.view(519, 520, 537, 515, 517, 516, 518, 522, 521)
2018-06-19 18:58:26,939 4809 INFO openerp_test odoo.addons.website_multi_theme.models.website: Updated multi website theme views for Website 0.0.0.0: ir.ui.view(527, 528, 542, 523, 525, 524, 526, 530, 529)
2018-06-19 18:58:27,229 4809 INFO openerp_test odoo.addons.base.res.res_users: Login successful for db:openerp_test login:admin from n/a
[...]

PR #489, opened 2018-06-20
[11.0][ADD] website_img_dimension

tarteo:
Simple module that (if you're an editor) shows the width and height (including original width and height) of
images on the website.

This module can be useful when you're e.g. optimizing your website.

image


tarteo:
The error doesn't seem to be related to this module.

simahawk (approved):

simahawk:
error is related to https://github.com/OCA/website/issues/487

website_img_dimension/static/src/js/website_img_dimension.tooltip.js
```diff
@@ -0,0 +1,40 @@
...
+            }).tooltip('show');
+        },
+        hide_tooltip: function($img) {
+            $img.tooltip('destroy');
+        }
+    });
+
+    var tooltip_manager = new TooltipManager();
+
+    return {
```
Daemo00, line 36: Just for info, why are you returning both an instance and the Class itself?

Daemo00 (changes requested):
Only code review, good in my opinion.

You should generate the README using https://github.com/OCA/maintainer-tools/blob/master/tools/gen_addon_readme.py

website_img_dimension/static/src/js/website_img_dimension.tooltip.js
tarteo: I return the class so other modules can extend the TooltipManager, and I return the instance so other modules can invoke the functions or access variables of it. As TooltipManager is normally only created once like web.WebClient / web.web_client

tarteo:
@Daemo00 I've generated the README. :+1:

tarteo:
@Daemo00 Can you change your review, please?

Daemo00 (approved):
Ok for me

tarteo:
Can this be merged? The travis error is not related to this module https://github.com/OCA/website/issues/487

simahawk:
@tarteo there you go :)

PR #394, opened 2017-10-16
[add] website_menu_permission

simahawk:
Replace or make https://github.com/OCA/website/pull/324 dependent on this.

**TODO**

- [ ] finish README
- [ ] move tests setup to class setup
- [ ] move shortcut flags to website_menu_by_user_status???
- [ ] not required but nice to have: advanced perm form

PR #491, opened 2018-07-12
[11.0][IMP] website_menu_by_user_status: allow menus to be visible to specific user groups

ddufresne:


website_menu_by_user_status/models/website_menu.py
```diff
@@ -38,3 +36,26 @@ def _compute_visible(self):
             self.is_visible = self.user_not_logged
         else:
             self.is_visible = self.user_logged
+
+
+class WebsiteMenuVisibleForSpecificGroups(models.Model):
```
simahawk, line 41: why double inheritance?

website_menu_by_user_status/models/website_menu.py
```diff
@@ -38,3 +36,26 @@ def _compute_visible(self):
...
+    """Optionally display the menu item for a given list of user groups."""
+
+    _inherit = 'website.menu'
+
+    group_ids = fields.Many2many(
```
simahawk, line 46: you might want to have a look at this https://github.com/OCA/website/pull/394 as I proposed this improvement a long time ago :)

simahawk (changes requested):

website_menu_by_user_status/models/website_menu.py
ddufresne: To clearly separate the code of 2 features. Remove one class, the code still works with the other.

website_menu_by_user_status/models/website_menu.py
ddufresne: @simahawk seems like you forgot some credentials in your module https://github.com/OCA/website/pull/394/files.

You barely copied website_menu_by_user_status, changed the name and added the feature.

I prefer augmenting the original one.

simahawk: are you kidding? read the full thread from the original PR :)
The name change was due to be used as a base module for "website_menu_by_user_status".
Also, I was using record rules which is much better as you'll have the same result no matter where you retrieve menu items...

website_menu_by_user_status/models/website_menu.py
simahawk: you don't need it. Why should you "delete a class"?
Unless you define a reusable mixin there's no reason to do it.

website_menu_by_user_status/models/website_menu.py
ddufresne: @simahawk oups, right, did not read the first comment ;)

The purpose of website_menu_by_user_status is not about security. It is only to show or hide menu entries. But this was never mentioned in the module's README.

In my opinion, security should be applied on the content (blog post, web pages, etc) rather than menu entries.

website_menu_by_user_status/models/website_menu.py
ddufresne: This is not the point. It's about grouping only one feature inside a class and not mixing two concepts that work in parallel.

rafaelbn:
Please @cristinamartinrod could you test this one? Thanks a lot!

rafaelbn:
Wow! This module comes from v8! And I've never tested!

This module is already in v11 but @GillesTephaneMeyomesse in https://github.com/OCA/website/pull/324 is trying to migrate from v8 to v10 .

Rebuilding runbot for testing

cristinamartinrod (approved):
Functionally, it works fine for me 👍🏼

website_menu_by_user_status/README.rst
```diff
@@ -18,8 +20,9 @@ To use this module, you need to:
 #. Go to Website > Configuration > Settings > Pages.
```
cristinamartinrod, line 20: Remove "Settings":
#. Go to Website > Configuration > Pages.

cristinamartinrod (changes requested):

website_menu_by_user_status/README.rst
GillesTephaneMeyomesse: @ddufresne It might be interesting to see the new way to make a read file of the OCA https://github.com/OCA/maintainer-tools/tree/master/template/module

ddufresne: @cristinamartinrod @GillesTephaneMeyomesse what's wrong with the README? Did you actually see the rendered file or did you just check the raw file ?

Also, when I submitted this PR, it respected OCA's way of making a readme. Now it does not ?

simahawk: @ddufresne what's wrong w/ you?

Please be polite and double check what YOU are reading and reviewing.

"It might be interesting" does not sound like an order!

This [is not the 1st time](https://github.com/OCA/website/pull/491#discussion_r202322975) and is not really nice...

Any reviewer here and on other PRs is here to help to improve the code for OCA not for stressing you.

So, take a breath, have a beer and start from scratch w/ a different mood pls :sunglasses:

ddufresne: @simahawk ok you are right. Sorry for this issue.

GillesTephaneMeyomesse: I agree with you, all experienced reviewers such as @simahawk are here to improve the community code. Don't make it a personal matter when there is a note on your pull request ;)

ddufresne:
I am closing this PR. Sorry again for my comments and thanks to reviewers. I think this module should be remodeled as suggested by @simahawk . In its current state, it does not seem relevant (for being maintained on OCA).

PR #496, opened 2018-08-08
[FIX] website_multi_theme: Customer -> HTML/CSS editor didn't have all views

yelizariev:

PR #499, opened 2018-08-14
[FIX] duplicates in Customized Views menu

yelizariev:
My colleagues found that my previous PR https://github.com/OCA/website/pull/496 introduced the problem

image

### DIFF
--- a/website_multi_theme/__manifest__.py
+++ b/website_multi_theme/__manifest__.py
@@ -6,7 +6,7 @@
 {
     "name": "Website Multi Theme",
     "summary": "Support different theme per website",
-    "version": "11.0.1.4.1",
+    "version": "11.0.1.4.2",
     "category": "Website",
     "website": "https://www.tecnativa.com",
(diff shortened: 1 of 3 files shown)


yelizariev:
@Yajo you probably missed the previous PR, so you have a chance to review this one :wink:

yajo (approved):

pedrobaeza:
@yelizariev fortunately we won't need this on v12. Have you seen the commits merged yesterday (or my tweet about that)?

pedrobaeza (approved):

yelizariev:
I saw the [tweet](https://github.com/odoo/odoo/commit/4f6ec1cd2a9253bb0c6c17d304490c7809facf8a), but didn't realised that there are two commits and checked only one with a preview... Another [commit](https://github.com/odoo/odoo/commit/4f6ec1cd2a9253bb0c6c17d304490c7809facf8a) actually looks like a general approach... I think you are right, multi-theme will not be needed in 12

PR #514, opened 2018-09-19
[ADD] website_form_firstname

trojikman:
This module adds the compatibility of partner_firstname module with website_sale,
i.e. allows you to use fields first name and last name in the form.

### DIFF
--- a/website_form_firstname/README.rst
+++ b/website_form_firstname/README.rst
@@ -1,0 +1,21 @@
+.. image:: https://img.shields.io/badge/license-LGPL--3-blue.png
+   :target: https://www.gnu.org/licenses/lgpl
+   :alt: License: LGPL-3
+
+======================================
+ Website form first name and last name
+======================================
+
+This module adds the compatibility of partner_firstname module with website_sale,
+i.e. allows you to use fields first name and last name in the form.
+
+Credits
+=======
+
+Contributors
+------------
+* Denis Mudarisov <mudarisov@it-projects.info>
+
+Sponsors
+--------
+* * `e-thos SSII <http://www.e-thos.fr/>`__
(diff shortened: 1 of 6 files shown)


pedrobaeza:
According description, this should be in OCA/e-commerce, and be called `website_sale_firstname`. Please also sign the CLA following http://odoo-community.org/page/cla (the other link is incorrect).

yelizariev:
@trojikman can the module depend on website_form only?

yelizariev:
@pedrobaeza the module has import 

``from odoo.addons.website_sale.controllers.main import WebsiteSale``

but it doesn't require ``website_sale`` be in the dependency list. Can we leave just ``website_form`` as dependency and keep this pull request in oca/website rather than send it to oca/e-commerce?

pedrobaeza:
If you make an import of that module, it should be in the dependency list.

yelizariev:
Is it some policy? Because there is no technical reason for that: some class is updated, but it's not used if website_sale is not installed

pedrobaeza:
Well, it's a nonsense to put something that is not going to be used, and controllers are an anomaly over the rest of the ORM, that allows this kind of "Python tricks", but I don't think the rule should change with them. I don't understand any way why not depending on `website_sale`, and to keep it here. It's clearly an e-commerce feature, and it uses `website_sale` features.

yelizariev:
It actually fixes two modules: website_form and website_sale. You can use this module for example in website_crm and don't install website_sale

pedrobaeza:
Then you have to create 2 modules: one for the one adding the field for website_form, and another for website_sale part IMO. Right now, you are patching all the DBs making not mandatory the name when this is installed. Am I wrong?

yelizariev:
I'm not sure about your question. It depends on whether odoo loads not installed modules to registry or not.

PS my statement about website CRM probably wrong as CRM.lead is created there rather than partner

yelizariev:
I think there is still theoretical possibility to use this module without website_sale, but I agree to rename it and move to OCA/e-commerce

PR #521, opened 2018-10-04
[ADD] Added new module website_seo_product

wpichler:
Does integrate website_seo_redirect directly with products


website_seo_product/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+    # see https://odoo-community.org/page/development-status
+    "development_status": "Production/Stable",
+    "category": "Website",
+    "website": "https://github.com/OCA/website",
+    "author": "Wolfgang Pichler <wpichler@callino.at>, Odoo Community Association (OCA)",
+    # see https://odoo-community.org/page/maintainer-role for a description of the maintainer role and responsibilities
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "external_dependencies": {
```
nikul-serpentcs: @wpichler If not any `external_dependencies` then remove a key.

website_seo_product/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+        "python": [],
+        "bin": [],
+    },
+    "depends": [
+        "website_seo_redirection", "website_sale"
+    ],
+    "data": [
+        "views/product.xml",
+    ],
+    "demo": [
```
nikul-serpentcs: same here

website_seo_product/__manifest__.py
```diff
@@ -0,0 +1,31 @@
...
+    ],
+    "qweb": [
```
nikul-serpentcs: same here

website_seo_product/models/product.py
```diff
@@ -0,0 +1,48 @@
...
+        if not seo.startswith('/'):
+            seo = "/%s" & seo
+        seo = seo.lower()
+        re.sub(r"[^\w-\d\/]", "", seo)
+        self.seo = seo
+
+    @api.depends('seo_redirect_id')
+    def _compute_seo_path(self):
+        for product in self:
+            product.seo_path = product.seo_redirect_id.destination if product.seo_redirect_id else product.website_url
```
nikul-serpentcs: It's too long line

website_seo_product/models/product.py
```diff
@@ -0,0 +1,48 @@
...
+            })
+
+    seo_redirect_id = fields.Many2one('website.seo.redirection', string='SEO Redirect', ondelete='set null')
+    seo = fields.Char('SEO Path', compute='_compute_seo_path', inverse='_set_seo_path', store=True)
+
+    @api.multi
+    def _compute_website_url(self):
+        super(ProductTemplate, self)._compute_website_url()
+        for product in self.filtered(lambda p: p.seo_redirect_id is not None):
+            product.website_url = product.seo
```
nikul-serpentcs: Add new Blank line

website_seo_product/views/product.xml
```diff
@@ -0,0 +1,20 @@
...
+                <!-- add state field in header -->
+                <group name="website"  position="inside">
+                    <field name="seo_redirect_id" invisible="1"/>
+                    <field name="seo"/>
+                </group>
+            </field>
+        </record>
+
+    </data>
+</odoo>
```
nikul-serpentcs: Add new Blank line

website_seo_product/views/product.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2018 Wolfgang Pichler
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <data>
```
nikul-serpentcs: Remove it

nikul-serpentcs (commented):
Improve Code

wpichler:
all done.
I am curios what is happening all the time with my newlines - pycharm seems to remove the last newline at every file - i do have to add 2 newlines - to get 1 newline at file end on real file...

wpichler:
I've reworked the functionality a little bit - module does now support renaming of product without losing the seo url !

rafaelbn:
Hello @wpichler !

Thanks!.

Have you read https://www.odoo.com/documentation/user/10.0/website/optimize/seo.html ?
I'm not sure in v12 hace improvements : https://www.odoo.com/documentation/user/12.0/website/optimize/seo.html ?

Note than in Odoo v12 you can rename URLs.

I set WIP until runbot :red_circle: and travis :red_circle: gets :green_heart: 

Ping @cristinamartinrod @Yajo what do you think?

yajo (changes requested):
Yikes! I'm very sorry to say this... :sweat_smile: but I think this module is wrong by design: `website_seo_redirection` uses 301 redirections by default. A 301 redirection is correctly handled by web crawlers (by indexing only the target url) and browsers (by caching the redirection and not asking for it more than once), so basically you shouldn't care at all about what this module is caring about.

So, IMHO, this addon adds no value to current implementation, although it adds complexity and reduces maintainability. We shouldn't merge this. :roll_eyes:

cristinamartinrod (changes requested):
I think that this module is necessary in v10 but it can has complications if any ecommerce wants to migrate to v12. Anyway in this case, only changes the URL visible in the browser but keeps canonical URL, so if you want to have "pretty" url's for the user it's ok, but if you want a new url more seo-friendly, then the canonical (the only one that crawls google) should be modified.

wpichler:
I do not agree here. First of all - this module is not about using 301 vs using internal redirect. This module does just help the user when the user does change the product name - to update the redirect rule according to the name change.

Behaviour without this module:
There is a Product named "Apple" with ID 3 - so path to product is /shop/product/Apple-3. User does now create a seo redirect rule (using 301 or internal does not matter here). So that /fruit/apple does point to /shop/product/Apple-3.
Now the user decides to rename the product to "Organic Apple" - because of this the url will change to /shop/product/OrganicApple-3.
So the existing redirect rule will not work anymore.

Behaviour with this module:
If the product name changes - it will update the seo rule to the new url.

wpichler:
And about 301 vs internal redirect - as far as i know (correct me when i am wrong) - you will get better ranking when not using 301.

yajo:
@wpichler:

>  If the product name changes - it will update the seo rule to the new url.

AFAICS the current code in this PR is buggy and will not do that, if that was indeed the intention, because the search for origin URL is made after the actual product write, which means it will never find a result.

> Now the user decides to rename the product to "Organic Apple" - because of this the url will change to /shop/product/OrganicApple-3.
> So the existing redirect rule will not work anymore.

AFAIK the redirect rule will still work, just with 1 extra step (automatic and transparent to the user/crawler), because Odoo uses in reality only the "3" part at the end; the rest of the slug is just aesthetic, and will be autoredirected by odoo if wrong, as you see in core v10 runbot:

peek 08-10-2018 11-47


> as far as i know (correct me when i am wrong) - you will get better ranking when not using 301

See the [Google support page on the matter](https://support.google.com/webmasters/answer/93633?hl=en). Quoting it:

> If you need to change the URL of a page as it is shown in search engine results, **we recommend** that you use a server-side 301 redirect. **This is the best way** to ensure that users and search engines are directed to the correct page. The 301 status code means that a page has permanently moved to a new location.

So, using 301 redirections is actually a Google recommendation for our use case at hand.

---

@cristinamartinrod:


> if you want a new url more seo-friendly, then the canonical (the only one that crawls google) should be modified

You are absolutely right. We have [`website_canonical_url`](https://github.com/OCA/website/tree/10.0/website_canonical_url), but it does not have any link or compatibility layer with `website_seo_redirection`. Support could be added there directly on a lazy-dependency fashion, or we could just leave it as it is and add support in v11 directly, where redirections are a core feature (although still the canonical URL is a core problem that should be instead handled by Odoo directly and automatically).

---

Summarizing, still I get to the same conclusion as before: this PR is not needed.

On behalf of being practical, I'm gonna close the PR, but please reopen if you still think I'm wrong (which might be true indeed), or if you still want to wait for others' opinions :wink:. Thank you very much for the contribution!

PR #527, opened 2018-10-24
[FIX] website_slides: Add fetch timeout

yajo:
This makes you actually able to install the module with demo data and unplugged from network. Otherwise, it's impossible.

@Tecnativa

### DIFF
--- a/website_slides/__openerp__.py
+++ b/website_slides/__openerp__.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 {
     'name': 'Slides',
-    'version': '8.0.2.0.0',
+    'version': '8.0.2.0.1',
     'license': 'LGPL-3',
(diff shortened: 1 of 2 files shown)


pedrobaeza:
But this will likely happen in future versions in Odoo core, isn't it?

yajo:
Indeed it has the same problem, but it's hitting me here right now...

pedrobaeza:
Well, we are going to migrate in 2 months, so we might find a definitive solution.

yajo:
I opened https://github.com/OCA/OCB/pull/808 and https://github.com/odoo/odoo/pull/28098 to provide the definitive solution, and travis :x: is not related.

rafaelbn (approved):

pedrobaeza:
Main branch is red, but I'm going to merge this old one for having it.

PR #459, opened 2018-05-11
[ADD][11.0] website_product

JordiBForgeFlow:
Website Product
===============

This module extends the functionality of the website to support viewing
a specific product that has been published.

With this module all products can potentially be published to the website.

Usage
=====

* Go to the Product form in the backend and press the button 'Unpublished On Website'.
  You will be redirected to the website, in the form view of that product.

* If the product is sealable the user will display the shopping options.
  Otherwise a simple form is displayed, with no option to purchase.

This module per se does not do more than that, and other modules such as
'cms_product_form' from the website-cms OCA repository make use of this module.

PR #528, opened 2018-10-25
[ADD] 11.0 website_product_sale

HviorForgeFlow:
Glue module for a correctly product URL computing allowing to use website_product #459 and website_sale core module simultaneously.


website_product_sale/models/product.py
```diff
@@ -0,0 +1,25 @@
...
+
+class ProductTemplate(models.Model):
+    _name = 'product.template'
+    _inherit = ['product.template',
+                'website.seo.metadata',
+                'website.published.mixin']
+
+    @api.multi
+    def _compute_website_url(self):
+        # If the product cannot be sold, then use the new form view.
```
simahawk: form view?

website_product_sale/models/product.py
```diff
@@ -0,0 +1,25 @@
...
+        # Otherwise, use the defult URL from website_sale, which will
```
simahawk: default

website_product_sale/models/product.py
```diff
@@ -0,0 +1,25 @@
...
+        # redirect the user to the shop.
+        super(ProductTemplate, self)._compute_website_url()
+        for product in self:
+            if product.sale_ok:
```
simahawk:
```python
if not product.sale_ok:
    product.website_url = "/product/%s" % slug(product)
```

website_product_sale/models/product.py
```diff
@@ -0,0 +1,25 @@
...
+    _inherit = ['product.template',
+                'website.seo.metadata',
+                'website.published.mixin']
+
+    @api.multi
+    def _compute_website_url(self):
+        # If the product cannot be sold, then use the new form view.
+        # Otherwise, use the defult URL from website_sale, which will
+        # redirect the user to the shop.
+        super(ProductTemplate, self)._compute_website_url()
```
simahawk, line 16: I would not call super here: is going to call all the overridden versions of this method included your custom one in website_product, for no benefit.

Plus, `sale_ok` comes from `product` module. So, why can't you do this check here? 
https://github.com/Eficent/website/blob/11.0-website_product/website_product/models/product.py#L21
In there you also say the same about not saleable products... :thinking:

simahawk (changes requested):

website_product_sale/models/product.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2018 Eficent Business and IT Consulting Services S.L.
+#   (http://www.eficent.com)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from odoo import api, fields, models
+from odoo.addons.http_routing.models.ir_http import slug
+
+
+class ProductTemplate(models.Model):
+    _name = 'product.template'
+    _inherit = ['product.template',
```
simahawk: all this inheritance is not needed as you already depend on the module that adds such inheritance, no?

website_product_sale/models/product.py
HviorForgeFlow: I prefer to keep this way, if I remove super I should add /shop/ route manually. Otherwise I can't ensure all url will be computed correctly. But I'm not 100% sure. Tests were failing when super was removed.

rafaelbn:
Please @cristinamartinrod could you review? Thanks!

rafaelbn:
Hello @hveficent ! 

¿ website_product_sale VS "Computes correctly product URLs" ?

I think the name of the module should be different. I thought that you were duplicating the website_sale functionality

That is my doubt

Thank you!

website_product_sale/models/product.py
simahawk: @hveficent pls read my 2nd comment: I don't think you need this override at all.
Isn't sale_ok already present w/out website_sale? Then why do you need this module at all?

Can't you check this flag in website_product?

If this is not always true, then I'd improve website_product to delegate the check to a separated method and then in this new glue module you just override this check and you won't have to touch website_url at all :wink:

chienandalu: I agree with @simahawk I don't see why this can't be made in the other module :slightly_smiling_face:

HviorForgeFlow:
The existence of this module is due to: https://github.com/odoo/odoo/blob/7718811809f1ccdcab93415c1351144061d0c9b6/addons/website_sale/models/product.py#L174
 URL is computed for shop no matter if is ```sale_ok``` or not.
```website_product``` by its own, just compute all URLs to show as a "catalogue", but if ```website_sale``` is installed it overwrite all URL to shop. Only adding this module is the solution I've found to at least do a last overwrite of the function to show URL properly. 

It is not the best solution, so yes, I would rather not to perform this module, but I can't find other solution.

I'll appreciate properly solutions. Thanks for reviewing.

**Comment by simahawk** [comment]
@hveficent read what I [proposed before](https://github.com/OCA/website/pull/528#discussion_r228276613), it should be something like:

in website_product
```python
def _compute_website_url(self):
    for rec in self:
        if rec._is_website_product():  # pick the name that fits best
            rec.website_url = '/product/' + slug(rec)
        else:
            rec.website_url = super(Product, rec)._compute_website_url()

def _is_website_product(self):
    return True
```
in website_product_sale
```python
def _is_website_product(self):
    return not self.sale_ok
```
BTW I wonder if you really want this every time... don't you want ppl to be able to reach the shop page if website_sale is installed?

HviorForgeFlow:
Yes, this is the point.

I've tried your approach and it is not working for me, as soon as first _compute_website_url method usage is from ```website_sale``` one. 

So I think the only solution possible is to call super from this module that depends from the others 2 ones, they do whatever they want, website_sale always will put its URL, then fix for not salable products only.

HviorForgeFlow:
Closing, maybe reopening with different functionalities due to #597

PR #531, opened 2018-10-27
Update index.html

AdvaitT17:
Bottom margin was too small...


pedrobaeza:
Index is auto-generated from README, so you should maybe propose a change on the template instead of this.

simahawk:
@AdvaitT17 that file is generated here https://github.com/OCA/maintainer-tools/blob/master/tools/gen_addon_readme.py#L200 
You should be able to pass a new stylesheet to it.

PR #536, opened 2018-11-14
[11.0][ADD] website_snippet_timeline

tarteo:
This adds a new website snippet that looks like this:

image


website_snippet_timeline/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+{
+    'name': 'Website Block Timeline',
```
simahawk, line 2: snippet? :)

website_snippet_timeline/templates/snippets.xml
```diff
@@ -0,0 +1,103 @@
...
+                 data-js="timeline_item"
+                 data-drop-in=".snippet_timeline ul" data-drop-near=".snippet_timeline ul > li" ></div>
+
+            <div data-js="margin-y"
+                 data-selector=".snippet_timeline_item">
+            </div>
+        </xpath>
+    </template>
+
+    <template id="snippet_timeline_item" name="Timeline - Item">
```
simahawk: prefix module name or "Snippet" to templates' names to ease lookup in the backend

simahawk (changes requested):
really nice... LG! Just minor things...

simahawk:
travis linting is red

website_snippet_timeline/__manifest__.py
tarteo: The user name is block, that's why in the readme and manifest I use block instead of snippet. In the code though I use snippet.

website_snippet_timeline/templates/snippets.xml
tarteo: The problem when I change the name here that the label in the block menu is also changed:

image

tarteo:
@simahawk The travis error seems to be from website_google_tag_manager. I wanted to fix it but the error is pointing to line 3 in the readme which seems fine to me :P So instead I changed some stuff in my readme to hopefully magically fix it ;)

simahawk (approved):

yung-wang (approved):
Really nice, I like the design of the snippet. It works here.

PR #539, opened 2018-11-20
12.0 mig website_cookie_notice

ernestotejeda:
Cc @Tecnativa


ernestotejeda:
This module add a div that has the the bootstrap class 'bg-warning' on the top of the pages in the website, but I do not like how this alert message looks with Boostrap v4, I think this background color does not contrast very well with the font color of the text, besides I do not think it fits very much with the rest of the colors. I don't really have a strong excuse from the graphic design point of view since I am not an expert in the subject, but I propose something and I would like to see opinions.
We could change the class 'bg-warning' for the classes "alert alert-warning" which are the classes that are used to display an alert, but the background color and it seems to me better adjusted.

In v11:
v11

Current:
v12_migration

Proposal:
proposal_v12_migration

chienandalu:
> We could change the class 'bg-warning' for the classes "alert alert-warning" which are the classes that are used to display an alert, but the background color and it seems to me better adjusted.

Sure that's better :smile: :+1:

ghost:
Thank you @ernestotejeda for migrating module.
This colour is much more better :) Maybe a configurable block would be the best solution.

One Idea:
The depency to website_legal_page is from technical side not really necassary. I want to use cookie_notice without legal_pages. There are not useful for me.
Maybe we could remove this depency. Everybody who want to use legal pages module is free to install it.

website_cookie_notice/readme/ROADMAP.rst
```diff
@@ -0,0 +1,16 @@
+* If you are using this module in version < 8.0.2.0.0 and update it, any other
```
chienandalu: If this information relevant to this version?

chienandalu (approved):
Great! Just a minor comment

pedrobaeza:
@Yajo should we change the text to accommodate it to GDPR?

website_cookie_notice/readme/ROADMAP.rst
ernestotejeda: No, I just removed it

yajo:
No need @pedrobaeza, and that goes into `website_legal` anyway.

Recently I was wondering if we should use https://cookieconsent.insites.com/ as an open source solution. :thinking:

website_cookie_notice/templates/website.xml
```diff
@@ -12,10 +12,10 @@
     <div t-if="request and
         not request.session.get('accepted_cookies')"
         id="website_cookie_notice"
-        class="container-fluid bg-warning cc-cookies">
+        class="container-fluid cc-cookies alert alert-warning" role="alert">
         <div class="container">
             <div class="row">
-                <div class="col-sm-10">We use cookies in this website. Read about them in our <a href="/legal/privacy-policy">privacy policy</a>. To disable them, configure your browser properly. If you keep using this website, you are accepting those.</div>
+                <div class="col-sm-10">We use cookies in this website. Read about them in our <a href="/legal/privacy-policy" class="alert-link">privacy policy</a>. To disable them, configure your browser properly. If you keep using this website, you are accepting those.</div>
```
yajo, line 18: Do you think we should link to https://cookiesandyou.com/?

yajo (approved):

yajo:
(Above comments are just ideas, but this is OK to merge as is)

pedrobaeza:
@ernestotejeda @eLBati what do you think about the https://cookieconsent.insites.com/ switch?

eLBati:
@pedrobaeza I think it would be very good

ernestotejeda:
The message of the notification is configurable, so no matter what this module sets by default, the person who configures it can adapt it to their needs, so I think we can remove the dependency.
We could make it clear in the description of the module in the readme that it's the responsibility of who configures the module that the message of the notification complies with the provisions of The EU legislations for data and privacy protection in The General Data Protection Regulation (GDPR) and the ePrivacy Directive.

yajo:
@ernestotejeda are you talking about https://cookieconsent.insites.com/ or about the current implementation of this module?

ernestotejeda:
I'm talking about the current implementation

yajo:
> I think we can remove the dependency

I guess you're talking now about `website_legal_page` then.

IMHO it's ok to leave that dependency as it is. There are quite a few legal addons that all depend on it, it's a good way to have a central legal text that works out of the box.

Of course, if you see the privacy pages and don't click on them, it means that you don't care much about the text. And if you care, it's the 1st thing you're gonna change anyways, so I wouldn't block this any longer.

About cookieconsent library... I have been thinking about it. It could have been a good start, but now that it's done in a way that fits perfectly with odoo system and with bootstrap, maybe changing is going to be worse than leaving it as it is.

So, IMHO, it's ready. :shipit:

eLBati:
Ok for me to keep the current implementation

yajo:
@ernestotejeda can you please squash migration commits, to merge?

PR #540, opened 2018-11-23
[FIX] website_form_recaptcha: Re-recaptcha

chienandalu:
- When website_crm_phone_validation is installed, the captcha validation
is called twice so the form allways fail as with a
'timeout-or-duplicate' error.

cc @Tecnativa

### DIFF
--- a/website_form_recaptcha/controllers/main.py
+++ b/website_form_recaptcha/controllers/main.py
@@ -32,6 +32,15 @@ class WebsiteForm(WebsiteForm):
         res = super(WebsiteForm, self).extract_data(model, values)
         if model.website_form_recaptcha:
             captcha_obj = request.env['website.form.recaptcha']
+            # Only check once: if a call to reCAPTCHA's API is made twice with
+            # the same token, we get a 'timeout-or-duplicate' error. So we
+            # stick to the first response data storing the token after the
+            # first invoke in the current request object. This duplicated
+            # call can be cause for instance by website_crm_phone_validation.
+            try:
+                getattr(request, captcha_obj.RESPONSE_ATTR)
+            except AttributeError:
+                return res
             ip_addr = request.httprequest.environ.get('HTTP_X_FORWARDED_FOR')
             if ip_addr:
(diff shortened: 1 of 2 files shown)


website_form_recaptcha/controllers/main.py
```diff
@@ -32,6 +32,11 @@ def extract_data(self, model, values):
         res = super(WebsiteForm, self).extract_data(model, values)
         if model.website_form_recaptcha:
             captcha_obj = request.env['website.form.recaptcha']
+            # Only check once
+            try:
+                getattr(request, captcha_obj.RESPONSE_ATTR)
+            except AttributeError:
+                return res
```
pedrobaeza, line 43: Why do you return earlier here?

chienandalu: The problem here is that with the reCaptcha API, if you make the same call twice, you get an error. So after the first invoke, I store the response's unique value in the request object. So the successive times this is call we'll return as it's already validated.

pedrobaeza: Why don't you put this as comment in the code then for avoiding wondering this on the future?

pedrobaeza (approved):

yajo (approved):
It looks OK. Just one suggestion: Maybe we should just move the check to the `website_form` method?

chienandalu:
I guess it would be valid, although isn't it better to channel the output with rest of the validations?: https://github.com/odoo/odoo/blob/11.0/addons/website_form/controllers/main.py#L33

chienandalu:
The LINT is failing on another module's readme links: https://travis-ci.org/OCA/website/jobs/459663549#L852

yajo:
I think it's this: https://travis-ci.org/OCA/website/jobs/459663549#L1055

OK about https://github.com/OCA/website/pull/540#issuecomment-442103180.

chienandalu:
I'd say is this one: https://travis-ci.org/OCA/website/jobs/459663549#L852

yajo:
Rebase please.

chienandalu:
There we are :+1:

PR #541, opened 2018-11-26
Fix icon anchor scroll

Ricardlol:
fix plugin in this situation:
```html
<a href="#aplicacio" class="btn btn-default">
    <span class="fa fa-suitcase fa-5x __web-inspector-hide-shortcut__" data-original-title="" title="" style=""></span>
</a>
```

### DIFF
--- a/website_anchor_smooth_scroll/static/src/js/website_anchor_smooth_scroll.js
+++ b/website_anchor_smooth_scroll/static/src/js/website_anchor_smooth_scroll.js
@@ -11,7 +11,7 @@ odoo.define('website_anchor_smooth_scroll', function(require) {
 
     var smooth_scroll = function(event) {
         event.preventDefault();
-        var anchor_fragment = event.target.hash;
+        var anchor_fragment = event.target.hash || this.hash;
 
         // Do this before scrolling so that browser history accurately reflects scroll position at time of click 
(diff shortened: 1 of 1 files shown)


yajo:
Can you explain better on what are you fixing, how to reproduce the error, etc.?

Ricardlol:
Sorry by not answer,
The mistake is that the anchor doesn't work with images, icons, etc, only
work with strings by reproduce the error create a image or icon and add a
anchor
example:
<a href="#example">
    example
</a>

website_anchor_smooth_scroll/static/src/js/website_anchor_smooth_scroll.js
```diff
@@ -11,7 +11,7 @@ odoo.define('website_anchor_smooth_scroll', function(require) {
 
     var smooth_scroll = function(event) {
         event.preventDefault();
-        var anchor_fragment = event.target.hash;
+        var anchor_fragment = event.target.hash || this.hash;
```
yajo, line 14: Change to `event.currentTarget.hash`

yajo (changes requested):

tarteo (changes requested):
Can you increase the version number in the manifest.

PR #548, opened 2018-12-05
[12.0][MIG] website_canonical_url

mfigueiredo-odoogap:
migration for website_canonical_url completed


website_canonical_url/models/res_config.py
```diff
@@ -8,4 +8,4 @@ class WebsiteConfigSettings(models.TransientModel):
 
     _inherit = 'res.config.settings'
 
-    canonical_domain = fields.Char(related='website_id.canonical_domain')
+    canonical_domain = fields.Char(related='website_id.canonical_domain', required=False)
```
simahawk: why do you remove required? You should set `readonly=False` instead since related fields in v12 are readonly by default.

simahawk (changes requested):

website_canonical_url/models/res_config.py
mfigueiredo-odoogap: my bad, you are right, i mistaked the word

pedrobaeza:
Travis and runbot are red. Can you please check them?

mfigueiredo-odoogap:
i'm checking the issues, althoug travis error i cant resolve it. it says flake8 fail. is this because i use pep8?

pedrobaeza:
This is the error:
```
website_canonical_url/models/res_config.py:11:80: E501 line too long (89 > 79 characters)
```

website_canonical_url/models/res_config.py
mfigueiredo-odoogap: have you seen the changes ?

mfigueiredo-odoogap:
@simahawk have you seen the changes, to the problem that you pointed out?

simahawk (approved):

PR #552, opened 2018-12-20
[12.0][MIG] website_google_tag_manager: Migration to v12

cristinamartinrod:
cc @Tecnativa


website_google_tag_manager/views/website_views.xml
```diff
@@ -1,16 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
pedrobaeza: Why removing this?

website_google_tag_manager/__manifest__.py
```diff
@@ -16,10 +16,9 @@
     ],
     "data": [
         "views/res_config_settings_view.xml",
-        "views/website_views.xml",
         'views/website_templates.xml',
     ],
     'images': [
-        'static/description/icon.png',
```
pedrobaeza: It's OK to remove the icon here, but don't remove the file per se, as it's the module icon (it's always automatically got from that path without being here declared).

pedrobaeza (changes requested):
Remove again `.DS_Store` file

website_google_tag_manager/views/website_views.xml
cristinamartinrod: En v12 Odoo remove all atributes (analytics, google maps, etc...) in this form. Now this form is just a popup to create a new website.

pedrobaeza: But then you need other form where putting this. There should be a `website` form...

website_google_tag_manager/__manifest__.py
cristinamartinrod: I don't remove it. Only was changed for new one :)

website_google_tag_manager/views/website_views.xml
cristinamartinrod: Doesn't look like it

website_google_tag_manager/__manifest__.py
pedrobaeza: This is saying the contrary:

https://github.com/OCA/website/pull/552/commits/4bad4da977df42c045b41012435226aa74012a32#diff-d6dd36ca45092f55ef21b2c115018a42

As said, there MUST BE a `static/description/icon.png` file in the module for having a module icon. If you add more images (which this section is for), great, but the other one is different.

cristinamartinrod: Of course and here is the icon file in his branch: https://github.com/Tecnativa/website/tree/12.0-mig-website_google_tag_manager/website_google_tag_manager/static/description

pedrobaeza: It's very weird, as GitHub is saying that is removed:

seleccion_004

but indeed it's OK.

pedrobaeza (approved):

website_google_tag_manager/__manifest__.py
cristinamartinrod: Yeah, is about the image size, now is smaller. I hadn't noticed it before either.

chienandalu (commented):
I couldn't make it work

image

cristinamartinrod:
@chienandalu this is the test on runbot with a real Container ID.
GTM approves the script and collects data:
captura de pantalla 2018-12-20 a las 12 06 38
captura de pantalla 2018-12-20 a las 12 06 30

chienandalu (approved):
Ok, then I just had it misconfigured it :+1:

Issue #554, opened 2019-01-02
[10.0] website_field_autocomplete - data-domain not working


ivantodorovich:
The domain is not being evaluated, so it's being concatenated as a string.
This raises a `ValueError: Invalid leaf`

For instance, this is a domain logged at the controller, right before searching `lef` with custom domain `[['id', 'in', [504, 505]]`:
```py
[[u'name', u'ilike', u'lef'], u"[['id', 'in', [504, 505]]]"]
```

@lasley

PR #555, opened 2019-01-03
[10.0][FIX] add_domain was not being evaluated (#554)

ivantodorovich:
This fixes https://github.com/OCA/website/issues/554

At first, I tried to use `pyeval`, but then realized it's not available on website, only on backend.


website_field_autocomplete/controllers/main.py
```diff
@@ -6,6 +6,8 @@
 from odoo import http
 from odoo.http import request
 from odoo.addons.website.controllers.main import Website
+import logging
```
pedrobaeza: Why the changes in this file?

ivantodorovich: Sorry, that shouldn't be there 👍 
Fixed

simahawk (approved):
LG

PR #556, opened 2019-01-04
[IMP] 11.0 website_form_recaptcha improve JS

mpanarin:
* Make JS easier to extend (for example, when captcha api params are required)
* Make enforce language when possible (auto-detect can fail if browser lang is different)

### DIFF
--- a/website_form_recaptcha/README.rst
+++ b/website_form_recaptcha/README.rst
@@ -25,6 +25,10 @@ To use this module, you need to:
 * Set `website_form_recaptcha` to `True` on that model (similar to enabling forms)
 * Add an element with the `o_website_form_recaptcha` class anywhere in the form
 
+Captcha will try to use the language of your website. If it, for whatever
+reason, can't find, it will default to google API and use the language of the
+browser.
+
 Look at `website_crm_recaptcha` module for example implementation
 
 
@@ -55,6 +59,7 @@ Contributors
 ------------
 
 * Dave Lasley <dave@laslabs.com>
+* Mykhailo Panarin <m.panarin@mobilunity.com>
 
 Maintainer
 ----------
--- a/website_form_recaptcha/static/src/js/field_recaptcha.js
+++ b/website_form_recaptcha/static/src/js/field_recaptcha.js
@@ -11,22 +11,53 @@ odoo.define('website_form_recaptcha.recaptcha', function (require) {
 
     snippet_animation.registry.form_builder_send = form_builder_send.extend({
 
+        // https://developers.google.com/recaptcha/docs/language
+        captcha_languages: [
+            "ar", "af", "am", "hy", "az", "eu", "bn", "bg", "ca", "zh-HK",
+            "zh-CN", "zh-TW", "hr", "cs", "da", "nl", "en-GB", "en", "et",
+            "fil", "fi", "fr", "fr-CA", "gl", "ka", "de", "de-AT", "de-CH",
+            "el", "gu", "iw", "hi", "hu", "is", "id", "it", "ja", "kn", "ko",
+            "lo", "lv", "lt", "ms", "ml", "mr", "mn", "no", "fa", "pl", "pt",
+            "pt-BR", "pt-PT", "ro", "ru", "sr", "si", "sk", "sl", "es", "es-419",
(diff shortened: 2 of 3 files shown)


website_form_recaptcha/static/src/js/field_recaptcha.js
```diff
@@ -15,18 +15,28 @@ odoo.define('website_form_recaptcha.recaptcha', function (require) {
...
             );
         },
+        _get_captcha_elem: function (data) {
+            return $('<div/>', {
+                'class': 'g-recaptcha',
+                'data-sitekey': data.site_key
+            });
+        },
+        _get_captcha_script_url: function (data) {
+            return 'https://www.google.com/recaptcha/api.js';
```
simahawk: I'd set `hl=session.user_context.lang` if available

simahawk (changes requested):
version upgrade missing.
If you add the lang improvement you should also mention it in README

website_form_recaptcha/static/src/js/field_recaptcha.js
mpanarin: I think the default parameters of api should be kept by default. This way, `_get_captcha_script_url` is easy to extend. So, I think its better to keep it as is.

simahawk: well, to me if the language in the captcha is not loaded properly based on the odoo lang... then it's a bug, not a feature.
Unless you show me the contrary... :wink: 
In my case, with the website in FR and my browser in EN I got captcha content in EN, which is wrong IMO.

website_form_recaptcha/static/src/js/field_recaptcha.js
```diff
@@ -11,22 +11,52 @@ odoo.define('website_form_recaptcha.recaptcha', function (require) {
...
+                'class': 'g-recaptcha',
+                'data-sitekey': data.site_key
+            });
+        },
+        _get_captcha_script_url: function (data) {
+            return 'https://www.google.com/recaptcha/api.js' + this._set_additional_params();
+        },
+        _set_additional_params: function() {
+            var lang = $("html").attr("lang").replace("_", "-");
+            if (this.captcha_languages.includes(lang)) {
```
simahawk, line 51: I'd move `'https://www.google.com/recaptcha/api.js'` to a static attribute `recaptcha_js_url` and keep only `_get_captcha_script_url`.

Also, I'd add 2 comments inside the IF blocks: `// lookup for specific localization (ie: fr-FR)` and `// fallback to main lang`

simahawk (commented):
LG, minor adjustments

yajo (approved):
I :heart: it.

website_form_recaptcha/README.rst
```diff
@@ -25,6 +25,9 @@ To use this module, you need to:
 * Set `website_form_recaptcha` to `True` on that model (similar to enabling forms)
 * Add an element with the `o_website_form_recaptcha` class anywhere in the form
 
+Captcha will try to use the language of your website, if it, for whatever
+reason, can't find it will default it google api and use a language of the
+browser.
```
pedrobaeza, line 30: Missing extra line

website_form_recaptcha/README.rst
```diff
@@ -25,6 +25,9 @@ To use this module, you need to:
 * Set `website_form_recaptcha` to `True` on that model (similar to enabling forms)
 * Add an element with the `o_website_form_recaptcha` class anywhere in the form
 
+Captcha will try to use the language of your website, if it, for whatever
```
pedrobaeza:
```suggestion
Captcha will try to use the language of your website. If it, for whatever
```

website_form_recaptcha/README.rst
```diff
@@ -25,6 +25,9 @@ To use this module, you need to:
...
+reason, can't find it will default it google api and use a language of the
```
pedrobaeza:
```suggestion
reason, can't find, it will default to google API and use the language of the
```

pedrobaeza (changes requested):

simahawk (approved):

simahawk:
@pedrobaeza your comments attended

PR #563, opened 2019-01-09
[REF] website_form_recaptcha: better API, allow per-website config

simahawk:
* API keys can now be configured via website settings
  which in turn allow to customize the values per website
  in a multi-website instance;

* move all internal APIs to the model `website.form.recaptcha`
  so that we do not depend anymore on the controller
  if we need to integrate it into other pieces of code.

* readme fragments


tarteo (approved):
Yes, yes! Very nice I like it! Thanks

website_form_recaptcha/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,12 @@
+First of all you must obtain
+a ReCaptcha key from `Google <http://www.google.com/recaptcha/admin>`_
+
+**Global setup**
```
yajo, line 4: I have a question. Does having a global setup make sense in any circumstance? Or is it just to keep backwards compatibility? :thinking: 

If it is the 2nd, then maybe you should add a migration script and we could forget about the global system now and forever.

website_form_recaptcha/controllers/main.py
```diff
@@ -12,6 +11,10 @@
 
 class WebsiteForm(WebsiteForm):
 
+    @property
+    def recaptcha_model(self):
+        return request.env['website.form.recaptcha'].sudo()
```
yajo: Is this property worth the effort? :thinking:

website_form_recaptcha/controllers/main.py
```diff
@@ -12,6 +11,10 @@
 
 class WebsiteForm(WebsiteForm):
 
+    @property
+    def recaptcha_model(self):
```
mpanarin: I'd switch it from property to a simple function.
1. Properties are a nightmare to inherit.
2. You didn't define a setter for it. And someone will definitely try to set it sometime and it will crash the whole thing.

yajo (approved):
A couple of little points to maybe rethink, but good overall.

mpanarin (changes requested):
Other than a small thing, LGTM

website_form_recaptcha/readme/CONFIGURE.rst
mpanarin: I think in multi-website you can easily have same captcha credentials set up, so I don't think this is an issue

simahawk: nope, you can easily have different domains. And this is also the answer for @Yajo 's question:

1. if you have N websites w/ same domain -> use the param
2. if you have N websites w/ different domain -> use website settings
3. it offers backward compat for free
4. if you have global conf you can set it via env files

website_form_recaptcha/controllers/main.py
simahawk: I don't see why it would be a nightmare, they work unless you do tricky things... 

no one will set it... if it happens, you are doing it wrong :)

That being said, not really worth it as Jairo pointed out. I'll strip it out.

website_form_recaptcha/controllers/main.py
simahawk: removed

website_form_recaptcha/readme/CONFIGURE.rst
yajo: >     1. if you have N websites w/ same domain

Is that possible?

mpanarin (approved):

website_form_recaptcha/readme/CONFIGURE.rst
simahawk: well, AFAIS you can set a domain and this will be applied to sub-domains too.
If you set a sub-domain, it works on it only.

I should reword that sentence w/ `N websites w/ different sub-domains but same domain using a single captcha key'

yajo: Oh ok, thanks for the info! :blush:

PR #565, opened 2019-01-14
[12.0] [MIG] website_event_filter_organizer: Migration to v12

sergio-teruel:
cc @Tecnativa


pedrobaeza:
This should go to OCA/event instead.

website_event_filter_organizer/__init__.py
```diff
@@ -0,0 +1,5 @@
+# (c) 2015 Incaser Informatica S.L. - Sergio Teruel
```
pedrobaeza: Remove these copyright lines.

website_event_filter_organizer/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# (c) 2015 Incaser Informatica S.L. - Sergio Teruel
```
pedrobaeza: Update copyright

website_event_filter_organizer/readme/CONTRIBUTORS.rst
```diff
@@ -0,0 +1,2 @@
+* Carlos Dauden <carlos@tecnativa.com>
```
pedrobaeza: Put the company and then the authors in the usual way we do it now.

website_event_filter_organizer/readme/USAGE.rst
```diff
@@ -0,0 +1,2 @@
+To use filter by organizar in website event page you active the filter in
```
pedrobaeza: s/organizar/organizer

pedrobaeza: s/active/should activate

pedrobaeza (changes requested):
Please generate README.rst file, move it to OCA/event, and I'm wondering if this should go inside https://github.com/OCA/event/tree/11.0/website_event_filter_selector

sergio-teruel:
@pedrobaeza Changes done

pedrobaeza (approved):

website_event_filter_organizer/__manifest__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2019 Tecnativa - Sergio Teruel
+# Copyright 2019 Tecnativa - Carlos Dauden
+# License AGPL-3 - See http://www.gnu.org/licenses/agpl-3.0.html
+
+{
```
nikul-serpentcs, line 5: Add `website` key: `https://github.com/OCA/website`

pedrobaeza: I'm seeing now that I said this should go to OCA/event, and it should still be moved.

nikul-serpentcs: @sergio-teruel then URL will change `https://github.com/OCA/event`

website_event_filter_organizer/demo/event_demo.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <!-- event.event -->
+    <record id="event_organizer_0" model="event.event">
```
pedrobaeza, line 5: Why adding this as demo instead on the `setUp` of the test?

pedrobaeza (changes requested):
My request to move this to OCA/event is still valid

website_event_filter_organizer/demo/event_demo.xml
sergio-teruel: Ups!!!!!
As in V11 due to database cursor... Anyway I think that is valid :smile:

sergio-teruel:
Ok, I will move to event repository... I will wait to green CI

website_event_filter_organizer/demo/event_demo.xml
pedrobaeza: Well, I prefer to not put data to be used in tests this way because they can be altered meanwhile, they add noise in real demo data, and any way, you haven't put them as noupdate="1".

sergio-teruel:
Moved to event repository.
@nikul-serpentcs manifest URL changed in event PR

PR #566, opened 2019-01-22
website_form_recaptcha: fix and backward compat

simahawk:
@chienandalu regarding your change from https://github.com/OCA/website/pull/540 can you check it is properly handled? 
I added a specific test for it, still, better to double check :)

I'd also like to use vcrpy to remove most of the mockings in tests, but this is something that can wait.

I added also some backward compat glue to address issues like https://github.com/OCA/apps-store/pull/46
/cc @bizzappdev @pedrobaeza

### DIFF
--- a/website_form_recaptcha/models/website_form_recaptcha.py
+++ b/website_form_recaptcha/models/website_form_recaptcha.py
@@ -4,7 +4,10 @@
 
 from odoo import models, api, http, _
 from odoo.exceptions import ValidationError
+import logging
 import requests
+
+_logger = logging.getLogger(__name__)
 
 
 class WebsiteFormRecaptcha(models.AbstractModel):
@@ -16,10 +19,12 @@ class WebsiteFormRecaptcha(models.AbstractModel):
     _description = 'Website Form Recaptcha Validations'
     URL = 'https://www.google.com/recaptcha/api/siteverify'
     RESPONSE_ATTR = 'g-recaptcha-response'
+    # name of the token attr to store on the request object
+    REQUEST_TOKEN = RESPONSE_ATTR.replace('-', '_')
 
     @api.model
     def _get_error_message(self, errorcode=None):
-        map = {
+        mapping = {
             'missing-input-secret': _('The secret parameter is missing.'),
             'invalid-input-secret':
                 _('The secret parameter is invalid or malformed.'),
@@ -27,8 +32,8 @@ class WebsiteFormRecaptcha(models.AbstractModel):
             'invalid-input-response':
                 _('The response parameter is invalid or malformed.'),
         }
-        return map.get(errorcode, _('There was a problem with '
-                                    'the captcha entry.'))
+        return mapping.get(errorcode, _('There was a problem with '
+                                        'the captcha entry.'))
 
     def _get_api_params(self):
         ICP = self.env['ir.config_parameter'].sudo()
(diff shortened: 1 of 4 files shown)


chienandalu (commented):
Well, if the call to recaptcha is made just once there shouldn't be any problem. `website_crm_phone_validation` provoked a double call, so recaptcha returned an error to the second one, making the validation fail. I think it should be ok if that doesn't happen.

website_form_recaptcha/models/website_form_recaptcha.py
```diff
@@ -80,32 +82,56 @@ def validate_response(self, response, remote_ip, website=None):
...
+        # Store reCAPTCHA's token in the current request object
+        setattr(request, self.REQUEST_TOKEN, req_value)
+        return validated
+
+    # TODO: backward compat, remove in v12
+    @api.model
+    def action_validate(self, req_value, ip_addr):
+        """Backward compatibility for old implementation.
+
+        Prior to API refactoring pre-validation steps
```
yajo: Could you specify which API refactoring? Maybe put a link or something. I'm confused if it's Odoo's, OCA's, or Google's.

website_form_recaptcha/tests/test_captcha.py
```diff
@@ -95,6 +95,43 @@ def test_no_success_raises(self, mk):
...
+        with self.assertRaises(ValidationError) as err:
+            self.model_obj.validate_request(request, req_values)
+        self.assertEqual(
+            err.exception.name, 'The secret parameter is missing.')
+
+    def test_validate_request_old_value(self):
+        request = mock.MagicMock()
+        request.g_recaptcha_response = 'all good here'
+        req_values = {}
+        self.assertTrue(self.model_obj.validate_request(request, req_values))
```
yajo: In this specific case, you probably need to ensure no calls to validation method are done, since the request is repeated.

It would be a bug if the validation is called when it shouldn't.

website_form_recaptcha/tests/test_captcha.py
```diff
@@ -95,6 +95,43 @@ def test_no_success_raises(self, mk):
...
+
+    def test_validate_request_validate_response(self):
+        # Ensure that w/ proper conditions `validate_response is called`
+        request = mock.MagicMock()
+        request.g_recaptcha_response = None
+        # backwarrd compat skip
```
yajo: "Backward"

website_form_recaptcha/tests/test_captcha.py
```diff
@@ -95,6 +95,43 @@ def test_no_success_raises(self, mk):
...
+            mocked.assert_called_with('validate_me', '1.2.3.4')
+
+    def test_action_validate_backward_compat(self):
+        # Ensure that w/ proper conditions `validate_response is called`
+        with mock.patch.object(
+                type(self.model_obj), 'validate_response') as mocked:
+            self.model_obj.action_validate(
+                'captcha_value', '4.5.6.7'
+            )
+            mocked.assert_called_with('captcha_value', '4.5.6.7')
```
yajo: In this specific case (different from the above) you are testing a method implementation instead of its results.

This is a good idea for the case explained above, but IMHO not here. The fact that it calls another method doesn't mean it's going to return that value. It would be better to check instead that the method output is expected.

Also, AFAICS, this test should disappear on v12, right? A TODO note will ease the migration.

yajo (changes requested):

simahawk:
@Yajo you are right on all the points. Addressed. For backward compat tests, to be 100% it behaves as the new api I replicated all the main tests. 
Also, I added a deprecation warning for the old method.

yajo (approved):
:clap: :clap:

rafaelbn (approved):

PR #570, opened 2019-02-19
[12.0][MIG] website_snippet_big_button

cristinamartinrod:
cc @Tecnativa


website_snippet_big_button/templates/snippets.xml
```diff
@@ -8,25 +9,19 @@
...
-                    <div class="text-center col-md-6 col-xs-6">
-                        <div class="btn btn-success js_big_button">
-                            <span class="fa fa-thumbs-o-up fa-5x"/>
-                            <h3 class="mt8 mb0">
-                                <a href="/page/contactus">I want this</a>
-                            </h3>
-                            <p>Request budget</p>
-                        </div>
+                    <div class="text-center col-6">
+                        <a href="/events" class="btn btn-success btn-lg font-weight-bold" style="" target="_blank" contenteditable="true">
```
yajo: Use  `href="/contactus"` as default. It will be always available once website is installed, but someone could not have website_event installed.

Why `style="" target="_blank" contenteditable="true"`?

website_snippet_big_button/templates/snippets.xml
```diff
@@ -8,25 +9,19 @@
...
+                            <span>I WANT THIS</span>
```
yajo: Leave original string to avoid having to re-translate it.

website_snippet_big_button/templates/snippets.xml
```diff
@@ -8,25 +9,19 @@
...
+                            <i class="fa fa-thumbs-o-up fa-5x mt16 mb16" style="text-align: center; float: none; margin: 0px auto;"></i>
```
yajo: Remove inline styles.

website_snippet_big_button/templates/snippets.xml
```diff
@@ -8,25 +9,19 @@
...
-                            <span class="fa fa-thumbs-o-down fa-5x"/>
-                            <h3 class="mt8 mb0">
-                                <a href="/">
-                                    I want another thing
-                                </a>
-                            </h3>
-                            <p>Continue searching</p>
-                        </div>
+                    <div class="text-center col-6">
+                        <a href="/events" class="btn btn-danger btn-lg font-weight-bold">
```
yajo: use href=/, as before

website_snippet_big_button/templates/snippets.xml
```diff
@@ -8,25 +9,19 @@
...
+                            <span>I want another thing</span>
+                            <i class="fa fa-thumbs-o-down fa-5x mt16 mb16" style="text-align: center; float: none; margin: 0px auto;"/>
```
yajo: No inline styles.

website_snippet_big_button/static/src/css/website_snippet_big_button.scss
```diff
@@ -0,0 +1,7 @@
+// Copyright 2015-2016 Jairo Llopis <jairo.llopis@tecnativa.com>
+// Copyright 2019 Tecnativa - Cristina Martin R.
+// License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+.big-buttons a {
+    display: grid;
```
yajo: If using the same styles as there were before in the sass file is enough, I'd prefer you to use that instead of grid, which is cool but seems like an overkill here.

yajo (changes requested):

website_snippet_big_button/static/src/css/website_snippet_big_button.scss
cristinamartinrod: It's not enough with the previous version class, because when the link is modified the elements  are unified. In this way, each element is kept in a line. Maybe seems overkill here because I left text above and text below so that the user has 2 options, If they want can delete one of them.

website_snippet_big_button/templates/snippets.xml
```diff
@@ -8,25 +9,19 @@
...
-                    <div class="text-center col-md-6 col-xs-6">
-                        <div class="btn btn-success js_big_button">
-                            <span class="fa fa-thumbs-o-up fa-5x"/>
-                            <h3 class="mt8 mb0">
-                                <a href="/page/contactus">I want this</a>
-                            </h3>
-                            <p>Request budget</p>
-                        </div>
+                    <div class="text-center col-6">
+                        <a href="/contectus" class="btn btn-success btn-lg font-weight-bold">
```
yajo: typo in link

website_snippet_big_button/templates/snippets.xml
```diff
@@ -8,25 +9,19 @@
...
+                            <span>I want this</span>
+                            <i class="fa fa-thumbs-o-up fa-5x mt16 mb16"></i>
+                            <span contenteditable="true">Request budget</span>
```
yajo: Isn't contenteditable everything? Remove that attr please.

website_snippet_big_button/static/src/css/website_snippet_big_button.scss
yajo: Isn't that because you changed the `<h3>` and `<p>` for 2 `<span>`?

Why did you need to change that? Does this mean that migrated websites buttons will no longer work?

If it was really needed, you can just do `.big-buttons a { span,i { display: block }}` instead.

yajo (changes requested):
Please review travis too. I can't test on runbot, let's see if with next commit... :confused:

website_snippet_big_button/templates/snippets.xml
```diff
@@ -0,0 +1,38 @@
...
+
+    <template id="s_big_button" name="Big Buttons">
+        <section class="oe_snippet_body readable big-buttons">
+            <div class="container">
+                <div class="row mt16 mb16">
+                    <div class="text-center col-6">
+                        <a href="/contactus" class="btn btn-success btn-lg font-weight-bold">
+                            <span>I want this</span>
+                            <i class="fa fa-thumbs-o-up fa-5x mt16 mb16"></i>
+                            <span>Request budget</span>
```
yajo: Why are these spans when the other blocks has h3/p elements?

website_snippet_big_button/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2015-2016 Tecnativa - Jairo Llopis <jairo.llopis@tecnativa.com>
+# Copyright 2019 Tecnativa - Cristina Martin R.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+{
+    "name": "Big Buttons Snippet",
+    "summary": "A snippet that adds two big buttons",
+    "version": "12.0.1.0.0",
+    "category": "Website",
+    "website": "https://www.tecnativa.com/",
```
nikul-serpentcs: @cristinamartinrod URL `https://github.com/OCA/website`

website_snippet_big_button/templates/assets.xml
```diff
@@ -0,0 +1,15 @@
...
+<!-- # Copyright 2015-2016 Tecnativa - Jairo Llopis <jairo.llopis@tecnativa.com>
+     # Copyright 2019 Tecnativa - Cristina Martin R.
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html). -->
+
+<odoo>
+
+<template id="assets_frontend" inherit_id="website.assets_frontend" name="Big buttons snippet frontend assets">
+    <xpath expr="." position="inside">
+        <link rel="stylesheet"
+              href="/website_snippet_big_button/static/src/css/website_snippet_big_button.scss"/>
```
nikul-serpentcs: Directory Name css/scss

website_snippet_big_button/templates/snippets.xml
```diff
@@ -0,0 +1,38 @@
...
+                            <p>Continue searching</p>
+                        </a>
+                    </div>
+                </div>
+            </div>
+        </section>
+    </template>
+
+    <template id="snippets" inherit_id="website.snippets">
+        <xpath expr="//div[@id='snippet_structure']/div[@class='o_panel_body']">
```
nikul-serpentcs: use `hasclass`

nikul-serpentcs (changes requested):
Improve Code

website_snippet_big_button/__manifest__.py
cristinamartinrod: Done!

website_snippet_big_button/templates/assets.xml
cristinamartinrod: Done!

website_snippet_big_button/templates/snippets.xml
cristinamartinrod: Done!

website_snippet_big_button/templates/snippets.xml
cristinamartinrod: Done!

website_snippet_big_button/templates/snippets.xml
```diff
@@ -8,33 +9,27 @@
...
+                        </a>
                     </div>
                 </div>
             </div>
         </section>
     </template>
 
     <template id="snippets" inherit_id="website.snippets">
-        <xpath expr="//div[@id='snippet_structure']/div[@class='o_panel_body']">
+        <xpath expr="//div[@id='snippet_structure']/div[hasclass='o_panel_body']">
```
yajo: hasclass is a function

yajo (changes requested):

website_snippet_big_button/static/src/scss/website_snippet_big_button.scss
```diff
@@ -0,0 +1,7 @@
+// Copyright 2015-2016 Jairo Llopis <jairo.llopis@tecnativa.com>
+// Copyright 2019 Tecnativa - Cristina Martin R.
+// License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+.big-buttons a {
+    display: grid;
```
yajo: Change for `width: 100%`

website_snippet_big_button/templates/snippets.xml
```diff
@@ -0,0 +1,38 @@
...
+
+<odoo>
+
+    <template id="s_big_button" name="Big Buttons">
+        <section class="oe_snippet_body readable big-buttons">
+            <div class="container">
+                <div class="row mt16 mb16">
+                    <div class="text-center col-6">
+                        <a href="/contactus" class="btn btn-success btn-lg font-weight-bold">
+                            <span>I want this</span>
```
yajo: Change for an `<h3>` as it was before, and as it is in the ":-1:" button.

website_snippet_big_button/templates/snippets.xml
yajo: Change for a `<p>` as it was before, and as it is in the ":-1:" button.

yajo (changes requested):
Here I'm asking you to remove some of the changes you have done for the migration.

I already asked for them before, but I'm not sure if I was clear enough, so now you know exactly what to do.

The purpose is:

- Change as few as possible to have it working fine.
- Still benefit from removal of JS code.
- Do not use `grid`, which is too new for some browsers and devs, when `width: 100%` just achieves the same effect.
- Have the same HTML structure in both buttons.

The result, from runbot, with these changes applied:
captura de pantalla de 2019-02-28 10-41-53

website_snippet_big_button/templates/snippets.xml
simahawk: if you include block elements inside anchors it would be better to set "display: block" on them

simahawk (changes requested):
BTW "big button" tells me that I get a nice big button, not that I get some big "like/dislike" set of buttons :trollface:

website_snippet_big_button/templates/snippets.xml
yajo: Why? :thinking: 

I tested it and it works fine...

simahawk: because A is by default an inline element and some browsers do strange things. `inline-block` probably is more appropriated indeed.

Not to mention that semantically a link that includes a paragraph makes little sense and w/ ol' school html < 5 is not valid to have block elems inside a link, but this is another story :man_teacher:

yajo: Well, this has been the structure used by this addon since 2015 and nobody notified a browser rendering problem, so I prefer to keep it as it was unless there's really a rendering problem. :thinking:  It seems that, although `<a>` should be inline, using it as a block is not really a problem in the Real World:tm:

yajo: Let me rectify. I tested on runbot and, when changing the link, it strips out all block tags, so yes, we have to use spans or nothing.

yajo:
@cristinamartinrod please respond/attend/resolve pending comments.

website_snippet_big_button/templates/snippets.xml
yajo: Fixed by adding `class="h3"`

yajo (approved):
All comments attended, design and responsiveness is polished now: 
Peek 17-04-2019 10-24

simahawk (approved):

pedrobaeza (approved):
Missing README by fragment, but OK to go for now

pedrobaeza:
Wait, I'm seeing also some non-squashed commit, so I'm rewriting history with both things.

yajo:
Oh friend plesae, never again do a `git push --force` in a stable branch, please. And even less to just do history makeup (if it were for fixing some information leak or something similar... ok, but for this... :roll_eyes:).

pedrobaeza:
It was just merged, so no consequences.

Issue #551, opened 2018-12-12
website_lazy_load_image: translation broken after install


thanhacun:
After installing the module, all translation (other than default) is broken (in my case, it is Vietnamese) for the website module:

- The translation is OK if a user is logged in
- The translation is Broken if a public user

Please see the attached  file for the website menu

![annotation 2018-12-12 113918](https://user-images.githubusercontent.com/7599781/49847431-9f409680-fe02-11e8-8317-6f6dea51a915.jpg)

Our temporary solution:
- Using unicode parser in /model/ir_ui_view.py file
`unicode_parser = etree.HTMLParser(encoding='utf-8')`
`...`
`html = etree.HTML(res, parser=unicode_parser)`

pedrobaeza:
Which module? Which version?

thanhacun:
Module: oca/website/website_lazy_load_image
Version: 11.0.1.0.0

Behavior:
- After installing the module, translation for Vietnamese (I had only
English and Vietnamese translations) in the website is broken
- It only effects when not login (public users)

Temporary fix:
- In oca/website/website_lazy_load_image/models/ir_ui_view.py file, replace:
   html = etree.HTML(res) with
   html = etree.HTML(res, parser=etree.HTMLParser(encoding='utf-8'))

Thank you and best regards,

Truong Quang Thanh
mob.: +84-94-806-3388

simahawk:
@tarteo ping :)

tarteo:
Thanks @thanhacun, your fix seems to be the solution. Can you make a PR with your fix, please.

thanhacun:
Hi Dennis,

I just found another  problem:
- After installing the module, both original one, and my-temporary-one:
robots.txt file is invalid according to Google Audit Lighthouse tool

- And I still do not know how to do!

Best regards,

Truong Quang Thanh
mob.: +84-94-806-3388

tarteo:
@thanhacun What is the error you get at Google Audit Lighthouse tool? I don't think it's related to the lazy loading module because it doesn't alter the robots.txt or sitemap.xml.

aek:
This simple change fix everything

`html = etree.fromstring(res)`

There is no need for more I think. I have test it and it works as expected. 
The usage of `etree.HTML` seems to add the html structure of `<html><body></body></html>` to enclose the rendered template and that breaks others stuffs.

simahawk:
@aek can you propose a PR and a test to precisely check that we get only the markup we expect?

[...]

PR #571, opened 2019-02-20
website_lazy_load_image: do no wrap non-html

accnops:
Pages like robots.txt got wrapped with a `<html>` and `<p>` tag. This change fixes this.

### DIFF
--- a/website_lazy_load_image/models/ir_ui_view.py
+++ b/website_lazy_load_image/models/ir_ui_view.py
@@ -21,7 +21,8 @@ class IrUiView(models.Model):
         res = super(IrUiView, self).render_template(template, values, engine)
         website_id = self.env.context.get('website_id')
         if website_id and not \
-                self.env['website'].browse(website_id).is_publisher():
+                self.env['website'].browse(website_id).is_publisher() and \
+                '<html' in res.decode():
             html = etree.HTML(res)
             imgs = html.xpath(
                 '//main//img[@src][not(hasclass("lazyload-disable"))]'


tarteo (commented):
In which case (on the website) is the rendered view not HTML?

tarteo:
Oh never mind

website_lazy_load_image/models/ir_ui_view.py
```diff
@@ -21,7 +21,8 @@ def render_template(self, template, values=None, engine='ir.qweb'):
         res = super(IrUiView, self).render_template(template, values, engine)
         website_id = self.env.context.get('website_id')
         if website_id and not \
-                self.env['website'].browse(website_id).is_publisher():
+                self.env['website'].browse(website_id).is_publisher() and \
+                '<html' in res.decode():
```
tarteo, line 25: Is there no better way to check if it's HTML?

accnops: I don't know, I'm not familiar with Odoo, but I agree it's not pretty. Do you have ideas?

tarteo: Is _robots.txt_ the only case? If so you can just do a check `if template == 'website.robots'`. I'm concerned that `<html in res.decode()` will make lazy loading on fixed pages like contact or forum not work anymore.

tarteo (approved):
I tested it for forum and contact page. It seems to be working.

Thanks!

accnops:
A far better approach would be disabling the automatic wrapping of content with the <html> (and <p>) tag. Then we wouldn't need to filter files based on their filename or content. Is this possible?

tarteo:
Like a setting on the page or other website object?

website_lazy_load_image/models/ir_ui_view.py
simahawk: I would not add an exception for specific templates...
Anyway what if you get HTML but no `<html />` tag?
In theory, something like this should do trick:
```py
>>> lxml.html.fromstring('no html').find('.//')
>>> lxml.html.fromstring('<html></html>').find('.//')
>>> lxml.html.fromstring('<html><body></body></html>').find('.//')
<Element body at 0x7f61236c5c78>
```
case 1 and 2: do nothing, case 3: go on.
In this case you should replace `etree.HTML(res)` of course.
Sounds good?

simahawk (changes requested):

simahawk:
related issue https://github.com/OCA/website/issues/551

hugho-ad:
@simahawk 

Could you test following pr please
Hi, could you test following pr
https://github.com/OCA/website/pull/651

I apply the patch you suggets and  the res.decode part of this pr

PR #572, opened 2019-02-25
[12.0][MIG] website_form_recaptcha: Migration to v12

cristinamartinrod:
cc @Tecnativa


pedrobaeza:
Check runbot and Travis status and squash some adjacent translation commits.

website_form_recaptcha/models/website_config_settings.py
```diff
@@ -1,15 +1,34 @@
...
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class ResConfigSettings(models.TransientModel):
     _inherit = 'res.config.settings'
 
     recaptcha_key_site = fields.Char(
         related='website_id.recaptcha_key_site',
+        readonly=False,
```
yajo, line 12: Related fields not readonly are a good idea? Or transient models are the exception? cc @pedrobaeza

website_form_recaptcha/models/website_config_settings.py
```diff
@@ -1,15 +1,34 @@
...
     )
     recaptcha_key_secret = fields.Char(
         related='website_id.recaptcha_key_secret',
+        readonly=False,
```
yajo, line 16: Same here.

website_form_recaptcha/models/website.py
```diff
@@ -1,11 +1,12 @@
...
-from odoo import models, fields
+from odoo import api, fields, models
 
 
 class Website(models.Model):
     _inherit = 'website'
 
     recaptcha_key_site = fields.Char()
     recaptcha_key_secret = fields.Char()
+    has_google_recaptcha = fields.Boolean()
```
yajo: This field does nothing AFAICS. Remove it please.

website_form_recaptcha/models/website_config_settings.py
```diff
@@ -1,15 +1,34 @@
...
+    )
+
+
+    @api.depends('website_id')
+    def has_google_recaptcha(self):
```
yajo: This method should be private (start by underscore), and never be called with the same name as a field. Call it `_compute_has_google_recaptcha`.

website_form_recaptcha/models/website_config_settings.py
```diff
@@ -1,15 +1,34 @@
...
+        self.has_google_recaptcha = bool(self.recaptcha_key_site)
+
+    def inverse_has_google_recaptcha(self):
```
yajo: Start with `_` also. Public methods are meant to be called from XMLRPC, and inverse methods should not.

website_form_recaptcha/models/website_config_settings.py
```diff
@@ -1,15 +1,34 @@
...
+        if not self.has_google_recaptcha:
+            self.recaptcha_key_site = False
+            self.recaptcha_key_secret = False
+
+    has_google_recaptcha = fields.Boolean(
+        "Google reCaptcha",
+        compute=has_google_recaptcha,
+        inverse=inverse_has_google_recaptcha,
```
yajo: Field definitions should go at the beginning of the class declaration per guideline.

Use a string that indicates the method names for compute and inverse instead of the raw method object.

yajo (changes requested):

website_form_recaptcha/data/ir_config_parameter_data.xml
```diff
@@ -0,0 +1,18 @@
...
+
+
+<odoo noupdate="1">
+    <record id="recaptcha_key_site" model="ir.config_parameter">
+        <field name="key">recaptcha.key.site</field>
+        <field name="value">Obtain from Google</field>
+    </record>
+    <record id="recaptcha_key_secret" model="ir.config_parameter">
+        <field name="key">recaptcha.key.secret</field>
+        <field name="value">Obtain from Google</field>
```
yajo: Remove this file. Having wrong configurations is wrong, in general. Also, isn't it supposed that these keys are used per website? Websites have fields for such. Do not store things in 2 places, it's confusing.

website_form_recaptcha/models/website_config_settings.py
pedrobaeza: This is on purpose, as this is for the settings part, where the change is "transferred" to the corresponding model, so this is totally correct.

website_form_recaptcha/models/website_form_recaptcha.py
```diff
@@ -0,0 +1,143 @@
...
+                _('The response parameter is invalid or malformed.'),
+        }
+        return mapping.get(errorcode, _('There was a problem with '
+                                        'the captcha entry.'))
+
+    def _get_api_params(self):
+        ICP = self.env['ir.config_parameter'].sudo()
+        return {
+            'site_key': ICP.get_param('recaptcha.key.site'),
+            'secret_key': ICP.get_param('recaptcha.key.secret'),
```
yajo: These should come from the `website` object, right?

website_form_recaptcha/models/website_form_recaptcha.py
```diff
@@ -0,0 +1,143 @@
...
+            ip_addr = ip_addr.split(',')[0]
+        else:
+            ip_addr = request.httprequest.remote_addr
+        # if not validated an exception is raised anyway
+        validated = self.validate_response(req_value, ip_addr)
+        # Store reCAPTCHA's token in the current request object
+        setattr(request, self.REQUEST_TOKEN, req_value)
+        return validated
+
+    # TODO: backward compat, remove in v12
```
yajo: Well... you know :wink: 

Check #566 for details.

website_form_recaptcha/tests/test_captcha.py
```diff
@@ -0,0 +1,210 @@
...
+            'recaptcha_key_secret': '0123456789',
+        })
+        creds = self.model_obj._get_api_credentials(website=self.website)
+        self.assertEqual(creds, {
+            'site_key': '1234567890',
+            'secret_key': '0123456789',
+        })
+
+    # TODO: backward compat tests for `action_validate`` remove all in v12
+    # START OF BACKWARD COMPAT TESTS
```
yajo: This section should be removed too.

cc @simahawk

yajo (changes requested):

website_form_recaptcha/views/website_config_settings.xml
```diff
@@ -1,31 +1,42 @@
...
     Copyright 2018 Simone Orsi - Camptocamp SA
+    Copyright 2019 Tecnativa - Cristina Martin R.
     License AGPL-3 or later (http://www.gnu.org/licenses/agpl.html).
 -->
 <odoo>
 
     <record id="view_website_config_settings" model="ir.ui.view">
+        <field name="neme">reCaptcha Settings</field>
         <field name="model">res.config.settings</field>
         <field name="inherit_id" ref="website.res_config_settings_view_form" />
```
MiquelRForgeFlow, line 12: website_form.res_config_settings_view_form

cristinamartinrod: Thank you for your review @mreficent , but I think it's better display it on global view, in the communication section, next to configuration forms. Or is it for another reason?

MiquelRForgeFlow: This module depends of `website_form`, so if `website_form` has a view, then I think you should here inherit from this view instead to avoid possible conflicts between views. You can be calm, it doesn't affect the rest of the code. The things you display in this view will still be displayed in the same place.

pedrobaeza: But the arch in which this inherited view relies is in `website_form` or in `website`?

MiquelRForgeFlow: Is it relevant?

pedrobaeza: Yes, you always have to try to inherit from the top view for not depending on useless inheritances.

MiquelRForgeFlow: But here the top view should be which? The one directly affected by the arch (website) or the one that depends directly by the module dependencies (website_form)? "Useless inheritances" is subjective.

MiquelRForgeFlow: Hey, I remind that this is a non-blocking "comment", not a "request changes". If you consider that it's better not do "useless inheritances", then don't do it.

pedrobaeza: I know, but as always, I like to have with you a sane discussion about the topic for being instructive for all. Imagine this simplified archs:
id='module_a.view1'
```xml
<form>
  <field name="X"/>
</form>
```
id='module_b.view1' inherit_id='module_a.view1'
```xml
<field name="X" position="after">
  <field name="Y"/>
</form>
```
if you need to make a new view for placing something after field X, you don't need to inherit from `module_b.view1`, but from `module_a.view1`, as the field X is already defined there. If you need to place your field referencing field Y, then you must inherit indeed from module_b, and this is independent if you need to depend your module in module_b in the manifest due to other thing.

Do you understand me?

MiquelRForgeFlow: If module_b.view1 is the following:
```xml
<field name="X" position="replace">
</field>
```
or similar with `position="move"`, then, you may have problems.

pedrobaeza: Well, then the problem is because module_b is bad designed. See my talk about module compatibility, which basically forbids the `replace` operation if you want a "compatible module". But if not, `module_b.view1` priority should be set to a very low priority for being processed the last one. That way, there's no problem.

MiquelRForgeFlow: But I think you can't avoid bad designed modules or people using `replace` 🤔

MiquelRForgeFlow: Buy hey, you got my point and I got your point. @cristinamartinrod feel free  😁

pedrobaeza: Not really, because there can always be other views outside your radar that makes the same, so, always inherit on the lower view that you need, and don't put in your system bought apps from Apps store :stuck_out_tongue_closed_eyes:

website_form_recaptcha/data/ir_config_parameter_data.xml
simahawk: it has been done to:

* support server env for global conf on all websites
* support per-website override

I agree on removing fake values

website_form_recaptcha/models/website_form_recaptcha.py
simahawk: no, as explained above they can come via params

simahawk (changes requested):

website_form_recaptcha/data/ir_config_parameter_data.xml
yajo: I have double-checked this and in v12, now that multiwebsite is mainstream, all settings for a website are always stored in the website object itself, and configurable by the config view.

I think the all-websites setting makes no sense anymore. The new approach makes more sense, and is UI-friendly. We'd have less code to maintain, and the workaround would be as simple as copy-paste from one website to the other, and a simple website admin user can do that without asking anybody.

Let's remove this completely please @cristinamartinrod

yajo (approved):
All comments attended, tests fixed.

website_form_recaptcha/__manifest__.py
```diff
@@ -13,6 +13,11 @@
     "depends": [
         "website_form",
     ],
+    "external_dependencies": {
+        "python": [
+            "openupgradelib",
```
MiquelRForgeFlow: packages for the migration are not needed to put here.

MiquelRForgeFlow (changes requested):

website_form_recaptcha/migrations/11.0.1.0.0/post-migrate.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2019 Jairo Llopis - Tecnativa
```
pedrobaeza: This should be removed

pedrobaeza (changes requested):
Remove the migration script and the library in external dependencies, and check Travis + runbot.

website_form_recaptcha/migrations/11.0.1.0.0/post-migrate.py
yajo: Why? It's in the template.

pedrobaeza: I mean the whole file. It's from the previous version.

yajo: Oh, no, I just got the number wrong.

pedrobaeza: But is it something you have added now?

yajo: Yes, it's needed.

pedrobaeza (approved):

pedrobaeza:
@simahawk do you approve the PR with latest changes?

MiquelRForgeFlow (approved):
LGTM code review

HviorForgeFlow (approved):

pedrobaeza:
I think this one has consensus, so I merge.

website_form_recaptcha/models/website_config_settings.py
```diff
@@ -0,0 +1,32 @@
...
+        inverse='_inverse_has_google_recaptcha',
+        readonly=False,
+    )
+
+    @api.depends('website_id')
+    def _compute_has_google_recaptcha(self):
+        self.has_google_recaptcha = bool(self.recaptcha_key_site)
+
+    def _inverse_has_google_recaptcha(self):
+        if not self._compute_has_google_recaptcha:
```
simahawk, line 30: hmm this seems wrong: `self._compute_has_google_recaptcha` is an object, is always True :thinking:

yajo: Indeed :roll_eyes: 

Let me open a new PR

yajo: Fixed in #589

PR #578, opened 2019-03-06
[10.0] Allow to install website_menu_by_user_status

BenjaHe:
Enable installation of this module for V10


claudiagn (commented):
@BenjaHe only installation or also functional review?

PR #581, opened 2019-03-20
[12.0] [MIG] website_blog_excerpt_image: Migration to v12

cristinamartinrod:
cc @Tecnativa


website_blog_excerpt_img/README.rst
```diff
@@ -0,0 +1,68 @@
+.. image:: https://img.shields.io/badge/licence-LGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/lgpl-3.0-standalone.html
+   :alt: License: LGPL-3
+
+=======================
+Excerpt + Image in Blog
```
nikul-serpentcs: Excerpt And Image in Blog

website_blog_excerpt_img/__manifest__.py
```diff
@@ -0,0 +1,26 @@
+# © 2016 Grupo ESOC Ingeniería de Servicios, S.L.U. - Jairo Llopis
+# © 2019 Tecnativa - Cristina Martin R.
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+{
+    "name": "Blog List view",
+    "summary": "New layout for blog summary, including an excerpt and image",
+    "version": "12.0.1.0.0",
+    "category": "Website",
+    "website": "https://grupoesoc.es",
```
nikul-serpentcs: @cristinamartinrod `https://github.com/OCA/website`

nikul-serpentcs (changes requested):
nitpicking

website_blog_excerpt_img/__manifest__.py
```diff
@@ -1,14 +1,15 @@
...
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
 {
-    "name": "Excerpt + Image in Blog",
+    "name": "Blog List view",
     "summary": "New layout for blog summary, including an excerpt and image",
-    "version": "9.0.1.0.0",
+    "version": "12.0.1.0.0",
     "category": "Website",
     "website": "https://grupoesoc.es",
     "author": "Grupo ESOC Ingeniería de Servicios, "
```
pedrobaeza: You can remove this one, as `Tecnativa` replaces it

website_blog_excerpt_img/__manifest__.py
```diff
@@ -18,10 +19,8 @@
     "depends": [
         "website_blog",
         "html_image_url_extractor",
-        "html_text",
```
pedrobaeza: Don't we need this one?

website_blog_excerpt_img/templates/blog.xml
```diff
@@ -8,34 +8,39 @@
...
+        <!-- Read more button -->
+        <xpath expr="//div[@class='o_sharing_links']" position="inside">
+        <div class="float-right">
+                <a t-attf-href="/blog/#{ slug(blog_post.blog_id) }/post/#{ slug(blog_post) }"
+                    class="btn btn-primary o_default_snippet_text">
+                    Read more
+                </a>
+        </div>
+        </xpath>
+        <xpath expr="//hr" position="replace">
```
pedrobaeza: Don't use `position="replace"`. Hide the element instead.

website_blog_excerpt_img/__manifest__.py
```diff
@@ -18,10 +19,8 @@
...
     ],
     "data": [
-        "templates/assets.xml",
```
pedrobaeza: Don't we need styles?

website_blog_excerpt_img/README.rst
tarteo: @Nikul-Chaudhary Why? This file will be generated.

website_blog_excerpt_img/README.rst
```diff
@@ -0,0 +1,68 @@
+.. image:: https://img.shields.io/badge/licence-LGPL--3-blue.svg
```
tarteo: Please use the split readme

website_blog_excerpt_img/__manifest__.py
```diff
@@ -18,10 +19,8 @@
     "depends": [
         "website_blog",
         "html_image_url_extractor",
```
martintorresvm, line 21: @cristinamartinrod Would need add `oca_dependencies.txt` with `server-tools` to provide this dependency

martintorresvm (changes requested):
To fix Travis

yajo (changes requested):
I was fixing this but realized that possibly this addon is no longer required. See:

1. In older odoo versions, we had the problem of huge blog post covers. Not everybody wanted to add such a big cover. Now, in odoo 12, we have the possibility to use a narrow blog cover image: 
Captura de pantalla de 2019-04-17 10-59-23 This gives an option to have a properly-sized image; good for most cases. Themes can style this too if we want it other size.

1. The excerpt is automatically generated now. It comes from [a new field called `teaser` (in fact there's a `teaser_manual`](https://github.com/odoo/odoo/blob/c2f15d3f1ab87516fc2b928a2136a74b34890a02/addons/website_blog/models/website_blog.py#L155-L156), which is hidden but probably used in enterprise to define manually a teaser). Default odoo 12 upstream view:
Captura de pantalla de 2019-04-17 11-06-44

1. If we had to develop a module that wanted to show a different image in the post list view, probably the best source to get it would be now to reuse the [`website_meta_og_img` field](https://github.com/odoo/odoo/blob/5a692c34174bd9b9f1460e8fb5df1028db7dcca5/addons/website/models/website.py#L813), which chooses a main image for social sharing and is easily accessible from the _Optimize SEO_ wizard. In fact it would be a good place where to extract a manually-written excerpt also: 
Captura de pantalla de 2019-04-17 11-12-26

1. If we want to have a different look for the list view, the best way is to use customize, just like upstream odoo does (which, BTW, provides an alternate display quite good looking too): 
Captura de pantalla de 2019-04-17 11-16-35

So, if we want to still have this module, both the excerpt and the img autoselection features have to be refactored. The best we can do is drop this and develop independently those new versions if ever needed.

yajo:
OK, this is how it looks now, in action:

ezgif-4-4d99dc3525eb

Details in last commit. It should be ready now.

yajo (approved):

pedrobaeza:
Please check runbot and Travis status

yajo:
:heavy_check_mark: now.

website_blog_excerpt_img/templates/assets.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright 2019 Tecnativa - Jairo Llopis
+     License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl). -->
+
+<data>
```
nikul-serpentcs: @cristinamartinrod ` <data>` --> `<odoo>`

website_blog_excerpt_img/templates/assets.xml
```diff
@@ -0,0 +1,14 @@
...
+
+    <template id="assets_frontend" inherit_id="website.assets_frontend">
+        <xpath expr=".">
+            <link rel="stylesheet"
+                  href="/website_blog_excerpt_img/static/src/css/website_blog_excerpt_img.scss"/>
```
nikul-serpentcs: `css` --> `scss`

nikul-serpentcs (changes requested):
Minor Change

website_blog_excerpt_img/templates/assets.xml
pedrobaeza: `<data>` is also valid and it's preferred by our colleague @Yajo who has been the last to touch this module, due to its neutrality about branding.

website_blog_excerpt_img/templates/assets.xml
pedrobaeza: It's already converted...

website_blog_excerpt_img/README.rst
```diff
@@ -22,47 +52,59 @@ Usage
...
+
+Do not contact contributors directly about support or help with technical issues.
 
 Credits
 =======
 
+Authors
+~~~~~~~
+
+* Grupo ESOC Ingeniería de Servicios
```
pedrobaeza: Should we still keep this author? I don't think so.

website_blog_excerpt_img/models/blog_post.py
```diff
@@ -1,32 +1,25 @@
...
-                cover,
-                html,
-                q='"' if '"' not in cover else "'",
-            )
-        # Return the first found image URL or None
-        try:
-            return next(Converter.imgs_from_html(html, 1))
-        except StopIteration:
-            return None
+    def _default_website_meta(self):
```
pedrobaeza, line 10: We should do this on each write if no previous default meta-image, not only on creation, or this is called each time?

pedrobaeza (dismissed):
Please add a migration script that activates the custom view and deactivates the others.

website_blog_excerpt_img/templates/assets.xml
nikul-serpentcs: @pedrobaeza I mean a directory name

`/website_blog_excerpt_img/static/src/css/` --> `/website_blog_excerpt_img/static/src/scss/`

pedrobaeza: The directory is the same for less, sass, scss or css: `css` as generic one for all.

website_blog_excerpt_img/README.rst
yajo: I think it's the most elegant. They still exist and they actually authored the module, so I see no reason for removing them. This is not "maintainers", after all.

website_blog_excerpt_img/models/blog_post.py
yajo: This is called when obtaining the social share details. [This is the method used to provide defaults](https://github.com/odoo/odoo/blob/3ac23bef79e3e0f8179c0e9441d6a81acdc55d1f/addons/website/models/website.py#L865-L867) (which is actually overriden in blog posts to return the blog cover, if set).

So, we don't want to write an option, but provide a default value, while still letting the user to choose a different value if he wants.

In short, this is the place to do this. :wink:

yajo (commented):
> Please add a migration script that activates the custom view and deactivates the others.

Sorry I don't understand... Which views do you want me to activate/deactivate? :thinking:

website_blog_excerpt_img/README.rst
pedrobaeza: I don't agree. Here elegance has nothing to do. You have even rewritten almost everything and you are Tecnativa's employee, not Grupo ESOC's one. They already have the credit for previous versions.

website_blog_excerpt_img/models/blog_post.py
pedrobaeza: I think you are not understanding me. Imagine this situation:

- You create a blog page with no image.
- Save.
- Then you edit again and add an image.

My question is: in this case, the later added image is going to be auto-selected?

pedrobaeza:
> Sorry I don't understand... Which views do you want me to activate/deactivate?

The view with the custom layout for blog list that is adding this module. Putting `customize_show` makes the view (template) to not being active by default, while other option (grid or list) is activated, so I want that this new behavior is active by default.

website_blog_excerpt_img/models/blog_post.py
yajo: Yes.

yajo:
But it is also marked as `active="True"`... Isn't that enough?

website_blog_excerpt_img/templates/blog.xml
```diff
@@ -0,0 +1,49 @@
...
+        priority="50"
+    >
+
+        <xpath expr="//*[hasclass('o_blog_post_teaser')]/.." position="replace">
+            <div class="row">
+                <!-- Wrap text content with a .row element -->
+                <div class="col">$0</div>
+
+                <!-- Insert social share thumbnail next to teaser -->
+                <div class="col-5 text-center">
```
Tardo, line 22: Here uses text-center, but 'excerpt-img' has the "float:right;" style...

yajo: Good catch, thanks. Fixed.

yajo:
Can we merge?

website_blog_excerpt_img/README.rst
```diff
@@ -0,0 +1,109 @@
+==============================================
+Blog Automatic Share Image Selection & Display
```
pedrobaeza: This title is incorrect

pedrobaeza (changes requested):

website_blog_excerpt_img/README.rst
yajo: Fixed.

pedrobaeza (approved):

Issue #580, opened 2019-03-18
[10.0] website_cookie_notice: cookie removed after closing browser


oerp-odoo:
Is it intended to clear cookies after browser is closed? To me it looks like a bug, but want to make sure if it is.

In JS part:
```javascript
document.cookie = 'accepted_cookies=1; path=/';
```
If it is done like this, browser uses default to clear cookie after browser is closed. To make it stay, one must specify either `expires` or `max-age` attribute alongside. For example:

```javascript
document.cookie = 'accepted_cookies=1; path=/; max-age=31536000';
```
Where in this example one year of age is specified. I could create PR, if this is a bug, if not, would be nice to get a reasoning why it was chosen to clear cookie after browser is closed (It should be annoying for user to get spammed with cookie notice when he already agreed/read it, every time he reopens browser and opens website).

yajo:
IMHO that's a bug. Please feel free to open the PR. Thanks!

PR #582, opened 2019-03-20
[FIX] website_cookie_notice: max-age for cookie

oerp-odoo:
Default cookie behavior is to expire on session. But it is not expected
behavior for cookies consent. So instead we specify max-age to store
cookie up to a year.

Fixes https://github.com/OCA/website/issues/580

### DIFF
--- a/website_cookie_notice/README.rst
+++ b/website_cookie_notice/README.rst
@@ -69,6 +69,7 @@ Contributors
 * Rafael Blasco <rafael.blasco@tecnativa.com>
 * Jairo Llopis <jairo.llopis@tecnativa.com>
 * Wolfgang Pichler <wpichler@callino.at>
+* Andrius Laukavičius <dev@focusate.eu> (Focusate)
 
 Maintainer
 ----------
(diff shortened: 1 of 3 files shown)


pedrobaeza (approved):

pedrobaeza:
This should be backported to 11.0/12.0

oerp-odoo:
@pedrobaeza I guess, but cookies acceptance is implemented differently in 11.0/12.0, using session storage (don't know the reasoning it was changed on newer versions). Do you suggest replacing it with same approach as on 10.0 or reimplement it to clear after some time using session? I think with session storage, it won't be cleared if not done manually (like with cookies it is)?

pedrobaeza:
No, I don't have any strong opinion about it. @Yajo might tell.

yajo:
The v10 branch has #461, which added this behavior. It is an improvement done after it was migrated to branches 11 and 12, that's why implementation differs.

The ideal would be to forward-port #461 and this patch for those branches.

To know why it's better, see https://github.com/OCA/website/pull/461#pullrequestreview-162361527.

yajo (approved):
Travis error is not related, main branch is :x:

hugho-ad (approved):

Issue #584, opened 2019-03-23
[11.0] Website Form Builder: *Required Fields


UnixedIT:
Hello, 

When Submitting the Form without "user log in" all mandatory fields appear as optional fields.

please find the attached 1.png.

if the user log-in there is no problem as attached  2.png.

![1](https://user-images.githubusercontent.com/32422485/54866122-e1219700-4d78-11e9-95ec-a32abfea5305.png)
![2](https://user-images.githubusercontent.com/32422485/54866123-e1219700-4d78-11e9-8012-b6c79280d810.png)

yajo:
See fix at #585

UnixedIT:
Thank you @Yajo for fixing. 

-as you know-please note the fix is working for Styling only. but the public users still able to submit the form event they don't fill the mandatory fields.

seyuf:
@pedrobaeza can you please reopen this? As the issue is still there.
The css has been fixed but the wrong behavior is still there. I.e
![required-field-not-set](https://user-images.githubusercontent.com/3765910/65377528-9d6b0400-dcad-11e9-9655-33357980fc31.png)

The form appears valid even though the email (courriel) field has not been filed.
On the screenshot, one can notice the say input is missing the required attribute like on the first one.

Possible workaround: set `required="required"` through the html editor

rafaelbn:
Dear @seyuf , reproduce the error here in runbot. I cannot reproduce:

http://3385143-12-0-9bc24f.runbot2-2.odoo-community.org/

![2019-09-21_21-51-58](https://user-images.githubusercontent.com/8736623/65378582-94ccfa80-dcba-11e9-8e73-2268c4cb492d.png)

seyuf:
Hi @rafaelbn ,
Im using odoo 11 after a quick test on your runbot.
The form is sent as long as the first field is filled. Which shouldn't be the case.
It should be sent only when all the required fields are.
![error-required-field](https://user-images.githubusercontent.com/3765910/65378670-28eb9180-dcbc-11e9-8847-6b7ffabd3caf.jpeg)

Also, on my version, the backend check, contrary to the runbot, considers the form as valid.
Maybe we're not using the same version of the module. I'll check and keep you posted tomorrow.

pedrobaeza:
You can't try this on runbot without uninstalling `website_form_recaptcha`, as both are incompatible. Remove it and all should be OK.

seyuf:
[...]

PR #585, opened 2019-03-25
[FIX] website_form_builder: Fix missing required field indicators

yajo:
Styling should be available for public users too.

Fixes https://github.com/OCA/website/issues/584.

@Tecnativa


pedrobaeza (approved):

luismontalba (approved):
Tested on runbot

PR #534, opened 2018-10-31
12.0 mig website logo

eslAmer:

PR #623, opened 2019-05-09
[12.0][MIG] website_logo: Migration to 12.0

ernestotejeda:
Cc @Tecnativa


pedrobaeza:
There's no CLA for eslam, so please start the migration from scratch.

website_logo/models/website.py
```diff
@@ -4,7 +4,4 @@
 class Website(models.Model):
     _inherit = 'website'
 
-    logo = fields.Binary(
-        string="Website logo",
-        help="This field holds the logo for this website, showed in header. "
-             "Recommended size is 180x50")
+    logo = fields.Binary()
```
chienandalu: Why this change?

chienandalu (approved):
Just a little comment, otherwise ok

website_logo/models/website.py
ernestotejeda: You are right, i think *string* and *help* attributes should not have been removed. I have added them again. thanks.

ernestotejeda:
Changes done

website_logo/views/website_templates.xml
```diff
@@ -0,0 +1,13 @@
+<odoo>
+    <template id="add_logo_to_website" inherit_id="website.layout_logo_show" customize_show="False" name="Show Logo">
+        <xpath expr="//header//a[hasclass('navbar-brand')]" position="after">
+            <a href="/" class="navbar-brand logo">
+                <span t-field="website.logo" t-options="{'widget': 'image'}" role="img" t-att-aria-label="'Logo of %s' % website.name" t-att-title="website.name"/>
```
Tardo: Here can use 't-attf' instead of 't-att'

website_logo/views/website_templates.xml
```diff
@@ -0,0 +1,13 @@
...
+            </a>
+        </xpath>
+        <xpath expr="//header//a[hasclass('navbar-brand')]" position="attributes">
+            <attribute name="class" add="invisible" separator=" "/>
```
Tardo: Here can use 'd-none' (bootstrap4) instead of 'invisible'... this hidden the element in the layout.

Tardo (changes requested):

ernestotejeda:
Changes done

Tardo (approved):
LGTM, Tested on runbot

Perhaps can be good use the company logo when no logo defined for the website.. ?¿??

website_logo/views/website_templates.xml
```diff
@@ -0,0 +1,13 @@
+<odoo>
+    <template id="add_logo_to_website" inherit_id="website.layout_logo_show" customize_show="False" name="Show Logo">
+        <xpath expr="//header//a[hasclass('navbar-brand')]" position="after">
+            <a href="/" class="navbar-brand logo">
+                <span t-field="website.logo" t-options="{'widget': 'image'}" role="img" t-attf-aria-label="Logo of {{website.name}}" t-att-title="website.name"/>
+            </a>
+        </xpath>
+        <xpath expr="//header//a[hasclass('navbar-brand')]" position="attributes">
+            <attribute name="class" add="d-none" separator=" "/>
+            <attribute name="style" add="Max-width:0px;margin:0px;"/>
```
Tardo: I missed this, you don't need 'ensure' that the element is 'hidden'... because it is not present in the layout (d-none).

Tardo (changes requested):

website_logo/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,6 @@
+To configure this module, you need to:
+
```
sergio-teruel, line 2: Remove this two lines

website_logo/readme/USAGE.rst
```diff
@@ -0,0 +1,4 @@
+To use this module, you need to:
+
```
sergio-teruel, line 2: Remove this two lines.

sergio-teruel (changes requested):

website_logo/readme/CONFIGURE.rst
ernestotejeda: And I was guided by this.
https://github.com/OCA/maintainer-tools/blob/master/template/module/readme/CONFIGURE.rst
 Should I remove those two lines anyway?

website_logo/readme/USAGE.rst
ernestotejeda: > And I was guided by this.
> https://github.com/OCA/maintainer-tools/blob/master/template/module/readme/CONFIGURE.rst
> Should I remove those two lines anyway?


and here https://github.com/OCA/maintainer-tools/blob/master/template/module/readme/USAGE.rst

ernestotejeda:
> LGTM, Tested on runbot
> 
> Perhaps can be good use the company logo when no logo defined for the website.. ?¿??

@Tardo , to achieve this, i think, it would be necessary to do something similar to what was done in previous versions, make a treatment in the controller to return the logo of the company in case a logo has not been specified in the website. But this version of the module, simpler, seems better to me. And I think it maintains the idea of the module, which is to have the company logo on one side and, on the other hand, the logo of the website.
I think this is debatable, maybe we should see the opinion of other people.

pedrobaeza:
Add it to the ROADMAP and lets' move on with this.

ernestotejeda:
Thanks @Tardo , @pedrobaeza , i added it to the roadmap

website_logo/views/website_templates.xml
```diff
@@ -1,13 +1,13 @@
-<?xml version="1.0" encoding="utf-8"?>
 <odoo>
-
-    <template id="layout_logo_show" inherit_id="website.layout_logo_show">
-        <xpath expr="//a//span" position="attributes">
-            <attribute name="class" add="hide" separator=" "/>
+    <template id="add_logo_to_website" inherit_id="website.layout_logo_show" customize_show="False" name="Show Logo">
```
yajo: Better `customize_show="True" active="True"`. This way, this default behavior is enabled, but can be disabled per website.

Just adding that should clear the task that you specified in `ROADMAP.rst`.

website_logo/views/website_templates.xml
yajo: Yes, you can remove this.

yajo (changes requested):
I'm not saying it's wrong, but I see you removed all translations and tests... Why?

Some other notes below.

ernestotejeda:
> I'm not saying it's wrong, but I see you removed all translations and tests... Why?
> 
> Some other notes below.


@Yajo  The author who start with this migration in other PR https://github.com/OCA/website/pull/534 didn't respond anymore and I made the new PR from his code. The author made several changes, he removed the controllers/main.py file, so, the tests were removed too.
He had already deleted all translations when I took the PR, and I left it in that way. Now i added an es.po file.

yajo (dismissed):
:man_facepalming: Oh sorry, I'm very sad to say this now, but this module is not needed. It's supported upstream in Odoo 12. See: 
Peek 03-06-2019 09-53

pedrobaeza:
Then 2 things are needed:

- Merge the module in OpenUpgrade with `website`.
- Add code for not losing all logos with new scheme.

chienandalu:
@Yajo I'd say this is still needed. Although you can have the company logo as the default website logo (we could remove that from the roadmap), an website specific logo is going to be needed in a lot of cases.

yajo:
I see, I didn't notice that it was the company logo.

Couldn't we just add a `customize_show=True` view that replaces the company logo for another logo, that the user can just double-click and change, like the gif I'm doing?

When you alter that view, the altered copy is made specific for the website and there's all the data you need. No need to alter models or wizards for example.

Also, in v12 AFAIK svg is supported, so you can use a company logo in SVG that works fine for both reports and website, which could be the main reason why we actually needed this addon after all... :thinking:

pedrobaeza:
@Yajo there are more reasons for having this module:

* Multi websites and each one with a specific trademark.
* Not same "visible" logo than company logo.
* Others.

In general, we want to detach the logo we want in website from the one we have in company.

yajo:
OK, then how about implementing it just in a template?

pedrobaeza (approved):

Issue #629, opened 2019-05-11
[10.0] travis build fails


tbaden:
the module website_sale this causing the build TESTS="1" ODOO_REPO="odoo/odoo" to fail.

2019-05-11 17:04:24,146 7269 ERROR openerp_test odoo.addons.website_sale.models.sale_order: Fail to find pricelist for partner "Public user" (id 5)

to see in PR:
https://github.com/OCA/website/pull/567
https://github.com/OCA/website/pull/624
https://github.com/OCA/website/pull/625

First occurance around 11 june 2018 PR: https://github.com/OCA/website/pull/479
Correction: 479 was rebuild 3 months ago. First ist occurance ist 567 4 month ago, so far

pedrobaeza:
@sergio-teruel is this the error you have mentioned before?

tbaden:
currently struggeling to reproduce this error locally, but I'm on it.

tbaden:
after some late night investigation it seems that website_sale_line_total and website_sale_hide_empty_category are both causing this issue. disabling just one of them still causes the issue.it's not Website_sale either, added this dependency in unrelated repo(project) without any issues. Must be a combination of different dependencies. further investigation is planned for tomorrow, to find the cause and a fix for this.

tbaden:
the test:
```
    def test_127_0_0_1(self):
        """Check 127.0.0.1 downloads its default assets."""
        response = self.url_open("http://127.0.0.1:%d" % PORT, timeout=60)
        self.assertEqual(response.getcode(), 200)
        ...
```
is the cause of all this. seems like urllib2 has problems with 127.0.0.1. this needs further investigation, but the train wifi is killing my researches...
in v11 it got changed to:
```
    def test_0_0_0_0(self):
        """Check 0.0.0.0 downloads its default assets."""
        response = self.url_open("http://0.0.0.0:%d" % PORT, timeout=60)
        self.assertEqual(response.status_code, 200)
        ...
```

@pedrobaeza do you know by any change any known issues with urllib2 and 127.0.0.1.
was there a travis update or somehting like this around january/february? this would explain the spontaneous error.

btw.
```
    def test_localhost(self):
        """Check localhost downloads its multiwebsite-enabled assets."""
        response = self.url_open("http://localhost:%d" % PORT, timeout=60)
        self.assertEqual(response.getcode(), 200)
        ...
```
is working just fine.

tbaden:
[...]

PR #634, opened 2019-05-26
[10.0][FIX] website_legal_page fix travis warn

tbaden:
issue https://github.com/OCA/website/issues/604
fix of:
************* Module website_legal_page.manifest
website_legal_page/manifest.py:6: [C8103(manifest-deprecated-key), ] Deprecated key "description" in manifest file

travis fails: https://github.com/OCA/website/issues/629

### DIFF
--- a/website_legal_page/__manifest__.py
+++ b/website_legal_page/__manifest__.py
@@ -5,7 +5,7 @@
 
 {
     'name': "Website Legal Page",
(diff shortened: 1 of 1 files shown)


pedrobaeza:
It seems all the builds are red. It would be better to fix that before fixing warnings.

tbaden:
@pedrobaeza I created an issue for this: travis fails: #629
planning to check/do this.

PR #646, opened 2019-06-04
[11.0][FIX] website_form_builder: Fix modal select view

sergio-teruel:
cc @Tecnativa
Before PR:
Selección_251

After PR:
Selección_250

@Yajo Can you review please?

### DIFF
--- a/website_form_builder/static/src/css/website_form_builder.less
+++ b/website_form_builder/static/src/css/website_form_builder.less
@@ -15,7 +15,7 @@
     }
 }
 
-/* Make modals with multiselect with a usable height */
+// Make modals with multiselect with a usable height
 body .modal.o_technical_modal {
     textarea, select[multiple] {
         &.form-control {
--- a/website_form_builder/static/src/css/website_form_builder.less
+++ b/website_form_builder/static/src/css/website_form_builder.less
@@ -15,10 +15,12 @@
     }
 }
 
-body .modal.o_website_modal {
+/* Make modals with multiselect with a usable height */
+body .modal.o_technical_modal {
     textarea, select[multiple] {
         &.form-control {
             height: auto;
+            background: None;
         }
     }
 }
--- a/website_form_builder/static/src/css/website_form_builder.less
+++ b/website_form_builder/static/src/css/website_form_builder.less
@@ -20,7 +20,7 @@ body .modal.o_technical_modal {
     textarea, select[multiple] {
         &.form-control {
             height: auto;
-            background: None;
+            background-image: none;
         }
     }
 }
--- a/website_form_builder/static/src/xml/widgets.xml
+++ b/website_form_builder/static/src/xml/widgets.xml
@@ -13,7 +13,7 @@
         <form>
             <div class="form-group">
                 <label for="field">Choose the new field for the form</label>
-                <select id="field" class="form-control" multiple="1" size="6">
+                <select id="field" class="form-control" multiple="" size="6">
                     <t t-foreach="widget.sorted_fields()" t-as="field_pair">
                         <option t-att-value="field_pair[1]">
                             <t t-esc="field_pair[0]"/>
--- a/website_form_builder/static/src/xml/widgets.xml
+++ b/website_form_builder/static/src/xml/widgets.xml
@@ -13,7 +13,7 @@
         <form>
             <div class="form-group">
                 <label for="field">Choose the new field for the form</label>
-                <select id="field" class="form-control" multiple="">
+                <select id="field" class="form-control" multiple="1" size="6">
                     <t t-foreach="widget.sorted_fields()" t-as="field_pair">
                         <option t-att-value="field_pair[1]">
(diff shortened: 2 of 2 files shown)


pedrobaeza (approved):

carlosdauden (approved):

rafaelbn (approved):
Tested in runbot :+1:  but Better wait until @Yajo review because I'm not sure if it was a change because a collateral effect or similar

yajo (approved):

website_form_builder/static/src/css/website_form_builder.less
```diff
@@ -22,3 +22,11 @@ body .modal.o_website_modal {
         }
     }
 }
+
+.modal.o_technical_modal {
```
yajo: I think that you need to fix line 18 instead of adding these.

yajo (changes requested):

website_form_builder/static/src/css/website_form_builder.less
chienandalu: I'd say is the right one

chienandalu (commented):
Just as suggestion, here:

https://github.com/OCA/website/blob/afe6e8762ce58f563f04a2a7b521cda830ba5984/website_form_builder/static/src/xml/widgets.xml#L16

A better control of height could be achieve using the `select` `size` parameter (numer of rows to display):

```html
<select id="field" class="form-control" multiple="1" size="6">
```

Tardo (approved):
LGTM, looks good! I agree with what Jairo says...

website_form_builder/static/src/css/website_form_builder.less
sergio-teruel: @Yajo How do you do in line 18?

sergio-teruel:
`<select id="field" class="form-control" multiple="1" size="6">`
This do not work

chienandalu:
> `<select id="field" class="form-control" multiple="1" size="6">`
> This not works

I meant in combination with `height: auto`

website_form_builder/static/src/css/website_form_builder.less
yajo: AFAICS you just have to change `o_website_modal` for `o_technical_modal` in [line 18 of this same file](https://github.com/OCA/website/pull/646/files#diff-5e0c8f09edd6ac67691c8881ecc2c7d9R18).

sergio-teruel:
Ok guys changes done... thanks!!

Tardo:
Tested on runbot, still looking good ;)

pedrobaeza:
@Yajo, update your review and we are good to go

website_form_builder/static/src/xml/widgets.xml
```diff
@@ -13,7 +13,7 @@
         <form>
             <div class="form-group">
                 <label for="field">Choose the new field for the form</label>
-                <select id="field" class="form-control" multiple="">
+                <select id="field" class="form-control" multiple="1" size="6">
```
yajo:
```suggestion
                <select id="field" class="form-control" multiple="" size="6">
```

`multiple` is a boolean attribute, whose valid values are either an empty string (`multiple=""`) or a string with the same name as the attribute itself (`multiple="multiple"`).

Check this out: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#Boolean_Attributes

website_form_builder/static/src/css/website_form_builder.less
```diff
@@ -15,10 +15,12 @@
     }
 }
 
-body .modal.o_website_modal {
+/* Make modals with multiselect with a usable height */
```
yajo:
```suggestion
// Make modals with multiselect with a usable height
```

In Scss, `/* these comments */` appear in the compiled CSS. That's why they are better suited for i.e. license comments (or just block comments as usual).

Since you are commenting over the code itself and it's a single line, it makes more sense to use `// Single line comment`.

It's not like this is a critical change, but I think you probably didn't know this little detail, so it's good to learn. :blush:

website_form_builder/static/src/css/website_form_builder.less
```diff
@@ -15,10 +15,12 @@
...
+body .modal.o_technical_modal {
     textarea, select[multiple] {
         &.form-control {
             height: auto;
+            background: None;
```
yajo: Let's be more specific to allow theming things such as bg color, etc. Also standard CSS is lowercase always:

```suggestion
            background-image: none;
```

yajo (changes requested):
Little details. :blush:

sergio-teruel:
Chances accepted

chienandalu (approved):

website_form_builder/static/src/xml/widgets.xml
yajo: This one is still missing.

sergio-teruel:
Suggestions accepted

yajo (approved):

PR #650, opened 2019-07-08
[ADD] Website Search module

gfcapalbo:
Adds a module to support website page search widget.
Structure easily extendible to other objects other than pages.


website_search/README.rst
```diff
@@ -0,0 +1,74 @@
...
+
+Usage
+=====
+
+To use this module:
+-------------------
+
+* Drag the widget in any website container
+* Save page
+* search for any term
```
hbrunn, line 49: or link to `/search`

website_search/README.rst
```diff
@@ -0,0 +1,74 @@
...
+   :alt: License: LGPL-3
+
+==============
+Website Search
+==============
+
+Module implements a website snippet that allows a global search on website pages.
+Structure allows for extendibility for searches in other resources like 
+blogs, attachments, or product descriptions.
+This module provides the ability to categorize blog posts.
```
hbrunn: ?

website_search/README.rst
```diff
@@ -0,0 +1,74 @@
...
+   * more objects searched: blogs, attachments, product descriptions...
+   * prerendering via cron job to speed up search on big high traffic deployments.
+
+
+Bug Tracker
+===========
+
+Bugs are tracked on `GitHub Issues <https://github.com/OCA/website/issues>`_.
+In case of trouble, please check there if your issue has already been reported.
+If you spotted it first, help us smash it by providing detailed and welcomed feedback.
```
hbrunn: use split readme

website_search/README.rst
```diff
@@ -0,0 +1,74 @@
...
+-------------------
+
+* Drag the widget in any website container
+* Save page
+* search for any term
+
+
+To extend this module:
+----------------------
+The module sctrucure is easily extendible by extending the website.search model
```
hbrunn: typo

website_search/static/src/js/website_search.js
```diff
@@ -0,0 +1,7 @@
+//-*- coding: utf-8 -*-
+//Copyright 2019 Therp BV <https://therp.nl>
+//License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+openerp.website_search = function(instance)
+{
+};
```
hbrunn: remove empty files

hbrunn (changes requested):

hbrunn (changes requested):
we'll also need a test that calls the controller and checks we do what we should.

This case is simple enough that you should be able to just mock the request object to return the current env and things should work. See https://github.com/OCA/crm/pull/277/files#diff-77d6ccec458e8681e0230c4e01d7c1abR50 for an example

website_search/README.rst
gfcapalbo: removed.

NL66278:
Please fix lint and test errors. Lint errors are mainly due to unused imports, should be easy to fix. Some test errors are not related to this module, but one is.

gfcapalbo:
@NL66278   @hbrunn linted ,  hopefully will be green after this. 
i just found an incompatibility between this search and website_livechat.
It seems that that module introduces a "page" that makes the preview renderer fail when website_livechat.channel_page   is included in search results.

The error is always a mistaken rendering of the publish button, as we render the preview in  QWEB.

uninstalling that module makes it work perfect.

Looking for what causes this. any ideas?

hbrunn:
it's not, because this repo uses an outdated .travis.yml. Please update this in a separate commit, the most important part is to use postgres 9.6.

Questions about errors are only answerable with a backtrace, preferrably with some links to code where you think it goes wrong

website_search/models/website_search.py
```diff
@@ -38,14 +38,14 @@ def _do_search_ir_view_page(self):
...
-                           id, page,
+                           id, page, name,
                            lower (array_to_string(
                                 xpath(
                                     '//*//text()', arch_db::xml), '\n'
                                 )) as view_text
                     from ir_ui_view
                     where type='qweb' and page='t') as table_text
-                where table_text.view_text ilike  %s """,
+                where table_text.view_text ilike  %s or name ilike %s""",
```
hbrunn: goode idea. You can also expand that by giving a higher priority to pages where the name matches vs a match in the text, because they probably are more relevant

gfcapalbo:
i have seen ronald's https://github.com/OCA/website/pull/654  , and i think the errors that are remaining are not of this code.

hbrunn:
[that's nonsense, ignore the below. I thought you were talking about another PR]
cherry pick his travis related commits, people tend not to review red prs. and you get helpful coverage info

gfcapalbo:
now includes @NL66278 's commit to silence unrelated errors and make green.

website_search/models/website_search.py
```diff
@@ -0,0 +1,69 @@
...
+
+        Take the search string from self.name, and add your results to
+        self.result_ids"""
+        for func in inspect.getmembers(self, lambda x: inspect.ismethod(
+                x) and x.__name__.startswith('_do_search_'), ):
+            func[1]()
+
+    @api.multi
+    def _do_search_ir_view_page(self):
+        # specific search method for a type of resource
```
NL66278, line 37: Make this a doc string

website_search/models/website_search.py
```diff
@@ -0,0 +1,69 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017-2019 Therp BV <https://therp.nl>.
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+# pylint: disable=missing-docstring
+import inspect
+from odoo import api, fields, models
+
+
+class WebsiteSearch(models.TransientModel):
+    '''
```
NL66278: doc strings use double quotes. As this is a new module, why not use black on it and minimize changes needed for future migrations to 13.0 and beyond?

website_search/models/website_search.py
```diff
@@ -0,0 +1,69 @@
...
+                    select
+                           id, page, name,
+                           lower (array_to_string(
+                                xpath(
+                                    '//*//text()', arch_db::xml), '\n'
+                                )) as view_text
+                    from ir_ui_view
+                    where type='qweb' and page='t') as table_text
+                where table_text.view_text ilike %s or name ilike %s""",
+            ('%' + (self.name or '') + '%', '%' + (self.name or '') + '%', )
```
NL66278: Repeated parameter value means code duplication. Pass the search value from a dictionary instead of a tuple:
```
...
where table_text.view_text ilike %(search_string)s or name ilike %(search_string)s""",
            {'search_string': self.name or ''}
)
```
You have to ask: does a search on an empty string make any sense? Or even a string of length 1? In case you make self.name required (and at least 1 non-whitespace char long), you do not need the `or ''`.

website_search/models/website_search.py
NL66278: And you can go wild by having a secondary ranking criterium based on how often the search string occurs in the text. Check this (and adapt the info) to see how you can make the count part of your SQL search: https://stackoverflow.com/questions/36376410/counting-the-number-of-occurrences-of-a-substring-within-a-string-in-postgresql

NL66278 (changes requested):
See other comments.

Tardo:
@gfcapalbo ping :)

PR #651, opened 2019-07-16
[FIX] website_lazy_load_image: HTML parsing

hugho-ad:
fromstring function.

Before this commit, the render_template has an issue with views retrieved
from a controller, like the wizard of event detail registration, the
the content was introduced into an html DOM, this leads a wrong modal
behavior.

Now the first attempt is with fromstring function if fails, then use
the original process with html function.

Before
image

After
image

Before
image

After
image

### DIFF
--- a/website_lazy_load_image/models/ir_ui_view.py
+++ b/website_lazy_load_image/models/ir_ui_view.py
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import api, models
-from lxml import etree
+import lxml
 
 
 class IrUiView(models.Model):
@@ -22,7 +22,7 @@ class IrUiView(models.Model):
         website_id = self.env.context.get('website_id')
         if website_id and not \
                 self.env['website'].browse(website_id).is_publisher():
-            html = etree.HTML(res)
+            html = lxml.html.fromstring(res.decode())
             imgs = html.xpath(
                 '//main//img[@src][not(hasclass("lazyload-disable"))]'
             ) + html.xpath(
@@ -32,5 +32,5 @@ class IrUiView(models.Model):
                 src = img.attrib['src']
                 img.attrib['src'] = self.LAZYLOAD_DEFAULT_SRC
                 img.attrib['data-src'] = src
-            res = etree.tostring(html, method='html')
+            res = lxml.etree.tostring(html, method='html', encoding='UTF-8')
         return res
--- a/website_lazy_load_image/models/ir_ui_view.py
+++ b/website_lazy_load_image/models/ir_ui_view.py
@@ -22,7 +22,7 @@ class IrUiView(models.Model):
         website_id = self.env.context.get('website_id')
         if website_id and not \
                 self.env['website'].browse(website_id).is_publisher():
-            html = lxml.html.fromstring(res.decode())
+            html = lxml.html.fromstring(res.decode('UTF-8'))
             imgs = html.xpath(
                 '//main//img[@src][not(hasclass("lazyload-disable"))]'
             ) + html.xpath(
--- a/website_lazy_load_image/tests/test_lazy_load_image.py
+++ b/website_lazy_load_image/tests/test_lazy_load_image.py
@@ -27,7 +27,7 @@ class LazyLoadTest(SavepointCase):
             'arch_base': arch_2,
             'mode': 'primary'
         })
-        arch_3 = 'content not wrapped'
+        arch_3 = '<t t-name="test"><span>content not wrapped</span></t>'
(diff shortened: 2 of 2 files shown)


hugho-ad:
@oscarolar could you  review this please

simahawk (changes requested):
Please check https://github.com/OCA/website/pull/571 and https://github.com/OCA/website/issues/551
If we change this should do it safely and only once. And we should have specific tests to cover the use cases that were broken.
In any case IMO we should not parse as HTML  something that is not.

hugho-ad:
@simahawk Thanks for your feedback!

I committed your changes proposed at https://github.com/OCA/website/pull/571#discussion_r300283047
I before passing res to html.fromstring method I decode res, I mean:
```python
html = ehtml.fromstring(res.decode())
```

What do you think?

website_lazy_load_image/models/ir_ui_view.py
```diff
@@ -2,7 +2,7 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
 from odoo import api, models
-from lxml import etree
+from lxml import html as ehtml, etree
```
simahawk: just import lxml use `lxml.etree` and `lxml.html`

website_lazy_load_image/models/ir_ui_view.py
```diff
@@ -22,7 +22,7 @@ def render_template(self, template, values=None, engine='ir.qweb'):
         website_id = self.env.context.get('website_id')
         if website_id and not \
                 self.env['website'].browse(website_id).is_publisher():
-            html = etree.HTML(res)
+            html = ehtml.fromstring(res.decode())
```
simahawk: hmm I don't understand: where is the check on whether this is or not an HTML source? :thinking:

simahawk (changes requested):

website_lazy_load_image/models/ir_ui_view.py
hugho-ad: as you mentioned, lxml.html.fromstring keep the content as it's if it's not an html content.
I just tested it

hugho-ad: Also, I didn't face any problem with robots.txt as mentioned on the issue https://github.com/OCA/website/issues/551

simahawk: ok, cool! Can you update tests [here](https://github.com/OCA/website/blob/11.0/website_lazy_load_image/tests/test_lazy_load_image.py) to avoid regressions?

hugho-ad:
@simahawk 
Done

Before
image
After
image


Before
image
image

website_lazy_load_image/tests/test_lazy_load_image.py
```diff
@@ -60,3 +77,21 @@ def test_disabled_render(self):
...
+        self.assertEqual(res, arch)
+
+    def test_encoding_render(self):
+        """Check content is correctly enconded"""
+        public_user_id = self.ref('base.public_user')
+        ui_view = self.env['ir.ui.view'].sudo(
+            public_user_id).with_context(website_id=self.website_id)
+        res = ui_view.render_template(self.view_4.id).decode('UTF-8')
+        arch = '<main><span>Teléfono, means phone</span></main>'
+        self.assertEqual(res, arch)
```
simahawk, line 97: Thanks! One last thing... 
Can you test rendering of txt?
Like, `self.env.ref('website.robots').render()` and verify the content is not HTML?

hugho-ad: Done

simahawk:
@hugho-ad pls check travis.
@tarteo can you pass by for a review?

hugho-ad:
Travis red job build no related to this
image


I just forced a travis rebuild

simahawk (approved):

hugho-ad:
@tarteo could you review please

website_lazy_load_image/models/ir_ui_view.py
```diff
@@ -22,7 +22,7 @@ def render_template(self, template, values=None, engine='ir.qweb'):
         website_id = self.env.context.get('website_id')
         if website_id and not \
                 self.env['website'].browse(website_id).is_publisher():
-            html = etree.HTML(res)
+            html = lxml.html.fromstring(res.decode())
```
moylop260: Using py2.7 for odoo 11.0 you can get a error here:
 - Screen Shot 2019-07-25 at 7 41 29 PM


Please, consider adds `.decode(UTF-8)`

moylop260 (approved):
👍 
Thanks!

moylop260:
@oscarolar 
Could you check it, please?

hugho-ad:
@Yajo  could you merge please

PR #653, opened 2019-07-26
[WIP][10.0][ADD] website_base_form_handler.

NL66278:
Provide a rock solid framework for loading and displaying website forms, validate the user input, redisplay errors and redirect on succesfull completion.

This will be a programmers tool, not to be confused with website_form_builder which is directed at the technical savvy end-user.

### DIFF
--- a/website_base_form_handler/README.rst
+++ b/website_base_form_handler/README.rst
@@ -1,0 +1,25 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+    :target: https://www.gnu.org/licenses/agpl-3.0-standalone.html
+    :alt: License: AGPL-3
+
+=============================================
+General support for programming website forms
+=============================================
+
+This module provides a base framework for handling website Forms. It contains
+all the logic of loading data, displaying a form, validating submitted data,
+redisplay on errors, going to a next step on success.
+
+Programmers of website forms can, with this module, concentrate on that
+which is specific for their form, and not on the always repeating boilerplate
+code.
+
+While the name website_form_handler would be most appropiate for this module,
+this might cause confusion with te existing website_form* modules, that are
+basically an enduser tool.
+
+=========================
+website_base_form_handler
+=========================
+
+Base module to handle forms in website.
--- a/website_base_form_handler/__init__.py
+++ b/website_base_form_handler/__init__.py
@@ -1,0 +1,4 @@
+# -*- coding: utf-8 -*-
+# Copyright 2019 Therp BV <https://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+from . import controllers
--- a/website_base_form_handler/__manifest__.py
+++ b/website_base_form_handler/__manifest__.py
@@ -7,7 +7,7 @@
     "author": "Therp BV,Odoo Community Association (OCA)",
     "license": "AGPL-3",
     "category": "website",
-    "summary": 'WWebsite Base Form Support',
+    "summary": 'Website Base Form Support',
     "depends": [
         'website',
     ],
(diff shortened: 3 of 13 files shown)


pedrobaeza:
@simahawk did you do something similar to this in https://github.com/OCA/website-cms?

NL66278:
Hi @pedrobaeza I know of the website-cms module. That is like the module I am proposing a programmers tool, but also with a quite different usecase. Basically the module from @simahawk tightly integrates a particular model to a display on the web. The module I am proposing does not have any limitations on the source of the data to be displayed on the form, does not do any automatic or fixed layout, and does not make any assumption on where data is stored after validation.

I extracted this code from a customer module where we had at first tried to use website-cms, but had to abandon that idea, because of the limitations we encountered.

pedrobaeza:
OK, as I don't know that modules, that's why I was asking to Simone.

NL66278:
Travis errors have nothing to do with this module...

simahawk:
> Hi @pedrobaeza I know of the website-cms module. That is like the module I am proposing a programmers tool, but also with a quite different usecase. Basically the module from @simahawk tightly integrates a particular model to a display on the web. The module I am proposing does not have any limitations on the source of the data to be displayed on the form, does not do any automatic or fixed layout, and does not make any assumption on where data is stored after validation.

Sorry, I think you misunderstood website-cms modules, in particular cms_form (or maybe the docs are not good enough :stuck_out_tongue: ). It does exactly what you seem to do here and even more. You are in control of _everything_. Is not tightly coupled to a model. You can build your own stand-alone form or wizard. You can hook and override storage, validators, loaders, templates, permissions, widgets. Really everything. By default they have pre-defined behaviours but you can customize all of them.

Maybe if you explain the use case you solved with your module I can tell how to achieve the same (likely) w/ the CMS suite. Let's talk about it :wink: 

For starting, can you elaborate on "because of the limitations we encountered"? Which limitations?

NL66278:
It might well be that I have not discovered all the possibilities, as I found it quite difficult to figure out how things worked. But everything I see has to do with linking forms to models. Anyway, not saying it is not possible, just that I could not find it in a reasonable amount of time, but one of the difficulties for me was how to use sudo to read models for public users.

simahawk:
> But everything I see has to do with linking forms to models.

Then I don't understand: do you want or not to read/write data from odoo models straight?
Once again, can you explain the whole use case of your forms to understand what you are trying to achieve?

>  [...] but one of the difficulties for me was how to use sudo to read models for public users.

That's super-easy: just pass `main_object` already sudo-ed. 
The form is initialized at controller level here https://github.com/OCA/website-cms/blob/11.0/cms_form/controllers/main.py#L74.
As you see you pass explicitly the record to play with.

Then you have 2 options: pass it already sudo-ed from the controller (if you want different behavior by controller) or just override the init of the main object https://github.com/OCA/website-cms/blob/11.0/cms_form/models/cms_form_mixin.py#L135 (as you see form init sets the main object).
I just did a quick test w/ this stupid form:

```py
class PublicForm(models.AbstractModel):

    _name = 'cms.form.res.partner'
    _description = 'Public edit form for partners'
    _inherit = 'cms.form'
    _form_model = 'res.partner'
    _form_model_fields = (
        'name',
    )

    def form_init(self, request, main_object=None, **kw):
        form = super().form_init(request, main_object=main_object, **kw)
        form.main_object = main_object.sudo() if main_object else None
        return form
```
and it works like a charm. Of course, it's pretty dangerous especially w/ partners :sweat_smile: .

NL66278:
@simahawk Thank you for your reply. I see there is more possible with cms_form then I discovered previously. Busy finishing a customer project now, but will think about what can be done with examples or documentation.

PR #655, opened 2019-07-30
[12.0][FIX] Make it possible to edit pages created by "website_legal_page"

katyukha:
#### Before this PR:
After installing module 'website_legal_page' we have 3 pages (terms of use, privacy policy, legal advice). When you try to edit all of them, then only (randomly choosen) one (or two) page(s) is editable.

#### After this PR:
All three pages are editable.

#### Migration Notes
If pages introduced by `website_legal_page` addons were modified before this PR, then there is website-specific views exists (in my case only one view) and usualy that views are incorrect. The simple way to fix them is to remove website-specific view(s) and create new-ones that could use correct templates.

### DIFF
--- a/website_legal_page/views/website_legal.xml
+++ b/website_legal_page/views/website_legal.xml
@@ -10,13 +10,7 @@
         </xpath>
     </template>
 
-
-    <record id="advice" model="ir.ui.view">
-        <field name="name">Legal Advice</field>
-        <field name="type">qweb</field>
-        <field name="key">website.advice</field>
-        <field name="arch" type="xml">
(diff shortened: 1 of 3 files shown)


pedrobaeza:
Seems, OK, but 2 things:

* Squash all the commits together.
* Don't change version number. The bot will do it on merging.

katyukha:
@pedrobaeza Hello, thanks for fast feedback. Your code updated according to your comments. Also i have added migration notes to PR description.

pedrobaeza (approved):

Tardo (approved):
LGTM :+1:

PR #657, opened 2019-08-10
[ADD] website_portal_comment_attachment

luffah:
Hello, this PR add module ~~website_portal_comments_attachments~~  website_portal_comment_attachment.
It allows portal users to add one attachment with chatter comments.
Currently, the choice is to allow only one comment in order to have distinct message for each attachment from portal.


rafaelbn:
@fgi-odoo could you add this in v13? For me is very frustrating when in odoo.com I cannot attach in portal files. As v13 is focused in retail this would be very needed for final customer

fgi-odoo:
Yes it is already done in master I think. Or going to be.

luffah:
Force pushed to pass jslint. I was using `let` keyword with statement  `for (let i of items) { ... }` (es6).

eLBati (changes requested):
Hi @luffah it seems that, after the first message, it's not possible anymore to attach files:

so

pedrobaeza:
@luffah please avoid plurals in module name according guidelines:

https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#modules

flotho:
Hi @luffah , thanks for the work already done.

I've tested as admin user on FF and Chrome successfully. Yet with a portal user the dialog box for choosing the attachment doesn't appear.
Any access right issue ?
BTW, nothing special in the js console

luffah:
FYI, i hope to have time to look at these issues in 2 or 3 weeks to 
- [x]  [rename](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#modules) module to "website_portal_comment attachment" / 2020-03-06
- [ ]  check/solve [issue on adding more than 1 attachment](https://github.com/OCA/website/pull/657#pullrequestreview-306331762)
- [ ] consolidate [access rights](https://github.com/OCA/website/pull/657#issuecomment-546591973)

eLBati:
@luffah https://github.com/OCA/website/pull/657#issuecomment-546591973 https://github.com/OCA/website/pull/657#pullrequestreview-306331762 I think they are both the same problem: I noticed that sometimes `$('form.o_portal_chatter_composer_form')` is empty (so the button doesn't wotk) and sometimes is not.
So something wrong about js loading sequence.
Any idea?

eLBati:
@luffah please have a look at https://github.com/osiell/website/pull/1

flotho:
ping @luffah

luffah:
pong

luffah:
@flotho retesting with eLBati mods, i got indeed an access right issue "403 : Forbidden" with an user.

luffah:
I get why there is an access error : `ir.attachment` is linked to an `ir.model` object. This right is not in Odoo base for a portal user.

It is probably not a security issue to add this line.
```csv
access_ir_model_portal,ir.model portal,base.model_ir_model,base.group_portal,1,0,0,0
```

flotho:
Hi @luffah , I'm not sure that it is a good idea. 
First, if there is already file attachments that  portal users are able to display. So they already have the read access on attachments. 
base.access_ir_attachment_all already open the read property for portal user.
additionally the rule ir_model_all implemented by odoo is placed only on the employee group
image

so maybe if you want to do this you'll have to combine it with an ir.rule for protal user with a filter like this [('name', '=', 'ir.attachment')]

otherwise, portal users will be able to scann all the available model through API.

What is your point of view on this ?

Regards

luffah:
I would like to have a better solution...

Details about the issue:
```
Traceback (most recent call last):
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/addons/base/models/ir_http.py", line 203, in _dispatch
    result = request.dispatch()
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/http.py", line 835, in dispatch
    r = self._call_function(**self.params)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/http.py", line 346, in _call_function
    return checked_call(self.db, *args, **kwargs)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/service/model.py", line 97, in wrapper
    return f(dbname, *args, **kwargs)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/http.py", line 339, in checked_call
    result = self.endpoint(*a, **kw)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/http.py", line 941, in __call__
    return self.method(*args, **kw)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/http.py", line 519, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/odoo/parts/website/website_portal_comment_attachment/controllers/mail.py", line 19, in portal_chatter_post
    res_model, res_id, message, **kw)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/http.py", line 519, in response_wrap
    response = f(*args, **kw)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/addons/portal/controllers/mail.py", line 74, in portal_chatter_post
    _message_post_helper(res_model, int(res_id), message, **kw)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/addons/portal/controllers/mail.py", line 63, in _message_post_helper
    **kw)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/addons/rating/models/mail_thread.py", line 14, in message_post
    message = super(MailThread, self).message_post(**kwargs)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/addons/mail/models/mail_thread.py", line 2052, in message_post
    attachment_ids = self._message_post_process_attachments(attachments, kwargs.pop('attachment_ids', []), values)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/addons/mail/models/mail_thread.py", line 1908, in _message_post_process_attachments
    new_attachment = IrAttachment.create(data_attach)
  File "<decorator-gen-163>", line 2, in create
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/api.py", line 430, in _model_create_single
    return create(self, arg)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/addons/website/models/ir_attachment.py", line 25, in create
    return super(Attachment, self).create(vals)
  File "<decorator-gen-41>", line 2, in create
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/api.py", line 451, in _model_create_multi
    return create(self, [arg])
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/addons/base/models/ir_attachment.py", line 517, in create
    return super(IrAttachment, self).create(vals_list)
  File "<decorator-gen-3>", line 2, in create
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/api.py", line 452, in _model_create_multi
    return create(self, arg)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/models.py", line 3596, in create
    fields[0].determine_inverse(batch_recs)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/fields.py", line 1117, in determine_inverse
    getattr(records, self.inverse)()
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/addons/base/models/ir_attachment.py", line 227, in _inverse_datas
    super(IrAttachment, attach.sudo()).write(vals)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/models.py", line 3356, in write
    self.recompute()
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/models.py", line 5270, in recompute
    vals = {n: rec[n] for n in ns}
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/models.py", line 5270, in <dictcomp>
    vals = {n: rec[n] for n in ns}
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/models.py", line 5117, in __getitem__
    return self._fields[key].__get__(self, type(self))
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/fields.py", line 988, in __get__
    value = record.env.cache.get(record, self)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/api.py", line 1043, in get
    return value.get() if isinstance(value, SpecialValue) else value
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/api.py", line 1086, in getter
    raise exception
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/fields.py", line 1057, in compute_value
    self._compute_value(record)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/fields.py", line 1043, in _compute_value
    getattr(records, self.compute)()
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/addons/base/models/ir_attachment.py", line 53, in _compute_res_model_name
    model = self.env['ir.model'].search([('model', '=', record.res_model)], limit=1)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/models.py", line 1561, in search
    res = self._search(args, offset=offset, limit=limit, order=order, count=count)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/models.py", line 4098, in _search
    self.sudo(access_rights_uid or self._uid).check_access_rights('read')
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/models.py", line 3026, in check_access_rights
    return self.env['ir.model.access'].check(self._name, operation, raise_exception)
  File "<decorator-gen-23>", line 2, in check
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/tools/cache.py", line 93, in lookup
    value = d[key] = self.method(*args, **kwargs)
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/addons/base/models/ir_model.py", line 1278, in check
    raise AccessError(msg % msg_params)
odoo.exceptions.AccessError: ("Désolé, vous n'êtes pas autorisé à accéder à ce document. Seuls les utilisateurs avec les niveaux d'accès suivants sont actuellement autorisés à faire cela:\n- Administration/Access Rights\n\t- User types/Internal User\n\n(Modèle de document: ir.model) - (Operation: read, User: 7)", None)
```
When creating an 'ir.attachment',  it computes the res_model_name. This is the (at least one) source of the issue.

The idea of adding the read right for portal user comes from the issue raised by this method.
An alternate way is to monkey patch the res_model_name compute function by adding sudo() in the request that search model_name.             
`model = self.env['ir.model'].sudo().search([...`

```python
class IrAttachment(models.Model):
    _inherit = 'ir.attachment'

    @api.depends('res_model')
    def _compute_res_model_name(self):
        for record in self:
            if record.res_model:
                model = self.env['ir.model'].sudo().search([
                    ('model', '=', record.res_model)], limit=1)
                if model:
                    record.res_model_name = model[0].name
```


(similarly we can  monkey patch `portal_chatter_post()` from portal module — instead ir_attachment — to get the id of the message created in order to add attachment)

(still, it is monkey patching...)

Not used yet to cheat with ir.rules. It can do the job to allow to access ir.model ?
As the log saids, it fails on raise exception on `ir.model.access` `.check('ir.model', 'read')` before checking `ir.rules`.
```
  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/models.py", line 4098, in _search
    self.sudo(access_rights_uid or self._uid).check_access_rights('read')
    ...
    self._apply_ir_rules(query, 'read')

  File "/home/odoo/odoo/parts/OCB-e5345c139f65369a3b3f2f7c317ffcc474276dc7/odoo/models.py", line 3026, in check_access_rights
    return self.env['ir.model.access'].check(self._name, operation, raise_exception)```

**Comment by flotho** [comment]
Hi @luffah 
thanks for sharing your tests.
Monkey patch is probably the worst way to do this, too intrusive IMHO.
Did you try to add both ir.access.rule and ir.rule :
* the first like you already did https://github.com/OCA/website/pull/657#issuecomment-595557405
* The second by a rule like [('name', '=', 'ir.attachment')] on ir.model

**Comment by luffah** [comment]
That's why it is called monkey patching. Just want to expose where are the locks.

> add both ir.access.rule and ir.rule 

i don't know ir.access.rule.  Where to set it ? (if it is different from `ir.model.access.csv`)

Sorry, i don't understand how to apply your solution. As i tried to explain with logs, the issue i have found is related to `read` access to `ir.model` in the a compute function in `ir.attachment` (which use `search(...)` of `ir.model`). I'll try to ask to colleagues.

**Comment by flotho** [comment]
Hi @luffah 

In your Web UI interface you have 2 menus about security
one for ir.access.rules that you know and another one for ir.rule.
The first is allowing access to the entire table, the second is filtering. Thanks to the second you could define multi company rules, own leads user acess etc.
Try a rule there and export the conf in a security.xml file like there : https://github.com/OCA/OCB/blob/12.0/addons/sale/security/sale_security.xml#L41

**Comment by luffah** [comment]
Adding a rule to filter models seems to work, although i'm not an odoo security export.

The main problem is to find a domain that allow to add attachment to all accessible  messages fields.
With  
```xml
        <field name="domain_force">[('is_mail_thread', '=', True)]</field>
```
we got restriction to
```python
>>> self.env['ir.model'].sudo(pu).search([]).mapped('name')
['Analytic Account', 'Bank Statement', 'Consent of data processing', 'Contact', 'Data processing activities', 'Discussion Channel', 'Email Thread', 'Event', 'Helpdesk Ticket', 'Invoice', 'Lead/Opportunity', 'Mail Blacklist', 'Payments', 'Product', 'Product Template', 'Project', 'Sale Order', 'Sales Channels', 'Task']
```

luffah:
In case of IR Rules above are incorrects : instead of changing security rules, i just noticed that the compute access_right issue can be overridden.
```python
class IrAttachment(models.Model):
    _inherit = 'ir.attachment'
    res_model_name = fields.Char(compute_sudo=True)
```

flotho:
Hi @luffah 

Great and elegant way to deal with your issue, congrats !
You rocks, I succeed in uploading attachments. 
I think you could add in the roadmap that html editor with inline attachment preview could be usefull but it's not required

flotho (approved):
Successfully tried there 
image

luffah:
@flotho : sorry, i added the message while i was working on another project to got an advice. But it wasn't committed. Can you retry the test ?

(given i force pushed — you'll have to 
`git fetch origin 12.0-website_portal_comments_attachments` and
 `git reset --hard origin/12.0-website_portal_comments_attachments`
where origin is the name of the remote)

flotho:
Hi @luffah , 
I made a new test and it works perfectly.
I think we should add in the roadmap that preview is the next step in this chatter.

flotho:
@eLBati , your comments have been corrected here, any more review to merge this awesome PR ?

eLBati:
I would mention in readme the potential security issues of letting anyone  upload files to your server

luffah:
> I would mention in readme the potential security issues of letting anyone upload files to your server

Do you have a test case(s) ?

My tests covered only the case on portal users (logged peoples). 

If there is a security issue, we shall propose restrictions (at least in future — options to reject bigfiles and any filetype).

luffah:
Is this disclaimer ok ?
```
Warning : letting external peoples upload files raises security risks.
Be sure to take measure against malwares, server overload and illegal content.
Too, some files types may be misinterpreted by Odoo or you web browser.
```

eLBati:
@luffah ok for me, thanks

luffah:
Pushed

eLBati:
@luffah it seems the last 2 commits are not recognized by github. Could you check which email you used for those commits?
Thanks

luffah:
> @luffah it seems the last 2 commits are not recognized by github. Could you check which email you used for those commits?
> Thanks

Changed author email in last commits

luffah:
codev raise a strange error  next to that author email change.
~~ New file website_crm_recaptcha/hooks.py` ; but it is not related to my change.~~
_**Edit**: i changed capitalization in author name,  and codecov no more raise the issue._

luffah:
Note: apriori no need to make a version of this module for v13.
It is in core.

website_portal_comment_attachment/security/ir.model.access.csv
```diff
@@ -0,0 +1 @@
+id,name,model_id:id,group_id:id,perm_read,perm_write,perm_create,perm_unlink
```
Tardo: :fire:

website_portal_comment_attachment/controllers/mail.py
```diff
@@ -0,0 +1,19 @@
...
+
+    @http.route(['/mail/chatter_post'], type='http', methods=['POST'],
+                auth='public', website=True)
+    def portal_chatter_post(self, res_model, res_id, message, **kw):
+        attachment = kw.pop('attachment', None)
+        if attachment:
+            attachments = [(attachment.filename, attachment.read())]
+            kw['attachments'] = attachments
+
+        return super(PortalChatterWithAttachments, self).portal_chatter_post(
```
Tardo:
```suggestion
        return super().portal_chatter_post(
```

website_portal_comment_attachment/static/src/js/portal_chatter.js
```diff
@@ -0,0 +1,67 @@
...
+    'use strict';
+    var core = require('web.core');
+    var ajax = require('web.ajax');
+    var PortalChatter = require('portal.chatter').PortalChatter;
+
+    PortalChatter.include({
+        _loadTemplates: function(){
+            var res = this._super.apply(this, arguments);
+            ajax.loadXML('/website_portal_comment_attachment/static/src/xml/portal_chatter.xml', core.qweb);
+            return res;
```
Tardo: ```res``` is a deferred object, use it:
```suggestion
            return this._super.apply(this, arguments).then(function () {
                ajax.loadXML('/website_portal_comment_attachment/static/src/xml/portal_chatter.xml', core.qweb);
            }
```

website_portal_comment_attachment/static/src/js/portal_chatter.js
```diff
@@ -0,0 +1,67 @@
...
+            if (val == "") {
+                btn_add.prop("disabled", true);
+            }
+            else {
+                btn_add.prop("disabled", false);
+            }
+        },
+        init: function () {
+            this._super.apply(this, arguments);
+            this.events["click .o_portal_chatter_attachments_add"] = '_onClickAttachmentsAdd';
```
Tardo: Can't assign this before initialization?
Something like this: https://github.com/OCA/web/blob/13.0/web_responsive/static/src/js/web_responsive.js#L101

website_portal_comment_attachment/static/src/js/portal_chatter.js
```diff
@@ -0,0 +1,67 @@
...
+        },
+        _onTextareaMessageComposerKeyup: function(ev){
+            var val = ev.target.value;
+            var btn_add = $('.o_portal_chatter_attachments_add');
+            if (val == "") {
+                btn_add.prop("disabled", true);
+            }
+            else {
+                btn_add.prop("disabled", false);
+            }
```
Tardo:
```suggestion
            btn_add.prop("disabled", ev.target.value === "");
```

website_portal_comment_attachment/static/src/js/portal_chatter.js
```diff
@@ -0,0 +1,67 @@
...
+                item.textContent = input[0].files[i].name;
+                item.prepend(fileindicator);
+                item.appendChild(abort_btn);
+
+                list.append(item);
+                btn.addClass('d-none');
+            }
+        },
+        _onTextareaMessageComposerKeyup: function(ev){
+            var val = ev.target.value;
```
Tardo:
```suggestion
```

Tardo (changes requested):
Thanks for the work!

website_portal_comment_attachment/security/ir.model.access.csv
luffah: No new model added in this module.
Access right to the attachment for portal user is managed in code because it is too much complex to manage in ir.model.access.

But indeed i found no ''empty'' ir.model.access file in other modules.

Note : to delete

website_portal_comment_attachment/static/src/js/portal_chatter.js
luffah: Can you add the final parenthesis ? (in order i add your suggestion to the batch commit. ^^)

luffah: Tested : defering `ajax.loadXML` (after _loadTemplate), avoid the view to be used...

website_portal_comment_attachment/security/ir.model.access.csv
luffah: deleted in last commit (force pushed)

eLBati:
Hi @luffah it seems too many commits appeared in this PR.
Could you clean them?

eLBati:
Also, I think commit author `luffah@users.noreply.github.com` should be fixed

luffah:
Replaced all by only one commit.

luffah:
> Also, I think commit author `luffah@users.noreply.github.com` should be fixed

Is there a rule for that ? It is the same fake email i use on all oca repositories.

eLBati:
@luffah sorry, I saw the grey user icon so I thought it was set by mistake.
No problem then

eLBati (approved):
Thanks

marcelsavegnago (approved):

website_portal_comment_attachment/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "version": "12.0.1.0.0",
+    "category": "Website",
+    "website": "http://github.com/OCA/website",
+    "author": "Osiell, "
+              "Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "website",
```
eLBati: Is website dep needed? Wouldn't `portal` be enough?

eLBati: I confirm `portal` is enough. I would change this consequently

eLBati (changes requested):

HaraldPanten (approved):
LGTM 👍 

Even though, I see another potential security issue. IMHO the module should have the following IMP settings:

- Internal users could limit the max size of the files.
- Internal users could limit, the max size of the total files attached to a document.
- Internal users could define which format was accepted.
- Internal users could define which portal users have permissions to attach files, for example:

1- Imagine that you have several clients and each one has his/her own portal user to check their documents. Let's say Client A (portal user A), Client B (portal user B) and Client C (portal user C),
2- You may not want that all portal users have rights to upload files in your server. Maybe you just want that only the portal user from client A has rights enough.
3- I'd propose a checkbox (or similar) for portal users that could be set, depending on if you want that this portal user has permissions (or not) for uploading files.

What do you think?

website_portal_comment_attachment/__manifest__.py
luffah: true (updated to 'portal')

eLBati (approved):
@luffah thanks.

@HaraldPanten I think your suggestions are valid, we could add them to ROADMAP, what do you think?

HaraldPanten:
> @luffah thanks.
> 
> @HaraldPanten I think your suggestions are valid, we could add them to ROADMAP, what do you think?

As a "nice to have". It's OK for me 👍

eLBati:
Is this mergeable?
@Tardo is it OK for you?

ferranmateu (approved):

website_portal_comment_attachment/static/src/js/portal_chatter.js
```diff
@@ -0,0 +1,65 @@
...
+            list.empty();
+            btn.removeClass('d-none');
+            for (var i=0; i<input[0].files.length; i++) {
+                var fileindicator = document.createElement('i');
+                fileindicator.className = 'fa-fw fa fa-file pr8';
+
+                var abort_btn = document.createElement('button');
+                abort_btn.className = 'btn';
+                abort_btn.textContent = '×';
+                abort_btn.title = 'Remove attachment';
```
Tardo, line 35:
```suggestion
                abort_btn.title = _t('Remove attachment');
```

website_portal_comment_attachment/static/src/js/portal_chatter.js
```diff
@@ -0,0 +1,65 @@
+odoo.define('website_portal_comment_attachment.chatter', function (require) {
+    'use strict';
+    var core = require('web.core');
+    var ajax = require('web.ajax');
+    var PortalChatter = require('portal.chatter').PortalChatter;
+
```
Tardo, line 6:
```suggestion
    var _t = core._t;
```

website_portal_comment_attachment/static/src/js/portal_chatter.js
```diff
@@ -0,0 +1,65 @@
...
+                list.append(item);
+                btn.addClass('d-none');
+            }
+        },
+        _onTextareaMessageComposerKeyup: function(ev){
+            var btn_add = $('.o_portal_chatter_attachments_add');
+            btn_add.prop("disabled", ev.target.value == "");
+        },
+        events: _.extend(
```
Tardo, line 53: This is usually placed in the first lines.

Tardo (changes requested):

website_portal_comment_attachment/static/src/js/portal_chatter.js
luffah: Is using directly core._t("Remove attachement") ok ?
If no, shall the same thing be done with `core.qweb`  ?

flotho:
Any chance to have a runboat instance to test this @sbidoul @pedrobaeza

pedrobaeza:
A rebase/push should serve

sbidoul:
This PR predates runboat. A rebase or push should trigger a build.

flotho:
ok thanks
Hi @guilhom34 could you fork and propose a new PR for this one so that runboat will help. Please also integrate comments form othe reviwers

marcelsavegnago:
Hello @luffah. Can you rebase please?

marcelsavegnago:
@luffah I created a new PR just for testing purposes and the functional tests are ok for me. #910

marcelsavegnago:
@luffah can you rebase please ?

luffah:
@marcelsavegnago  Sorry, i m full time on a completely different subject.

Feel free to take the lead on this subject.

PR #658, opened 2019-08-12
[11.0] website menu permission easy maintain

NL66278:
This consists of two parts.
- a migration done by @gfcapalbo of the 10.0 module by @simahawk 
- an adaption to ensure just installing the module does not take away access rights for many existing users. Only after consciously setting groups on a menu well permissions change. To edit this easily, the editing of website menu's has been added to the website configuration menu.

The record-rule itself remains the same. Basically as also in the PR made for 8.0  by @hbrunn for this purpose.


hbrunn:
looks good, and when you've changed the tests I think it will be green too. Let's then finally merge a version of this, the 8 and 10 version is hanging around like forever already

NL66278:
For some reason travis cannot setup database connection:
```
Creating instance:

createdb: could not connect to database template1: could not connect to server: No such file or directory

	Is the server running locally and accepting

	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?

Using previous openerp_template database.
```

hbrunn:
you need to update this branch's .travis.yml. Best adapt to the current version in [MQT](https://github.com/OCA/maintainer-quality-tools/blob/master/sample_files/.travis.yml) - the crucial part here is to set a postgres version

hbrunn (approved):

website_menu_permission/README.rst
```diff
@@ -0,0 +1,83 @@
...
+   :local:
+
+Usage
+=====
+
+To use this module, you need to:
+
+#. Go to Website ==> Configuration ==> Website menu's.
+#. Edit the menu that you want to limit to certain groups, and add
+   the names of the groups that should have access.
```
rafaelbn, line 42: This explanation is confusing because:

- In Website -> Configuration there are 2 different menu entries for the same model website.menu
- You need developer mode in order to see "Website menu's" entry. The other entry "Menu" you can see without developer mode.

I think is enough one menu entry and not two.

NL66278: I only have one menu entry for website menu's. And I am in developer mode. If you see a second option for website menu's can you check from what module that is installed?

hbrunn: that's runbot installing all modules from the repository. I see xmlid `website_multi_theme.action_configure_menu_website`

website_menu_permission/tests/test_menu_perm.py
```diff
@@ -0,0 +1,77 @@
+# Copyright 2017 Simone Orsi.
+# Copyright 2019 Therp BV <https://therp.nl>.
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+import odoo.tests.common as test_common
+
+
+class TestMenuPerm(test_common.TransactionCase):
+
+    def setUp(self):
```
simahawk: please use SavepointCase and move `setUp` to `setUpClass` and replace the env like this:

```py
cls.env = cls.env(context=dict(cls.env.context, tracking_disable=True, no_reset_password=True))
```
It's going to speed up tests a lot (actually we should use this approach for all tests :man_teacher: )

website_menu_permission/security/record_rules.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0"?>
+<odoo>
+
+  <record id="website_menu_access" model="ir.rule">
+    <field name="name">website_menu_permission group access</field>
+    <field name="model_id" ref="model_website_menu"/>
+    <field name="domain_force">['|',('group_ids','in', [g.id for g in user.groups_id]), ('group_ids','=',False)]</field>
```
simahawk: this one can use `user.groups_id.ids` instead of the loop

simahawk (changes requested):
@NL66278 @gfcapalbo thanks for taking care of this :)
Mostly ok, just some small things :)

NL66278:
@simahawk Changes done!

simahawk (approved):
LG, can we squash?

NL66278:
@simahawk I squashed my commits, so that leaves 2.

simahawk:
@rafaelbn fine for you?
@NL66278 what's the plan for website_menu_by_user_status?
The original plan - from my POV - was to make it dependant on this one. Is this still the case?
There are a bunch of PRs that probably need some update or references https://github.com/OCA/website/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aopen+status+

NL66278:
@simahawk I think the module website_menu_by_user_status does not add really anything, that cannot be done with website_menu_permission. But for some people it might be easier to just have the choice betwee users that are logged in or not. So I think the ..user_status module could be rebased om --.permission, doing away with any overlaps. But that can be done/proposed after the current PR has been merged.

simahawk:
@NL66278 yeah, that was the original plan, also to keep backward compat: *_status module should rely on this one and just have a flag that sets the groups automatically.

PR #660, opened 2019-08-19
[FIX] website_snippet_big_button: Wrap big buttons content

yajo:
Without this patch, a big button with a lot of content will fail to
render properly in phones.

@Tecnativa


pedrobaeza (approved):
Fast-tracking this as bugfix

PR #661, opened 2019-08-26
[FIX] website_form_recaptcha: Fix loading

yajo:
Added a dependency. Without it, you get:

    form_builder_send is undefined

@Tecnativa

### DIFF
--- a/website_form_recaptcha/static/src/js/field_recaptcha.js
+++ b/website_form_recaptcha/static/src/js/field_recaptcha.js
@@ -7,6 +7,7 @@ odoo.define('website_form_recaptcha.recaptcha', function (require) {
 
     var ajax = require('web.ajax');
     var snippet_animation = require('website.content.snippets.animation');
+    require("website_form.animation");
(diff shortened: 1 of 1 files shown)


website_form_recaptcha/static/src/js/field_recaptcha.js
```diff
@@ -7,6 +7,7 @@ odoo.define('website_form_recaptcha.recaptcha', function (require) {
 
     var ajax = require('web.ajax');
     var snippet_animation = require('website.content.snippets.animation');
+    require("website_form.animation");
```
pedrobaeza, line 10: Don't you assign it to any variable?

simahawk: it's just declaring a dependency and in any case it does not expose anything (unfortunately).

yajo: No, this module just adds an animation to the registry. We get it from there below.

simahawk (approved):
makes sense

rafaelbn (approved):

PR #663, opened 2019-08-27
[DEL] website_blog_mgmt: Remove module

yajo:
This module is not needed. Remove code to avoid confusions (I just got confused today).

See https://github.com/OCA/OpenUpgrade/pull/1656.

@Tecnativa


pedrobaeza:
Merging any way as the branch seems broken before this and this is only removing deprecated non migrated module.

PR #665, opened 2019-09-12
[FIX] website_no_crawler: Revert on uninstall

yajo:
Before this patch, uninstalling the addon didn't revert its behavior.

@Tecnativa TT19451

### DIFF
--- a/website_no_crawler/views/disable_robots.xml
+++ b/website_no_crawler/views/disable_robots.xml
@@ -1,8 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <odoo>
-    <template id="website.robots">
+    <template id="robots" inherit_id="website.robots">
+        <xpath expr="." position="replace">
+<t>
 User-agent: *
 Disallow: /
 Sitemap: <t t-esc="url_root"/>sitemap.xml
+</t>
+        </xpath>
     </template>
 </odoo>


carlosdauden (approved):

website_no_crawler/views/disable_robots.xml
```diff
@@ -1,8 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <odoo>
-    <template id="website.robots">
+    <template id="robots" inherit_id="website.robots" primary="True">
```
pedrobaeza: Why `primary="True"`? This will duplicate applicable views and won't be used at all.

yajo: Because the view is no XML but TXT, how do you replace a node there?

pedrobaeza: But primary is not for that AFAIK.

rafaelbn (approved):
:+1:

website_no_crawler/views/disable_robots.xml
rafaelbn: Please solve this discussion :smile:

yajo: Yes sorry. AFAIK primary serves for the purpose of replacing a view, so `website.robots` now resolves to this view instead of the other as the base.

This way, when the module is uninstalled and this `website_no_crawler.robots` view is removed, `website.robots` recovers its primary status and works as before.

At least, that's the purpose of this patch.

pedrobaeza: Then you must ensure that the priority of this view is smaller than the other one, or it won't "resolve" to this one. Any way, are you sure the XML-ID is not hardcoded and this one it's in fact used?

yajo: > are you sure the XML-ID is not hardcoded and this one it's in fact used?

Since v11, the important part is the `key` of the view, not the XMLID.

Anyway, I'm rebuilding the runbot. I think it's better to test this before continuing. I'll test and report the results. :blush:

yajo: You were right! This wasn't working properly.

It turns out you can replace text nodes and, to avoid a `<p>` tag automatically being printed, you have to wrap it in a `<t>` tag.

Well, it's working now. And it uninstalls properly.

pedrobaeza (approved):

pedrobaeza:
Please forward-port it to v12

PR #666, opened 2019-09-26
[FIX] website_blog_excerpt_img: Display OG image in latest posts snippet

yajo:
When somebody drags and drops the "Latest Posts - Lists" snippet anywhere in the website, if this module is installed, it is expected to be using the opengraph image, just like the main blog posts list view.

Before:
Captura de pantalla de 2019-09-26 10-54-31

After:
Captura de pantalla de 2019-09-26 10-55-10

@Tecnativa TT19639

### DIFF
--- a/website_blog_excerpt_img/__manifest__.py
+++ b/website_blog_excerpt_img/__manifest__.py
@@ -22,5 +22,6 @@
     ],
     "data": [
         "templates/blog.xml",
+        "templates/snippets.xml",
     ],
 }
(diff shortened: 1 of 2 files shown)


pedrobaeza:
But isn't this needed to go to Odoo itself?

yajo:
Odoo doesn't support an image that is not the cover image, anywhere, so it's not like they're missing something here.

There is a real bug in their code: the other snippet, the one that says "Latest posts - Big images", but I'm not targeting those because IMO they only make sense with a cover image. However, when there's no cover, you cannot see the content because of the white background.

pedrobaeza (approved):
OK, let's put this then

Tardo (approved):
LGTM :+1: tested on runbot

PR #669, opened 2019-10-09
[12.0][IMP] website_snippet_country_dropdown: custom input field classes

ernestotejeda:
Allow to set custom classes to input tag with id 'no_country_field'

Cc @Tecnativa

### DIFF
--- a/website_snippet_country_dropdown/views/snippets.xml
+++ b/website_snippet_country_dropdown/views/snippets.xml
@@ -22,6 +22,7 @@
 
 <template id="country_dropdown" name="Country Dropdown">
     <t t-set="default_value" t-value="default_value or ''"/>
+    <t t-set="no_country_field_extra_classes" t-value="no_country_field_extra_classes or ''"/>
     <t t-set="countries" t-value="request.env['res.country'].search([])"/>
     <t t-set="default_country" t-value="default_country or countries.search([('code', '=', default_value[:2])]) or request.env.user.country_id or website.company_id.country_id or countries[:1]"/>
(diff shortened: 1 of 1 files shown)


yajo (commented):
Is this really needed?

Downstream, wherever you use this snippet, can't you just call it like this?:

```xml
<div class="my-class">
  <t t-call="website_snippet_country_dropdown.country_dropdown"/>
</div>
```

Then, if you're styling the input:

```css
.my-class .js_no_country_field {
  /* Your style here */
}
```

website_snippet_country_dropdown/views/snippets.xml
```diff
@@ -22,6 +22,7 @@
 
 <template id="country_dropdown" name="Country Dropdown">
     <t t-set="default_value" t-value="default_value or ''"/>
+    <t t-set="extra_classes" t-value="extra_classes or ''"/>
```
yajo: Oh sorry I didn't see https://github.com/OCA/e-commerce/pull/305, so I now understand why you need this.

Since there are several inputs inside this snippet, I find this variable name not explicit enough. You wouldn't know which one is getting the class. Better:

```suggestion
    <t t-set="no_country_field_extra_classes" t-value=""/>
```

I think also that the code inside `t-value` is not necessary (that's why I removed it there). Could you check that too please?

ernestotejeda:
Changes done
@Yajo I changed the name to the variable but I didn't delete the code inside t-value because I test it and it doesn't work then I left it as it was

yajo (approved):

pedrobaeza (approved):

PR #670, opened 2019-10-16
[13.0][MIG] website_legal_page

emagdalenaC2i:


emagdalenaC2i:
@katyukha @Tardo @eLBati @simahawk and @Yajo Please, could you review this PR?

simahawk:
@emagdalenaC2i seems good. Can you rephrase your commit? Should be something like `[13.0][MIG] website_legal_page` and not `[IMP] website_legal_page: black, isort`. You could split them in 2 but you then you should disable pre-commit, do the migration commit, then enable pre-commit and do the formatting. Not worth for this small change IMO :)

emagdalenaC2i:
@simahawk done

simahawk (dismissed):

yajo (changes requested):
:point_up: I'd like to add a suggestion to this module, based on some experience we've had with it over the years.

This module has these problems:

1. If you actually care about the content of your legal pages and try to modify it (or even completely replace them), it's near to impossible. The way the 3 legal text templates are designed difficult user modifications (or even block them completely, making some sections auto-filled and unmodifiable).
1. If you don't care about their content, you are most likely assuming some legal responsabilities that you don't know and can be a real legal problem in the future.
1. If you changed a part of your template (which is totally encouraged), it stops receiving updates in migrations. There was a point in history<sup>1</sup> when the `fax` field was removed from Odoo, and that produced a very uncomfortable situation where you had to choose between having a 500 error (bad) or resetting the template losing legal text customizations (worse?). We cannot know the future, but there are chances that some other field disappears/gets renamed and we land in the same problem yet again.

So, we came to this conclusion: this module **should supply no default legal texts**. Instead, it should just:

1. Supply the base templates for legal acceptance, to be used in other forms <sup>2</sup>.
1. The legal pages should either:
   - Be removed<sup>3</sup>.
   - Or be replaced by views that are _unpublished by default_ and that only include content that:
     1. Has no logic (no `t-field` and similar).
     1. Only contain instructions for the user on how to fill those templates (maybe linking to reputed websites that have good legal templates; and always encouraging to ask a lawyer).
     1. Are completely removable with 1 click using normal website editor tools.

Cons of doing this:

1. Users will be forced to write their legal texts (is that a con? :thinking:)
1. Out of the box, the module will be useless until that's done.

Pros:

1. Nobody will have "default" legal texts.
1. No more broken updates.
1. No more frustrated users _being forbidden to modify their legal conditions_. (:bangbang:)

This changes how this module worked until today, so a migration is the perfect occasion to handle this change.

Thanks!

---

<sup>1</sup> It happened in [the migration from v10 to v11](https://github.com/OCA/website/commit/4571873162cf60f8364b30ffe7b0b77edd65b393#diff-f6bdbc1147d8ca001270a8d0851ca5cdL58-L77).

<sup>2</sup> In v13 `website_form` includes all editor features that were enterprise in previous versions, so we could use this chance to add support to that module that lets users just drag & drop legal acceptance required to any form, and that could theoretically let us get rid of modules like `website_crm_privacy_policy` or `website_signup_legal_page_required`.

<sup>3</sup> When a page doesn't exist and you land in it and are a website editor user, you automatically have a button that says "Create page". IMHO that's enough.

pedrobaeza:
@simahawk `[IMP] website_legal_page: black, isort` was perfectly fine. That commit is not a migration thing, and it's improving the code through that tools. And FYI a commit should never contain the number of version. @emagdalenaC2i please revert back to the original commit message. Sorry for the disturbance.

simahawk:
@pedrobaeza you are wrong. The migration was included IN the pre-commit commit, there was only one commit. :wink: 
What I said is that it would be better to split it in 2 commits but for such a small change he could skip it.

pedrobaeza:
Reading your comment, I didn't understand that, sorry, but yes, I agree with you it must be split in 2 commits like https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-13.0 states. Any way, my comment about not including version between brackets at the beginning in the commit still applies: it should be `[MIG] website_legal_page: Migration to 13.0`.

yajo:
Commits apart, @rafaelbn just explained to me that Odoo core has, since v12, a terms & conditions acceptance feature upstream for the checkout form in e-commerce:

Captura de pantalla de 2019-10-17 09-15-36

It leads the user to `/shop/terms`, which looks like this by default:

Captura de pantalla de 2019-10-17 09-15-52

Then possibly this module should link to that page instead.

Another interesting point that I forgot to mention in the review is that this module provides a very complex template linking to different pages for terms & conditions, privacy policy and legal advice. Specifically, this template _is very hard to translate_. Actually there's no legal requirement that all of these exist in different pages, and you could just have a single big page with all your legal stuff. This would simplify the template and would let us just use `/shop/terms` like upstream. A migration script could add redirections from the old locations to the new one.

Definitely this module needs a deep change IMHO.

pedrobaeza:
AFAIK, that page only appears when having `website_sale`, so not a valid option.

simahawk:
@Yajo I almost never used this module so I didn't new about these burdens. Judging from your feedback it makes sense to - at least - remove the contents (people should take care of providing their own text indeed).
Regarding the shop route: it should be something generic, like `/terms` or something similar.
Then maybe redirect the shop ones to this one in an e-commerce module.

Regarding creating the pages: we could simply add a tour / activity / reminder w/ information on which pages would be better to create and ease this process.

yajo:
I think it's worth it at least to do a quick test: what happens if we use `/shop/terms` and `website_sale` is not installed?

My bet is that it would ask the user to create the page. If that's the case, then it would make sense to use that path, because it's the easiest way to share logic with upstream.

Then: what happens if I later install `website_sale`?

I tested it [here](https://youtu.be/9oRMqcTAw9A) and you can see that the 1st part is OK, but the 2nd (installing `website_sale` later) overrides my legal terms. However, after recording the video, I checked that if I go to the template that was created by the user and I just change its `key` to be `website_sale.terms`, it properly overrides `website_sale` default one:

Captura de pantalla de 2019-10-17 09-45-18

So it seems to be **possible** to use `/shop/terms` while not having `website_sale` installed and still make it work as expected if it is later installed.

Another option that I didn't test: is what happens if `website_legal_page` adds a 301 redirection from `/shop/terms` to i.e. `/legal` or `/terms` and I later install `website_sale`? Is the user properly redirected? If so, this might be a good solution too.

pedrobaeza:
Sales terms and conditions are NOT THE SAME as these legal terms. For me we mustn't mix both.

yajo:
I'm not saying they're the same; I'm just saying there's no reason to have separate pages for all of them :man_shrugging:. All can coexist in the same page and then this module gets easier to maintain and translate.

pedrobaeza:
I prefer to keep them separated, being separated things.

yajo:
In fact, we've had in the past customers requiring to remove some of those links, and overwriting those templates was a PITA, while for those requiring to have more than one thing, it's as simple as adding anchors `#privacy, #legal-terms, #advice` in the same page and adding a summary with links in the top of the page, letting the user jump to wherever he wants.

It's simpler to maintain, simpler to translate, simpler to integrate with upstream odoo approach, simpler to migrate, and simpler to use. If we can make it so much simpler, why complicate it? KISS. :blush:

pedrobaeza:
@zeroheure as you work a lot with e-commerce websites, can you give your opinion here?

emagdalenaC2i:
@Yajo  I agree with you that a migration is the perfect time to handle this change.

But I am not the right developer to lead this kind of migration. 
I prefer to continue with other more easy /standard migrations and thus avoid that other developers more experienced in waste their time with simple tasks.

zeroheure:
Yajo has perfectly described the problems. 
I work with small companies that don't have legal department. Most use standard Legal Terms more or less validated by an accountant. This module gives the feeling that "everything is done" : just copy-paste legal terms and that's it — while it is not that easy and one should also carefully review the 2 other pages. Putting everything on the same page will force users to pay attention to the 3 parts.
IMHO the module miss an option to enable the parts, that would force users to think about it.

(and no, I don't build a lot of ecommerce website curently. This was in the past, during 90's and around 2008-2015, not with Odoo)

pedrobaeza:
Thanks for your opinion, @zeroheure. I assumed that as you always put issues about that in Odoo ;)

What about to put everything in one page or each one as isolated?

zeroheure:
@pedrobaeza it is just that I've build nearly 400 website (yes, really) in the 90's and not so much with Odoo recently. 

Putting everything on the same page will force website owner to pay attention to the 3 parts and that is necessary. It is also a lot more simple for website readers. 
Putting on 3 pages without an option to enable each page, will let website owner install and forget.

pedrobaeza:
OK, let's go for one page! Any way, the contributor has said that he is not able to do it.

emagdalenaC2i:
@pedrobaeza you can close this PR if you want
I will take a look and review the new PR but if it is not a standard migration is better that other people more involved in OCA lead the roadmap and initial development.

pedrobaeza:
Let's keep it opened for having the conversation in mind while no alternative is presented.

rafaelbn:
@emagdalenaC2i this module for GDPR should use as contact form metadata. See new v13 feature

yajo:
That's unrelated.

The module we had for that is `website_form_metadata`, independent from this one.

website_legal_page/views/website_terms.xml
```diff
@@ -0,0 +1,248 @@
...
+                                <div class="container">
+                                    <div class="row">
+                                        <div class="col-md-12 mb16 mt16">
+                                            <h2>Terms of Use ("Terms")</h2>
+                                        </div>
+                                    </div>
+                                    <div class="row">
+                                        <div class="col-md-12 mb16 mt16">
+                                            <p>
+                                                Please read these Terms of Use ("Terms", "Terms of Use") carefully
```
tarteo, line 36: I can't remove this section, i would like to be able to remove this.

website_legal_page/views/website_privacy.xml
```diff
@@ -0,0 +1,238 @@
...
+                            </section>
+                        </div>
+                        <div class="oe_structure">
+                            <section class="s_text_block">
+                                <div class="container">
+                                    <div class="row">
+                                        <div class="col-md-12 mb16 mt16">
+                                            <p>
+                                                <span t-field="res_company.name"/>
+                                                ("us", "we", or "our") operates
```
tarteo, line 32: I can't remove this section, i would like to be able to remove this.

website_legal_page/views/website_legal.xml
```diff
@@ -0,0 +1,99 @@
...
+                <t t-call="website.layout">
+                    <div id="wrap">
+                        <div class="oe_structure">
+                            <section data-snippet-id="title">
+                                <h1 class="text-center">
+                                    Legal Advice
+                                </h1>
+                            </section>
+                        </div>
+                        <div class="container">
```
tarteo, line 23: I can't remove this section, i would like to be able to remove this.

tarteo (changes requested):
I have no knowledge about the legal aspects so I can't comment on that. Tested functionality and reviewed code.

simahawk:
@emagdalenaC2i any progress?

emagdalenaC2i:
@simahawk see https://github.com/OCA/website/pull/670#issuecomment-543154131 and https://github.com/OCA/website/pull/670#issuecomment-543303518

Tardo:
I will try to finish this. Thanks for your efforts @emagdalenaC2i

emagdalenaC2i:
I am glad that another Galician guy continues with it ;)

yajo:
Closing in favor of #699.

Issue #417, opened 2018-02-06
[RFC] Piwik is now Matomo, what about website_analytics_piwik?


astirpe:
As explained in this [Piwik is now Matomo - Announcement](https://matomo.org/blog/2018/01/piwik-is-now-matomo/), we should rename the actual module `website_analytics_piwik` to `website_analytics_matomo`.

To be strict, not only the module name but also all the fields and js code should be renamed/updated as well. 
However this approach will break the backward compatibility with the running old versions of piwik servers.

In my opinion it would be convenient to add a new module `website_analytics_matomo`, and keep the former `website_analytics_piwik` for  backward compatibility. We will let the user switch the module at the time they will upgrade their analytic server from Piwik to Matomo.

I you have different opinions or better ideas, they will be appreciated.

astirpe:
cc @hbrunn

hbrunn:
I think there's no need to change anything, as the files to be included are still called $domain/piwik.js and $domain/piwik.php. So after an update, it just keeps working - we did that already.

A PR to the readme of the latest version specifying that Matomo is also supported, and a note for the next migration to rename the module seems sufficient to me.

PR #678, opened 2020-01-08
[13.0][MIG] website_analytics_piwik

tarteo:


astirpe:
What about renaming the module to `website_analytics_matomo`?
See https://github.com/OCA/website/issues/417

tarteo:
@OCA/website-maintainers Piwik was renamed to Matomo for a couple of years now. Should I change the technical name of the module?

website_analytics_piwik/migrations/11.0.1.0.0/post-migration.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2017 Onestein (<http://www.onestein.eu>)
```
Tardo: This file can be removed

website_analytics_piwik/views/website_config_settings.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+    <record id="view_website_config_settings" model="ir.ui.view">
+        <field name="model">res.config.settings</field>
+        <field name="inherit_id" ref="website.res_config_settings_view_form" />
+        <field name="arch" type="xml">
+            <div id="google_analytics_setting" position="after">
+                <div class="col-xs-12 col-md-6 o_setting_box" id="piwik_analytics_setting">
```
Tardo:
```suggestion
                <div class="col-12 col-md-6 o_setting_box" id="piwik_analytics_setting">
```

Tardo (changes requested):

hbrunn (approved):
I agree renaming would be nice, but seeing how old this PR is I won't bother you with it. Please just put a note on the roadmap or somewhere that we should actually do it for v14 when you follow up on the other remarks

astirpe (approved):

tarteo:
Can this be merged, please.

PR #517, opened 2018-10-01
[MIG][12.0] website_legal_page

eLBati:

PR #681, opened 2020-01-15
[13.0][MIG] website_cookie_notice

bjornbillen:
depends on website_legal_page #670


simahawk:
@bjornbillen pls check the build

bjornbillen:
> @bjornbillen pls check the build

Dependancy on website_legal_page #670 throws the error. -> **please read my suggestion below**
Think I still need to run pre-commit again since the addition of prettier though.

bjornbillen:
I would like to discuss this module and its dependancy to the website_legal_page module.
There's no reference to the website_legal_page module in the whole codebase of website_cookie_notice. So pure from a technical point of view, the module won't fail to work if you don't add the dependancy.

The dependancy is only there because there's a link inside the cookie message "visit our legal page" that refers to /terms, because that page is generated by the website_legal_page module. 

The end user could perfectly adapt the text and put a different URL in it (or no URL at all), create a static page with the legal information in it, but this dependancy forces him to install a second module and reuse pages that is being created by the other module. This kind of relates to the discussion in #670 

Also, in most website implementations these days, we have to customize the cookie bar a lot, so this module doesn't seem to have much value anymore. Website visitors need to accept not just one, but multiple cookies these days. We should think about adding that kind of flexibility.

I will give it a go and send a new PR

simahawk:
@bjornbillen I agree w/ you. One way to solve this is to add a config field to the website that links that page for privacy policy. Then we don't care who sets the page and users can configure it via UI.

bjornbillen:
> @bjornbillen I agree w/ you. One way to solve this is to add a config field to the website that links that page for privacy policy. Then we don't care who sets the page and users can configure it via UI.

I've just given your suggestion a try. Let me know what you think.

website_cookie_notice/__manifest__.py
```diff
@@ -12,6 +12,6 @@
     "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/website",
     "license": "AGPL-3",
-    "depends": ["website_legal_page"],
-    "data": ["templates/website.xml"],
+    "depends": ["base", "website"],
```
simahawk: base is not needed

website_cookie_notice/templates/website.xml
```diff
@@ -0,0 +1,45 @@
...
+    </template>
+    <template id="message" name="Cookies notice">
+        <t t-set="cookie_legal_page" t-value="website.cookie_notice_legal_page_id" />
+        <div
+            t-if="request and
+        not request.session.get('accepted_cookies')"
+            id="website_cookie_notice"
+            class="container-fluid cc-cookies alert alert-warning"
+            role="alert"
+        >
```
simahawk, line 22: weird formatting here

simahawk (approved):
small things to change but looks good IMO. Code review only.

simahawk:
BTW you could squash commits after migration together as "[REF] website_cookie_notice: get rid of legal page dep" or similar

website_cookie_notice/templates/website.xml
bjornbillen: I think that's what prettier did to the xml file.. still not optimal but just following the guidelines

feketemihai (approved):

Tardo (approved):
:+1: LGTM

rafaelbn (dismissed):
Why dependency of website_legal_page is removed?

rafaelbn:
> we have to customize the cookie bar a lot, so this module doesn't seem to have much value anymore

We not, install and run. Only need the legal text from the customer to copy-paste. Same need since we created in v8. I don't see any good reason to change it. The only thing I see with this removal is more dependency to technical staff. Functional lose, technical more work.

I want to maintain it then for "simple and fast"

Tardo (changes requested):
Totally agree with @rafaelbn this module must depends on website_legal_pages.

bjornbillen:
Different opinions, I can understand what you are saying, but... Maybe you guys should have a look at the discussion on website_legal_terms and realize that it will take months *again* before both of the modules are migrated. 

Don't understand me wrong, I don't want to get rid of the dependancy because people would get access to the cookie notice module any sooner.. I just really believe that the dependancy is unnecessary in a technical way, because again, there's no reason the module won't install or work properly without it. 

I don't agree with what you guys are saying about "functional lose" and "requires too much customization" . Again I would like to refer to the discussion in #670. "Simple and fast" is not the correct solution here. The user should be triggered to think about legal pages and cookie notices on their website, not just copy-pasting a default setup.

In my opinion, the problem is not inside this module, but in website_legal_page. I have implemented these two modules plenty of times for customers. The first two things that have been done every time were : changing the content of the legal pages and changing the URLs, which was almost impossible from a customer point of view and considered "too technical".

Maybe we should consider getting rid of the website_legal_page module first. Or could you just imagine for a second that the module doesnt exist? Would the website cookie notice then totally be useless, or still handy somehow? 

The modules have been the same since v8 indeed, but they've been incorrect and incomplete ever since. In reality : not at all "ready to use".

I feel like I've spent enough time on this already. If you don't like it, no problem, I accept that. Just remove my pull request and migrate the module yourself.

Kind regards

simahawk:
> We not, install and run. Only need the legal text from the customer to copy-paste. Same need since we created in v8. I don't see any good reason to change it. The only thing I see with this removal is more dependency to technical staff. Functional lose, technical more work.
> 
> I want to maintain it then for "simple and fast"

Hola @rafaelbn :)

You should read #670 indeed and have a chat w/ your mate @Yajo maybe.

From my POV there's no need to make this module depend on that one just because that one provides you some default pages, especially if - as described in #670 - those pages will probably go away. If you want an all-in-one install, you can create a new module that glues them (w/ auto-install you won't need to care about it neither).

What is important is that when you install cookie notice it's clear and prominent that you must create a page and link it.

rafaelbn:
Indeed I inform @Yajo about that new functionality about legal terms en 2019 and I inform about that in 2018 to @eLBati  -> https://github.com/OCA/website/pull/517#pullrequestreview-160770226

If you  don't want to depend in website_legal_page, OK. Then create in this module the new page "privacy-policy".  I don't think that a user install this module, thinks its all OK and the privacy policy goes to "Aboutus" and have to fight with that. Minimal create /privacy-policy with a disclosure text. See:



After all, I have fo fight after migrations with the customers.

I don't block this then, but delegate responsibility for the migrations of our customers from previous versions in @Tardo and @Yajo :smile: :innocent:

website_cookie_notice/models/website.py
```diff
@@ -0,0 +1,11 @@
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class Website(models.Model):
+    _inherit = "website"
+
+    cookie_notice_legal_page_id = fields.Many2one(
+        comodel_name="website.page", string="Linked legal page"
```
yajo, line 10: This is wrong IMHO.

If you use a page here, there's no way of adding an `#anchor` to it. As discussed in #670, this is most likely the path to follow for the future.

Also I don't think we really need to go for res.settings to change the link. We already have the website editor which lets users do that.

So I think the best is to remove all models and just provide a basic template.

website_cookie_notice/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,17 @@
...
+``cc-cookies`` class. You can style it at will too.
+
+To change the link to the privacy-policy page:
+* Go to *Website > Configuration*.
+* Search for the section *Cookie Notice*.
+* Select the website page or create a new one. The URL field will be put as a link in the cookie message.
+
+OR
+
+* Overwrite the *<a></a>* in the cookie_notice view
```
yajo, line 17: Only 2nd option should be enabled.

website_cookie_notice/readme/ROADMAP.rst
```diff
@@ -0,0 +1,17 @@
+* Before version 8.0.2.0.0 of this module, users had the ability to configure
+  the message functionality and appearance from the main company form.
+
+  Now, the message is generated in a view. This means that after upgrading to
+  >= 8.0.2.0.0 you will lose your previous customized messages. If you want to
+  customize it, please follow steps in the configuration section.
```
yajo, line 6: Is there really no way to rescue customizations? A migration script could do that in theory...

website_cookie_notice/templates/website.xml
```diff
@@ -0,0 +1,45 @@
...
+            <script
+                type="text/javascript"
+                src="/website_cookie_notice/static/src/js/accept_cookies.js"
+            />
+        </xpath>
+    </template>
+    <template id="message" name="Cookies notice">
+        <t t-set="cookie_legal_page" t-value="website.cookie_notice_legal_page_id" />
+        <div
+            t-if="request and
```
yajo, line 17: In general, don't include logic in a template meant to be overwriten by hand.

I think it's safer to include in the main layout something like this:

```xml
<t t-if="request and not request.session.get('accepted_cookies')">
    <t t-call="website_cookie_notice.message"/>
</t>
```

And let `website_cookie_notice.message` be a simple template without any logic in it, to be overriden by end users.

website_cookie_notice/templates/website.xml
```diff
@@ -0,0 +1,45 @@
...
+            id="website_cookie_notice"
+            class="container-fluid cc-cookies alert alert-warning"
+            role="alert"
+        >
+            <div class="container">
+                <div class="row">
+                    <div
+                        class="col-sm-10"
+                    >We use cookies in this website. Read about them in our <a
+                            t-att-href="cookie_legal_page and cookie_legal_page.url or '#'"
```
yajo, line 28: As explained above, do not use logic, just a raw link to `/shop/terms#privacy` (which most likely will be the default privacy link as discussed in #670).

yajo (changes requested):
I agree on the removal of the dependency. :+1: 

However this is now a little bit overengineered. There should be just 1 way to do something, and it should be the simplest one. See comments.

Tardo:
This PR can be closed.

PR #695, opened 2020-02-11
WIP:   10.0 website page security

gfcapalbo:
Add a module that allows to define per page group security.

The  admin user and  users belonging to a special group will be presented with an extra icon in the top right bar of the website. 
This will show a popup that will allow to choose groups per-page.

No group selected means , no security (normal page visualization).

WIP, still need to make OCA-compliant , porting tested code from a client project to Community.

@eLBati  would be interesting to have you ideas on how could this work with menu_security.


simahawk:
@gfcapalbo you might want to have a look at https://github.com/OCA/website-cms/pull/41
and https://github.com/OCA/website/pull/394

Issue #677, opened 2020-01-03
[12] [website_no_crawler] error when get robots.txt


luisplanes:
When get robots.txt directly browser odoo reponse with Internal Server Error:

```
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/werkzeug/serving.py", line 270, in run_wsgi
    execute(self.server.app)
  File "/usr/lib/python3/dist-packages/werkzeug/serving.py", line 258, in execute
    application_iter = app(environ, start_response)
  File "/opt/odoo/odoo/odoo/service/server.py", line 434, in app
    return self.app(e, s)
  File "/opt/odoo/odoo/odoo/service/wsgi_server.py", line 142, in application
    return application_unproxied(environ, start_response)
  File "/opt/odoo/odoo/odoo/service/wsgi_server.py", line 117, in application_unproxied
    result = odoo.http.root(environ, start_response)
  File "/opt/odoo/odoo/odoo/http.py", line 1320, in __call__
    return self.dispatch(environ, start_response)
  File "/opt/odoo/odoo/odoo/http.py", line 1293, in __call__
    return self.app(environ, start_wrapped)
  File "/usr/lib/python3/dist-packages/werkzeug/wsgi.py", line 766, in __call__
    return self.app(environ, start_response)
  File "/opt/odoo/odoo/odoo/http.py", line 1488, in dispatch
    result = ir_http._dispatch()
  File "/opt/odoo/odoo/addons/website_sale/models/ir_http.py", line 15, in _dispatch
    return super(IrHttp, cls)._dispatch()
  File "/opt/odoo/odoo/addons/auth_signup/models/ir_http.py", line 19, in _dispatch
    return super(Http, cls)._dispatch()
  File "/opt/odoo/odoo/addons/web_editor/models/ir_http.py", line 22, in _dispatch
    return super(IrHttp, cls)._dispatch()
  File "/opt/odoo/odoo/addons/http_routing/models/ir_http.py", line 403, in _dispatch
    result = super(IrHttp, cls)._dispatch()
  File "/opt/odoo/odoo/odoo/addons/base/models/ir_http.py", line 207, in _dispatch
    return cls._handle_exception(e)
  File "/opt/odoo/odoo/addons/website/models/ir_http.py", line 195, in _handle_exception
    return super(Http, cls)._handle_exception(exception)
  File "/opt/odoo/odoo/odoo/addons/base/models/ir_http.py", line 177, in _handle_exception
    return request._handle_exception(exception)
[...]

PR #698, opened 2020-02-19
[12.0] website_no_crawler fix inheritance

colmeirin:
[12.0] website_no_crawler fix inheritance in view to properly generate robots.txt file as discused at issue #677


yajo (approved):
Tested on runbot; it works.

PR #699, opened 2020-02-24
[13.0][MIG] website_legal_page: Migration to 13.0

Tardo:
**Superseding:** #670 

- Now uses a single page
- Fully compatible with the website builder
- Redirects '/shop/terms' to '/legal' (I don't know why really works, but works... maybe the dependencies?)
- Migration script to merge old pages: website_legal_page.advice, website_legal_page.privacy-policy, website_legal_page.terms-of-use and website_sale.terms into a new single one.

cc @tecnativa TT19849


chienandalu (commented):
Should there be a migration script to redirect old separate urls?

Tardo:
migration script done

chienandalu (approved):
Thanks :)

Tardo:
Ok now redirects '/shop/terms' to '/legal' (I don't know why really works, but works... maybe the dependencies?)

Ready to finish review @Yajo

website_legal_page/migrations/13.0.1.0.0/post-migrate.py
```diff
@@ -0,0 +1,112 @@
...
+    "website_sale.terms",
+]
+
+
+def _merge_old_views(env):
+    # Gets the old views that was changed by user (has a website_id)
+    old_view_ids = env["ir.ui.view"].search(
+        [
+            ("key", "in", OLD_VIEW_XML_IDS),
+            ("website_id", "!=", False),
```
yajo: I'm not sure this filter makes sense. If you didn't edit the templates, still the default templates should be migrated, isn't it? :thinking: 

```suggestion
```

website_legal_page/tests/test_controller.py
```diff
@@ -2,27 +2,10 @@
...
-        """ It should return a 200 for advice page. """
-        self._test_page("/legal/advice")
-
-    def test_tos(self):
-        """ It should return a 200 for ToS page. """
-        self._test_page("/legal/terms-of-use")
+    def test_page(self):
+        """ It should return a 200 for legal page. """
+        response = self.url_open("/legal", timeout=20)
+        self.assertEqual(response.status_code, 200)
```
yajo, line 12: Test the controller too please.

website_legal_page/views/reusable_templates.xml
```diff
@@ -1,14 +1,11 @@
...
     <!-- Base templates for submodules -->
     <template id="acceptance_full">
         <span class="acceptance_full">
-            I accept the <a href="/legal/advice">legal advice</a>, the
-            <a href="/legal/privacy-policy">privacy policy</a>, and the
-            <a href="/legal/terms-of-use">terms of use</a>
-            of this website.
+            <a
+                href="/legal"
+            >I accept the legal terms &amp; conditions of this website.</a>
```
yajo:
```suggestion
            >I accept the legal terms, the privacy policy &amp; conditions of this website.</a>
```

website_legal_page/views/website_legal_main_page.xml
```diff
@@ -0,0 +1,410 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <!-- Add legal terms and conditions page link after copyright and company name -->
+    <template
+        id="legal_page_link"
+        inherit_id="website.footer_custom"
+        name="Legal page Link"
```
yajo, line 7: Users should be able to toggle this per website.

```suggestion
        name="Legal page Link"
        customize_show="True"
```

Note: I'm not sure it's done like this in v12+, please double-check it.

yajo (changes requested):
Code seems good, I didn't test functionally.

There's something missing though:

1. When installing the module, if `/shop/terms` exists (a.k.a `website_sale` is installed), its contents should be added by default to `/legal`.
1. I'd appreciate a way to test the controller when `website_sale` is instaled along with this module. This is hard to achieve because there's no direct dependency between them, a suggestion:
   - Add a demo data file with:

     ```xml
        <data noupdate="1">
            <function model="ir.module.module" name="button_immediate_install">
                <value eval="[ref('base.module_website_sale')]"/>
            </function>
        </data>
     ```

    - Add a test that asserts the controller redirection. Not sure, but possibly this test should be in post-install mode.

    - If possible, test also that the process of importing contents from preexisting `/shop/terms` works fine.

Awesome work, BTW. :sunglasses:

website_legal_page/migrations/13.0.1.0.0/post-migrate.py
Tardo: Good question... but ok, if you think that old samples need be merged, i do it. No problem.
In case of merge old samples (no modified by user) i will preserve the alert message "THIS IS A SAMPLE".

Tardo:
Changes done! @Yajo I can't do the test for website_sale... but you can test it maually in a very easy way: Install website_sale then install this module.

yajo (approved):
It seems quite good. Thanks!

cvinh:
Hello
Thanks for this work that is quite mandatory on every website of the world.
By the way, it looks like the 'Privacy Policy' template is not in v13 anymore.
It's also quite mandatory with new regulations about Privacy in Europe and other countries.
For example, if you want to use facebook oauth, they ask for the privacy policy url. If not provided, the apps is deleted.
Do you intend to get the private policy template back in this module ?

Tardo:
@cvinh read this thread please: https://github.com/OCA/website/pull/670#issuecomment-543116294

If you come from 12.0 you should keep your previous "privacy policy" on the new '/legal' page. 
If you use a new installation, you must edit the current legal page with your use case and use 'anchors'.
If need a new page, you must create it.

cvinh:
Thank you for reply but my suggestion was only to add #privacy anchor and sample text about privacy... just to have something talking about privacy from scratch...
I can do a PR

cvinh:
Hello
I did a PR in order to add the Privacy Policy text in the template : https://github.com/OCA/website/pull/739

cvinh:
And I did a PR to replace the ** FIELDS by the real fields of res_company here : https://github.com/OCA/website/pull/740

PR #703, opened 2020-03-12
[FIX] website_form_builder: restore functionality of required & hidden fields

yajo:
This functionality was lost in v11 migration. Now it is recovered and tested.

What has been recovered:

- Required fields now actually are required.
- When hiding a required field, you are asked for a default value.

A migration script is added, which removes the `o_required` style for forms that were actually not required. This seems to be a safer approach than leaving them looking as required but not being that, or marking them as required and possibly breaking some workflows in forms built with previous bug.

# NOTE: Must be merged with ocabot merge minor

@Tecnativa TT22901

### DIFF
--- a/website_form_builder/migrations/11.0.1.2.0/post-migrate.py
+++ b/website_form_builder/migrations/11.0.1.2.0/post-migrate.py
@@ -1,0 +1,40 @@
+# Copyright 2020 Tecnativa - Jairo Llopis
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl).
+
+from openupgradelib import openupgrade as ou
+from openupgradelib.openupgrade_tools import (
+    convert_html_fragment,
+    convert_html_replacement_class_shortcut as _r,
+)
+
+REPLACEMENTS = (
+    _r(
+        # Use XPath mode because LXML's CSS mode doesn't support `:has()`
+        selector="""
+        //*
+        [hasclass('form-field', 'o_required')]
+        [//input[not(@required)]]
+        """,
+        selector_mode="xpath",
+        class_rm="o_required",
+    ),
+)
+
+
+@ou.migrate()
+def migrate(env, version):
+    """Fix existing form that lost their required behavior.
+
+    Setting fields to required might break existing functionality that passed
+    unnoticed thanks to the bug fixed in this version, so the fix is not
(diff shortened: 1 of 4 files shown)


pedrobaeza (approved):

rafaelbn (approved):
Thanks! :+1:

website_form_builder/migrations/1.2.0/post-migrate.py
```diff
@@ -0,0 +1,40 @@
...
+    convert_html_replacement_class_shortcut as _r,
+)
+
+REPLACEMENTS = (
+    _r(
+        # Use XPath mode because LXML's CSS mode doesn't support `:has()`
+        selector="""
+        //*
+        [hasclass('form-field', 'o_required')]
+        [//input[not(@required)]]
```
Tardo: What this xpath does? Im lost with the `[//input[not(@required)]]` part...

All elements with 'form-field' & 'o_required' that have an input without required?

yajo: > All elements with 'form-field' & 'o_required' that have an input without required?

exactly.

Tardo (approved):

website_form_builder/migrations/1.2.0/post-migrate.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2020 Tecnativa - Jairo Llopis
```
pedrobaeza: Incorrect migration folder!!! It should be fully qualified, or if not, we can have unpredictable migration

pedrobaeza (changes requested):

website_form_builder/migrations/1.2.0/post-migrate.py
yajo: I did this on purpose, to be able to cherry-pick this to v12, which is also affected.

pedrobaeza: Please do it the other way. Always fully qualified. When cherry-picking, you can do `git mv`. You don't even know if which version v12 is (the numbering is independent).

yajo: Done.

pedrobaeza (approved):

PR #704, opened 2020-03-19
[FIX] website_form_builder: restore functionality of required & hidden fields

yajo:
Forward port of #703.

This functionality was lost in v11 migration. Now it is recovered and tested.

What has been recovered:

- Required fields now actually are required.
- When hiding a required field, you are asked for a default value.

A migration script is added, which removes the `o_required` style for forms that were actually not required. This seems to be a safer approach than leaving them looking as required but not being that, or marking them as required and possibly breaking some workflows in forms built with previous bug.

@Tecnativa TT22901

### DIFF
--- a/website_form_builder/__manifest__.py
+++ b/website_form_builder/__manifest__.py
@@ -3,7 +3,7 @@
 {
     "name": "Website Form Builder",
     "summary": "Build customized forms in your website",
-    "version": "12.0.1.0.0",
+    "version": "12.0.1.1.0",
     "category": "Website",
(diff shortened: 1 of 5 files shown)


chienandalu (commented):
Looks like tests are failing

yajo:
Oh of course, the tour must have changed...

yajo:
Tour updated, it should go :heavy_check_mark: now.

chienandalu (approved):
Thanks! :+1:

PR #714, opened 2020-04-15
[FIX] website_adv_image_optimization: Changes the image selection after changes

Tardo:
Before this commit, when go back to the media dialog after use optimization dialog,  the image loses
the selection focus. With this commit the image doesn't loses the focus.

cc @tecnativa


pedrobaeza (approved):

pedrobaeza:
Fast-tracking for fixing the problem

/ocabot merge patch

PR #719, opened 2020-05-12
website_lazy_load_image: fix doctype missing (closes #718)

### LINKED ISSUE #718: v12 - Doctype tml is missing
Hi everyone,

I'm facing the same issue as https://github.com/odoo/odoo/issues/44612

I confirm that it's running properly on runbot Odoo and runbot OCA/OCB.

Yet OCA/website is obviously causing troubles , checked on this db http://3421796-12-0-317600.runbot2-2.odoo-community.org
image
no doctype in the page here.

flotho:
Fix #718 and probably https://github.com/odoo/odoo/issues/44612

### DIFF
--- a/website_lazy_load_image/__manifest__.py
+++ b/website_lazy_load_image/__manifest__.py
@@ -8,7 +8,7 @@
     'license': 'AGPL-3',
     'website': 'https://github.com/oca/website',
     'category': 'Website',
-    'version': '12.0.1.0.1',
+    'version': '12.0.1.0.0',
     'depends': [
         'website'
     ],
--- a/website_lazy_load_image/__manifest__.py
+++ b/website_lazy_load_image/__manifest__.py
@@ -8,7 +8,7 @@
     'license': 'AGPL-3',
     'website': 'https://github.com/oca/website',
     'category': 'Website',
-    'version': '12.0.1.0.0',
+    'version': '12.0.1.0.1',
     'depends': [
         'website'
     ],
--- a/website_lazy_load_image/models/ir_ui_view.py
+++ b/website_lazy_load_image/models/ir_ui_view.py
@@ -32,5 +32,6 @@ class IrUiView(models.Model):
                 src = img.attrib['src']
                 img.attrib['src'] = self.LAZYLOAD_DEFAULT_SRC
                 img.attrib['data-src'] = src
-            res = lxml.etree.tostring(html, method='html', encoding='UTF-8')
+            res = lxml.etree.tostring(html, method='html', encoding='UTF-8',
+                                      doctype=html.getroottree().docinfo.doctype)
         return res
--- a/website_lazy_load_image/tests/test_lazy_load_image.py
+++ b/website_lazy_load_image/tests/test_lazy_load_image.py
@@ -3,8 +3,10 @@
 
 from odoo.tests.common import SavepointCase
 from lxml import etree
+from odoo.tests import tagged
 
 
+@tagged('-at_install', 'post_install')
 class LazyLoadTest(SavepointCase):
 
     @classmethod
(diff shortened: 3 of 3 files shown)


website_lazy_load_image/models/ir_ui_view.py
```diff
@@ -32,5 +32,6 @@ def render_template(self, template, values=None, engine='ir.qweb'):
                 src = img.attrib['src']
                 img.attrib['src'] = self.LAZYLOAD_DEFAULT_SRC
                 img.attrib['data-src'] = src
-            res = lxml.etree.tostring(html, method='html', encoding='UTF-8')
+            res = lxml.etree.tostring(html, method='html', encoding='UTF-8',
```
simahawk: seems to make sense but according to lxml docs / issues this should work OOTB since some versions.
Which lxml version are you using?
@tarteo do you experience the same issue?

tarteo: Yes this seems to fix it.

tarteo (approved):
The doctype is persisting with this fix.

Before:
image

After:
image

simahawk:
@flotho can you check the build?

HaraldPanten (approved):
Small fix, seems to be OK

flotho:
Hi @simahawk , @HaraldPanten ,

More tests have been added and travis is green, the codecov is not concerning my PR.
Is a merge possible ?

(cc @pedrobaeza , I finally succeed in squashing and respecting pylint !!! so happy )

website_lazy_load_image/tests/test_lazy_load_image.py
```diff
@@ -93,8 +97,20 @@ def test_encoding_render(self):
...
         robots = self.env.ref('website.robots').render()
-        self.assertNotIn('<html>', robots.decode(
-            'UTF-8'), "Robots must not be wrapped into html DOM")
+        self.assertNotIn('<html>', robots.decode('UTF-8'),
+                         "Robots must not be wrapped into html DOM")
+
+    def test_doctype_in_main(self):
+        """ Check that at least doctype is preserved on website """
+        public_user_id = self.ref('base.public_user')
+#       TODO: Improve this part and find a way to render website template
```
simahawk: what's the problem? If it's the request not bound, you can use https://github.com/odoo/odoo/blob/master/addons/website/tools.py#L76

website_lazy_load_image/__manifest__.py
```diff
@@ -8,7 +8,7 @@
     'license': 'AGPL-3',
     'website': 'https://github.com/oca/website',
     'category': 'Website',
-    'version': '12.0.1.0.0',
+    'version': '12.0.1.0.1',
```
simahawk: this is not required as it will be done via bot. Also, ATM there's release on pypi w/ `nobump` option.

simahawk (changes requested):
thnx for adding tests!

website_lazy_load_image/tests/test_lazy_load_image.py
flotho: @simahawk , as it's my first test ever on Odoo I was not aware of this mockup class. I'll use it later on.

flotho:
Hi @simahawk ,

I need a little help on this one. I'm struggling on the test. It works locally with a non demo database and it failed on runbot but I'm not able to understand why and how to handle this case.

Regards

ypapouin:
@simahawk, thanks for the routing to the right issue/fix.
Ok for the fix but I'm not sure that putting test on this module is relevant as any module hijacking the final rendering can break everything. Don't you think that it should be moved as a base/web test in OCB instead ?

simahawk:
@ypapouin we must make sure THIS module which hacks the DOM works properly. 
If any other module is messing up w/ the DOM it should do the same.
What we could do is: have a base module like `qweb_render_hook` or `base_render_hook` or alike which provides the base features and tests utils to handle such case. 
But for now I'd fix the tests here and add maybe a TODO in the roadmap to split it later.

ypapouin:
Ok some tests could be needed but fixing this module should be a priority. Since this PR exists for more than 6 months it would be fair to split this PR (KISS mode) : one for the fix (that could be approved and merged quickly) and the other one for the test that way the PR for the test could stay here until it is fixed.

simahawk:
let's see if tests pass here https://github.com/OCA/website/pull/810

simahawk:
replaced by #810

flotho:
@simahawk , thanks for helping, your PR helps me to understand mochrequest.

Regards

PR #720, opened 2020-05-13
[IMP] website_snippet_country_dropdown: readonly mode

yajo:
- Remove some garbage code.
- Add readonly mode.
- Add demo data.
- Improve tour.

@Tecnativa TT17694

### DIFF
--- a/website_snippet_country_dropdown/demo/pages.xml
+++ b/website_snippet_country_dropdown/demo/pages.xml
@@ -16,15 +16,30 @@
                                 Test here the country dropdown snippet behavior
                             </h3>
                             <form action="" class="form-horizontal">
-                                <t t-call="website_snippet_country_dropdown.dropdown_element_form">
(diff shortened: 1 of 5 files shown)


pedrobaeza (approved):

Tardo (changes requested):
Looks Good! I think that you can remove all related with "this.$country_code"

yajo:
> I think that you can remove all related with "this.$country_code"

I think it's still useful. As it is right now, 3 fields are submitted in the form: full text, country code alone, vat alone. It seems the easiest way to support any use case downstream. Besides, that could potentially break backwards compatibility quite badly. Maybe we could consider that when migrating, but I don't really see a big benefit TBH. :thinking:

Tardo:
@Yajo I can't found where 'js_country_code' is defined...

yajo:
Ooh yes, understood. Removed now. Thanks!

Tardo (approved):
:+1:

PR #725, opened 2020-05-22
[FIX] website_canonical_url: use website domain

yajo:
It didn't make sense that a website with a specified domain didn't use it as the canonical one by default.

@Tecnativa TT23961

### DIFF
--- a/website_canonical_url/models/website.py
+++ b/website_canonical_url/models/website.py
@@ -35,6 +35,8 @@ class Website(models.Model):
         self.ensure_one()
         if self.canonical_domain:
             return self.canonical_domain
+        if self.domain:
+            return self.domain
         params = self.env['ir.config_parameter'].sudo()
         return params.get_param('web.base.url')
 


pedrobaeza (changes requested):
Website domain must have priority over the other IMO. I would even remove that field.

simahawk:
> Website domain must have priority over the other IMO. I would even remove that field.

You can have domain that is a subdomain and in case you have multi-website you might want one unique canonical for all of them.

It has been done for this reason: https://github.com/OCA/website/pull/449#discussion_r183980128

simahawk (approved):
@Yajo LG. If you have time, pls update tests :pray:  (pre-approving to not lock you)

yajo:
> I would even remove that field.

canonical tag is upstream in v13: https://github.com/odoo/odoo/pull/35852

So this module will disappear for that version and yes, that field will be removed. But it could be better done in the OU migration script, which we'll have to do anyways, don't you think?



> pls update tests

Tests seem to pass :heavy_check_mark: :roll_eyes:

yajo:
> You can have domain that is a subdomain and in case you have multi-website you might want one unique canonical for all of them.
> 
> It has been done for this reason: [#449 (comment)](https://github.com/OCA/website/pull/449#discussion_r183980128)

BTW I don't think that's applicable anymore. That was v10, where there was just 1 website really. In v12 we have multiwebsite built-in, and thus there's no reason why you would have canonical URLs shared among them because they're completely different. Any other kind of canonicalization should be done at the proxy level, applying 301 redirections when user/crawler lands in the non-canonical domain.

The main use of canonical URLS is well explained in https://github.com/odoo/odoo/issues/12532 and doesn't have anything to do with multiple domains IIRC.

IMHO we can just drop this module for v13.

simahawk:
> Tests seem to pass heavy_check_mark roll_eyes

Yeah, I meant: make sure `domain` usage is tested :wink: 

> BTW I don't think that's applicable anymore. That was v10, where there was just 1 website really. In v12 we have multiwebsite built-in, and thus there's no reason why you would have canonical URLs shared among them because they're completely different. 

That comment was for v11 whereas we had already OCA's multi website support.

> Any other kind of canonicalization should be done at the proxy level, applying 301 redirections when user/crawler lands in the non-canonical domain.

Of course, it was more a way of handling this via quick app conf instead of sys admin conf.

> The main use of canonical URLS is well explained in odoo/odoo#12532 and doesn't have anything to do with multiple domains IIRC.

Yes but AFAIR the field was helping to solve some special cases for multi website too. BTW, you asked for it, I never used them till that point! :stuck_out_tongue_winking_eye: :rofl: 

> IMHO we can just drop this module for v13.

I went trough the code of the PR. No test, but seems like it's done properly now, including skipping langs that are not available. 

I agree we can drop it for v13.

pedrobaeza (approved):

PR #726, opened 2020-05-25
[12.0][ADD] website_megamenu

Tardo:
Backport of 'mega menu' feature from 13.0 CE

megamenu

cc @tecnativa TT20965


website_megamenu/models/website_menu.py
```diff
@@ -0,0 +1,43 @@
...
+# Copyright 2020 Tecnativa - Alexandre Díaz
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+from odoo import api, fields, models
+from odoo.tools.translate import html_translate
+
+
+class Menu(models.Model):
+    _inherit = "website.menu"
+
+    def _compute_field_is_mega_menu(self):
```
sergio-teruel, line 11: rename to _compute_is_mega_menu

website_megamenu/models/website_menu.py
```diff
@@ -0,0 +1,43 @@
...
+        for menu in self:
+            menu.is_mega_menu = bool(menu.mega_menu_content)
+
+    def _set_field_is_mega_menu(self):
```
sergio-teruel, line 15: Rename to _inverse_is_mega_menu

sergio-teruel (approved):
Only little comment

rafaelbn (approved):
:+1: Thanks

@Tardo you have some minor comments from @sergio-teruel

website_megamenu/models/website_menu.py
yajo: Although I'd generally agree, I think all he's done is copy upstream code from v13: https://github.com/odoo/odoo/blob/13.0/addons/website/models/website_menu.py#L20

So I think we better leave it as this, so there's less diff with upstream when this backport module is dropped in v13.

website_megamenu/models/website_menu.py
yajo: Same as above, better as it is.

website_megamenu/readme/CREDITS.rst
```diff
@@ -0,0 +1,3 @@
+This module is a backport from Odoo SA and as such, it is not included in the
```
yajo, line 1: Please add a `ROADMAP.rst` indicating that this module should disappear for v13 and be merged into `website`.

website_megamenu/static/src/js/content.js
```diff
@@ -0,0 +1,103 @@
+// Copyright 2020 Odoo
+// Copyright 2020 Tecnativa - Alexandre Díaz
+// License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
```
yajo, line 3: This is LGPL. The same in all files.

Also, please, in each file add a little comment in the top pointing to the upstream v13 file where code was imported from. This way, we know better how to compare it with upstream.

Thanks!

yajo (changes requested):

website_megamenu/models/website_menu.py
sergio-teruel: Ok

Tardo:
Changes done, renamed "views" folder to "templates".

yajo (approved):

PR #716, opened 2020-04-30
[ADD] website_sale_odoo_debranding: module to remove powered by in portal

cubells:
If user has installed module `website_sale`, then a `Powered by odoo` link is shown in portal views.
Before:
before

With this module:
after

I've created another module to avoid install website_sale if it is no needed.

PR #729, opened 2020-05-28
[13.0][IMP] website_odoo_debranding: remove powered by in portal

cubells:
Related: https://github.com/OCA/website/pull/716


website_odoo_debranding/templates/disable_odoo.xml
```diff
@@ -12,4 +12,16 @@
...
     </template>
+    <!-- Remove "Powered by Odoo" -->
+    <template
+        id="layout_portal_powered"
+        inherit_id="portal.portal_record_sidebar"
+        customize_show="True"
+        name="Remove Odoo Powered by Link"
+        priority="20"
+    >
+        <xpath expr="//div[hasclass('d-lg-block')]" position="replace">
```
pedrobaeza: Please use a stronger `xpath` as this may match other parts.

And hide this part instead of removing it, or you may find errors by other modules taking this same xpath as reference

pedrobaeza (changes requested):

website_odoo_debranding/templates/disable_odoo.xml
cubells: > And hide this part instead of removing it, or you may find errors by other modules taking this same xpath as reference

ok , but the code is the same that already exists in module. I don't understand why are you so strict with me.

pedrobaeza: Sorry if you think so. I'm just reviewing current PR without being aware of previous code, and pointing potential problems, but if you think it's too much, I refrain from reviewing and let others to review.

cubells: No, if you are the best reviewer! 

But this surprise me because previous code should also be changed, logically.

cubells: > Please use a stronger xpath as this may match other parts.

class `d-lg-block` **only** contains _Powered by odoo_ and only exists one time. 

I think is enough stronger, but how can I do it more?

pedrobaeza: First of all, let's put the reference of the original template for all to have it in mind:

https://github.com/odoo/odoo/blob/aca17a48280e1c1641fde08076197210c926e5f1/addons/portal/views/portal_templates.xml#L155

And my guess in the other PR gets confirmed: you don't need website for showing that template. You can have an Odoo without website, but people accessing your portal for seeing invoices.

Thus, a module `portal_debranding` should be created depending only on `portal` for covering such case instead of adding this here.

About the xpath, you can concatenate several selectors for being sure of what is what you are selecting, and avoiding mismatches. For this one, I suggest:
```xml
<xpath expr="//div[hasclass('d-lg-block')]/a[@title='odoo']/.." position="attributes">
    <attribute name="class" add="hidden" separator=" "/>
</xpath>
```
as you can see, it's easy to have another `<div>` element with the class `d-lg-block`, but very difficult to have another `<div>` with that class that contains an `<a>` with that title.

cubells: @pedrobaeza 

> as you can see, it's easy to have another <div> element with the class d-lg-block, but very difficult to have another <div> with that class that contains an <a> with that title.

First at all, thanks.
But you know that I can't use attribute 'title' as a selector. I can't use neither title or alt or src.

And by adding class `hidden` the item does not disappear. I can't use neither `hidden` or `hidden-xs` or `oe_hidden`. Item must have a class that makes it have no effect.

cubells: I can't use either `<attribute name="invisible">True</attribute>`. It does not work.

rubenseijas:
Hi,

Since v.12 in templates you need use "*d-none*" attribute to get CSS
like "*display:
none;*"

You can try something like this:

*<xpath expr="//div[hasclass('d-lg-block')]" position="attributes">*
*        <attribute name="class" add="d-none" separator=" "/>*
* </xpath>*



Regards!

*Rubén Seijas*
Software Technician
*Tfno: *982 81 48 05 *Ext*: 104
ruben@comunitea.com

website_odoo_debranding/templates/disable_odoo.xml
pedrobaeza: Sorry, I didn't remember that title is translatable, then the xpath can be (and the class to put as pointed by another contributor):
```xml
<xpath expr="//div[hasclass('d-lg-block')]/a[@contains(@href,'www.odoo.com')]/.." position="attributes">
    <attribute name="class" add="d-none" separator=" "/>
</xpath>
```

pedrobaeza: Remember also about this belonging to `portal`, not to `website`.

cubells:
@rubenseijas Thanks, it works.

Problem:

if I use this selector:  `//div[hasclass('d-lg-block')]/a/img/..` 
then link is hidden as we expect:
Captura de 2020-05-29 08-55-51

If i use this in order to hide all div: `//div[hasclass('d-lg-block')]/a/img/../..`
then it doesn't work as we expect:
Captura de 2020-05-29 08-56-45

As I said above, div has some class that avoid to be able to hide item.

Maybe original developer and **Tardo**, member of **Tecnativa**, use position `replace` by this reason in previous code.

website_odoo_debranding/templates/disable_odoo.xml
cubells: @pedrobaeza 

As I said above, selector works, but by adding class `d-none`, by adding `hidden`, `oe_hidden`, `hidden-xs` item is shown as you can see:
captura

This `div` must be a class that avoid to div to apply hidden visibility.

cubells: The class that avoid to apply visibility to div is **d-lg-block**

So I think I should use replace in position. What do you think?

pedrobaeza: Yeah, indeed there's no option for hiding it. The explanation is:

Selección_002

adding other class like `d-lg-none` doesn't serve.

So let's replace it, but take this into account:

- The view needs to have a very high priority number (`9999` for example), for being evaluated in last place for not breaking other possible views that matches that element
- Please put a comment about this in the XML for avoiding all the investigation in the future.
- Remember that this should go to a new module `portal_debranding`, not here.

website_odoo_debranding/templates/disable_odoo.xml
```diff
@@ -12,4 +12,16 @@
             <div class="o_brand_promotion" />
         </xpath>
     </template>
+    <!-- Remove "Powered by Odoo" -->
+    <template
+        id="layout_portal_powered"
```
simahawk:
```suggestion
        id="portal_record_sidebar"
```

website_odoo_debranding/templates/disable_odoo.xml
simahawk: actually there's a better option to hide something: use `position="attributes"` and apply `t-if="False"`.
This way the element is still in the view markup but it won't be rendered at all :wink: 

Regarding the selector I'd complete skip references to classes and use:
```xml
<xpath expr="//a[contains(@href,'www.odoo.com')]/.." ...
```
because in `portal_record_sidebar` you have only that link.

simahawk (changes requested):

pedrobaeza:
And I still remark that this shouldn't go in website, but it's for portal without website.

pedrobaeza:
It can be superseded by https://github.com/OCA/server-brand/pull/32 forward-porting it to 13

cubells:
@pedrobaeza @simahawk sorry i'm too busy last months.

I'm trying to come back, :) 

can you recheck?

pedrobaeza:
@cubells the problem is still the module name and the repo. Please check the other PR and it's better to forward-port it.

cubells:
@pedrobaeza the module name is changed as you can see and the code is the same, but ok.

pedrobaeza:
But the repo is not good, as this has nothing to do with website.

cubells:
@pedrobaeza yes, you are right

PR #732, opened 2020-06-10
[12.0][ADD] website_cache_controller

PCatinean:
First off I am not sure if there is a similar solution or a better approach to this problem. I couldn't find one so I thought it was easier to propose the module and explain the issue alongside it.

The problem:

Odoo website snippets and frontend content makes extensive use of the /web/image controller to display image resources, almost all snippets of manually added content (not products or product categories) use this.

At the same time the backend uses this controller as well for images that are private (customer images, mrp / workorder images etc etc). Using a cache on this path (/web/image) would be really unsafe as they would be captured by the relevant proxy / cdn from an authenticated user and served to anyone requesting that link. This does not happen if the request goes through odoo as access rules are being checked with each request.

Proposed solution:

The only way I could think of safely caching such images and distinguishing them from the rest of the private ones was to use a new path specifically for cacheable content and serving data only when it is public to avoid the capture of any potential proxys and/or cdns. Thus taking inspiration from the website_lazy_load module a replace is done for non-internal users (including crawlers) from /web/image/ to /website/cache/image.

A second layer of safety was added in the /website/cache/image controller where another check is done and a redirect is made in case the user (for some reason) would manually change the url. This can also happen with website/image controller as well btw.

Hope this makes some sense and if not I am curious to see what the proper approach would be.

### DIFF
--- a/website_cache_controller/__init__.py
+++ b/website_cache_controller/__init__.py
@@ -1,0 +1,2 @@
+from . import models
+from . import controllers
--- a/website_cache_controller/__manifest__.py
+++ b/website_cache_controller/__manifest__.py
@@ -1,0 +1,11 @@
+{
+    "name": "Website Cache Controller",
+    "summary": "Website Cache Controller",
+    "author": "Odoo Community Association (OCA), Pledra",
+    "website": "http://www.pledra.com",
+    "category": "Website",
+    "license": "AGPL-3",
+    "version": "12.0.1.0.0",
+    "depends": ["website"],
+    "data": [],
+}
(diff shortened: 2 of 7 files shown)


flotho:
What could also be awesome is to be able to rewrite image url so that they could be cached by a proxy relatively to a Location or Directory directive

PCatinean:
I am taking another look at this and I think the approach is wrong. This change should be done at the qweb parsing level when the attributes are processed. The same way Odoo is doing where it is using `odoocdn.com` -> https://github.com/odoo/odoo/blob/b249c28d3cae7cfa96c153b413dba4c16d5c04d5/addons/website/models/website.py

There you can also include more complex logic as you mentioned @flotho and is much more efficient and less risky (encoding/decoding the final parsed result)

PR #738, opened 2020-07-15
[11.0][IMP] website_cookie_notice: remove not needed query

kaerdsar:
The clear_caches method of models is a class method, then I would say is not related to specific records. The controller is generating a query which is not actually needed, at least in Odoo version 11.0 code.

### DIFF
--- a/website_cookie_notice/controllers/main.py
+++ b/website_cookie_notice/controllers/main.py
@@ -12,7 +12,5 @@ class CookieNotice(http.Controller):
     def accept_cookies(self):
         """Stop spamming with cookie banner."""
         http.request.session["accepted_cookies"] = True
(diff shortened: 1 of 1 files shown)


kaerdsar:
@eLBati could you please take a look at this change?

eLBati (approved):
LGTM, thanks.

Do you plan to cherry-pick this also for 12.0 and 13.0?

kaerdsar:
I have not checked those versions but yes, I  can do it.  What should be the workflow? Cherry-pick the commit to the other branch and create PR for 12 and 13 or is there another way?

pedrobaeza (approved):

eLBati:
>  What should be the workflow? Cherry-pick the commit to the other branch and create PR for 12 and 13

:+1:

PR #739, opened 2020-07-21
[IMP] Add Privacy template text in Legal Page

cvinh:


yajo (changes requested):
Thanks for your intention, but the legal text was deliberately removed from the module. See https://github.com/OCA/website/pull/670#pullrequestreview-303063137 to understand why.

So, :-1: to this PR. Everybody must consult their respective lawyers to redact this kind of stuff. If somebody decides to copy&paste some template from the Internet, it must be done deliberately by hand.

Thanks for your intention, though!

PR #740, opened 2020-07-21
13.0 imp website legal page

cvinh:
I changed the ** FIELDS in the template into the real fields in the database


yajo (changes requested):
Thanks for your intention, but the legal text was deliberately removed from the module. See #670 (review) to understand why.

So, :-1:  to this PR. Everybody must consult their respective lawyers to redact this kind of stuff. If somebody decides to copy&paste some template from the Internet, it must be done deliberately by hand.

Thanks for your intention, though!

PR #743, opened 2020-07-24
[FIX] website_snippet_anchor: use correct assets

yajo:
`web_editor.assets_editor` is used in the backend too, where this module has no use and logs a dependency error on `website.utils` (which is only available in the website).

Switching to `website.assets_editor` will reduce the payload on the backend and remove that warning.

@Tecnativa

### DIFF
--- a/website_snippet_anchor/templates/assets.xml
+++ b/website_snippet_anchor/templates/assets.xml
@@ -4,7 +4,7 @@
 
 <odoo>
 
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):

chienandalu (approved):

HaraldPanten (approved):
Small fix, seems to OK

PR #745, opened 2020-08-10
[ADD] website_menu_icon

Rad0van:
This module is a result of my need to render website menus with icons/images. The backend is pretty straightforward but I wanted to extend also the WYSIWYG editor as a learning experience for QWeb and JS so expect some not pretty code. 

It works fine so I offer this for inclusion into OCA. With my still limited knowledge of Odoo/Python/Qweb/JS/JQuery/... couldn't think of any proper test for this.


Rad0van:
@Yajo , @pedrobaeza and @ivantodorovich would you be so kind and look at this lame attempt? My first attempt at brand new module. I couldn't solve few things without using dirty hacks. That may be because I am newbie in JQuery and JS... Thanx a lot.

website_menu_icon/models/website_menu.py
```diff
@@ -0,0 +1,60 @@
...
+from odoo import api, fields, models
+
+
+class Menu(models.Model):
+
+    _inherit = "website.menu"
+
+    image = fields.Char("Icon", required=False, translate=True)
+    image_type = fields.Selection(
+        [("none", "None"), ("image", "Image"), ("icon", "Icon")],
```
yajo: Since the field is not required, `False` is a valid value. Then what is `none` for? Better just remove it:

```suggestion
        [("image", "Image"), ("icon", "Icon")],
```

website_menu_icon/models/website_menu.py
```diff
@@ -0,0 +1,60 @@
...
+        string="Type",
+        required=False,
+        translate=False,
+        default="none",
+    )
+
+    @api.model
+    def save(self, website_id, data):
+        super(Menu, self).save(website_id, data)
```
yajo: This code does nothing, right?

```suggestion
```

website_menu_icon/models/website_menu.py
```diff
@@ -0,0 +1,60 @@
...
+
+    _inherit = "website.menu"
+
+    image = fields.Char("Icon", required=False, translate=True)
+    image_type = fields.Selection(
+        [("none", "None"), ("image", "Image"), ("icon", "Icon")],
+        string="Type",
+        required=False,
+        translate=False,
+        default="none",
```
yajo: After removing this option, we can leave the default default (:laughing:) `False`

```suggestion
```

website_menu_icon/models/website_menu.py
```diff
@@ -0,0 +1,60 @@
...
+            for child in node.child_id:
+                menu_node["children"].append(make_tree(child))
+            return menu_node
+
+        menu = (
+            menu_id
+            and self.browse(menu_id)
+            or self.env["website"].browse(website_id).menu_id
+        )
+        return make_tree(menu)
```
yajo: This code doesn't inherit from its parent. That is usually a blocking issue. :no_entry_sign: 

Please call `super()` and then alter the result as needed.

website_menu_icon/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,37 @@
...
+
+If you click the `Edit Menu Item` button enhanced menu item editing dialog.
+
+.. image:: https://raw.githubusercontent.com/Rad0van/website/website_menu_icon/website_menu_icon/static/description/edit_item.png
+
+Clicking on search icon will bring up `Select a Media` dialog that will allow
+for selecting either image or font awesome icon.
+
+The module itself does not alter rendering of the website menu. This can be achieved
+be extending menu rendering in your theme. Something like this:
```
yajo: If this is true, then the module is a bit useless, don't you think? :man_shrugging: 

You should add the template that extends menus to be rendered with icons. The template should be `customize_show="True"` to avoid disabling it per website.

website_menu_icon/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This module allows to assign image or icon to individual menu entries.
```
yajo:
```suggestion
This module allows to assign image or icon to individual website menu entries.
```

to avoid confusion with backend menus.

website_menu_icon/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
...
+These can be used in a custom theme to render menus with icons.
```
yajo:
```suggestion
These can be used in a custom theme to render menus with icons.
```

Again you should remove this sentence once you fix the problem and add the rendering here.

website_menu_icon/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
...
+The module also brings back option for opening menu item in new window
```
yajo: This seems something unrelated with the main module purpose. You should add this in a separate module instead (in a new PR of course).

website_menu_icon/static/src/xml/website.contentMenu.xml
```diff
@@ -0,0 +1,149 @@
...
+                    class="d-none"
+                    t-att-value="widget.data.image_type"
+                />
+            </div>
+        </t>
+    </div>
+    <!--
+    Had to copy and update the templates as I couldn't make the in-place extension work.
+    Also common mechanism didn't work for replacing t-call to "website.contentMenu.dialog.submenu.icon"
+-->
```
yajo: As you would expect, this is forbidden too :smirk: 

Try harder, it must be possible. Don't forget you're dealing with client-side QWeb here, which behaves different than server-side qweb: https://www.odoo.com/documentation/13.0/reference/qweb.html#javascript

website_menu_icon/static/src/xml/website.contentMenu.xml
```diff
@@ -0,0 +1,149 @@
...
+                    for="o_icon_dialog_label_input"
+                >Icon / Image</label>
+                <div class="col-md-9 input-group ml-auto">
+                    <input
+                        type="text"
+                        name="image"
+                        id="o_icon_dialog_label_input_value"
+                        class="form-control o_we_search"
+                        placeholder="Search icon or image"
+                        readonly="1"
```
yajo: Valid HTML boolean values are the name of the attribute or an empty string. Using `1` is confusing because it seems like using `0` would disable it, which is false. So:

```suggestion
                        readonly=""
```

website_menu_icon/views/website_menu_icon.xml
```diff
@@ -0,0 +1,35 @@
...
+        </record>
+    </data>
+    <template id="assets" inherit_id="web.assets_common">
+        <xpath expr="." position="inside">
+            <script
+                type="text/javascript"
+                src="/website_menu_icon/static/src/js/content.js"
+            />
+        </xpath>
+    </template>
```
yajo: This should exist in a `templates/assets.xml` file by convention.

website_menu_icon/static/src/js/content.js
```diff
@@ -0,0 +1,195 @@
+odoo.define("website_menu_icon.menu_entry_dialog", function(require) {
+    "use strict";
+
+    var contentMenu = require("website.contentMenu");
+    var MediaDialog = require("wysiwyg.widgets.MediaDialog");
+    var Dialog = require("web.Dialog");
+    var core = require("web.core");
+
+    var MyMenuEntryDialog = contentMenu.MenuEntryDialog.extend({
```
yajo: All will be much easier if you just `include` instead of `extend`ing this class:

```suggestion
    contentMenu.MenuEntryDialog.include({
```

This way, all `MenuEntryDialog` will automatically have your new code. This way, you can override the `save` event or any other you need to obtain the desired missing values.

Don't forget to call super in all overriden methods. Breaking inheritance is usually :no_entry_sign:.

website_menu_icon/static/src/js/content.js
```diff
@@ -0,0 +1,195 @@
...
+            } else {
+                Dialog.alert(null, "Could not find menu entry");
+            }
+        },
+    });
+
+    return {
+        MyMenuEntryDialog: MyMenuEntryDialog,
+        MyEditMenuDialog: MyEditMenuDialog,
+    };
```
yajo: Following the advice above, you probably don't need all this code.

website_menu_icon/static/src/xml/website.contentMenu.xml
```diff
@@ -0,0 +1,149 @@
...
+                        t-att-value="widget.data.image"
```
yajo: If data is false, this would render as `value="false"` which is wrong. Use `null` to avoid rendering the attribute.

```suggestion
                        t-att-value="widget.data.image || null"
```

website_menu_icon/static/src/xml/website.contentMenu.xml
```diff
@@ -0,0 +1,149 @@
...
+                            />
+                        </div>
+                        <div class="input-group-text o_we_delete_icon">
+                            <i
+                                class="fa fa-remove"
+                                title="Remove"
+                                role="img"
+                                aria-label="Remove"
+                            />
+                        </div>
```
yajo: These are not text addons, they are buttons. Also, replace the :x: for a :wastebasket: which is more analogous to the rest of odoo UI:

```suggestion
                        <button type="button" class="btn btn-primary o_we_search_icon">
                            <i
                                class="fa fa-search"
                                title="Search"
                                role="img"
                                aria-label="Search"
                            />
                        </button>
                        <button type="button" class="btn btn-secondary o_we_delete_icon">
                            <i
                                class="fa fa-trash"
                                title="Remove"
                                role="img"
                                aria-label="Remove"
                            />
                        </button>
```

website_menu_icon/static/src/xml/website.contentMenu.xml
```diff
@@ -0,0 +1,149 @@
...
+<templates id="template" xml:space="preserve">
+    <div t-name="website.contentMenu.entry.dialog.icon" t-extend="wysiwyg.widgets.link">
+        <t t-jquery="form" t-operation="append">
+            <div class="form-group row">
+                <label
+                    class="col-form-label col-md-3"
+                    for="o_icon_dialog_label_input"
+                >Icon / Image</label>
+                <div class="col-md-9 input-group ml-auto">
+                    <input
```
yajo, line 67: It would be way more friendly to display a thumbnail of the chosen icon/image than the text of its path or class.

website_menu_icon/models/website_menu.py
```diff
@@ -0,0 +1,60 @@
...
+
+from odoo import api, fields, models
+
+
+class Menu(models.Model):
+
+    _inherit = "website.menu"
+
+    image = fields.Char("Icon", required=False, translate=True)
+    image_type = fields.Selection(
```
yajo: This can easily be confused with a raw binary `image` field, so better rename it. Also... translatable? :thinking: 

```suggestion
    icon = fields.Char()
    icon_type = fields.Selection(
```

yajo (changes requested):

Rad0van:
@Yajo thanx for comments/suggestions. This will take some time and I also have some questions regarding some. Will ask where the suggestions re. But thanx again.

website_menu_icon/models/website_menu.py
Rad0van: Would you be so kind and help me here? What I really need is to extend the inner method `def make_tree(node)` that return the set of attributes - to include my added `image` and `image_type`. As I am quite a newbie in Python - is there a way to do that?

website_menu_icon/readme/DESCRIPTION.rst
Rad0van: Well it is and is not unrelated at the same time. Without this the only way to have menu opened in new window is to change the attribute in the backend which seems somewhat not-intuitive. So I wanted to be able to configure all apsects of the menu here in the same place. It could be taken into it's own separate module but that seems to me somewhat not worth doing (separate module just for this).

website_menu_icon/static/src/xml/website.contentMenu.xml
Rad0van: Believe me, I've spent many hours trying to make work just that. Wanted the template neat and minimal. Specifically this example there:
`<t t-inherit="base.template" t-inherit-mode="extension">
    <xpath expr="//tr[1]" position="after">
        <tr><td>new cell</td></tr>
    </xpath>
</t>`
I wanted to use that on that sub-template to add the new stuff. Never got it working. Thought it was because it didn't have JS class representation. But it didn't work even for the main template. Was looking over sources of both 12.0 and 13.0 for use of this mechanism. It is just never used. In 14.0 it seems to be used widely but with OWL.

Any help/pointer here would be highly appreciated.

website_menu_icon/static/src/js/content.js
Rad0van: I remember initially doing that way. Than I experimented for many hours with template inheritance and tried everything I could. So this is probably some remaining artifact of that. Will change that.

website_menu_icon/static/src/js/content.js
Rad0van: Well this is needed to provide values of `image` and `image_type` to `menu.fields`. The hack in the comment refers to (my) inability to extend the `_getData` in a way it would return these values. My finding was that calling it's super from my method also calls `save` method. So I was unable to extend the result with `image` and `image_type` - if doing after super call, the `save` method was already called. I cannot put it into result before because that variable doesn't exist yet and call to super will create its own.
But again: all this may be the result of my newbieness in JS.

website_menu_icon/views/website_menu_icon.xml
Rad0van: I put it into views/assets.xml Couldn't find any other sample where assets.xml would exist in templates directory - only in views. Is that OK?

Rad0van:
@Yajo I reworked everything I could - I even got the template extension inheritance working :-) Only few things left that I wasn't able to fix as requested. I would kindly ask you if you could provide me with some guidance on these. Thank you very much.

Rad0van:
@Yajo would you care to have a look and be so kind and help me resolve the remaining issues? Thank you very much.

website_menu_icon/__manifest__.py
```diff
@@ -0,0 +1,19 @@
+# Copyright 2020 Radovan Skolnik <radovan@skolnik.info>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+
+{
+    "name": "Website Menu Icons",
+    "category": "Website",
+    "license": "AGPL-3",
+    "author": "Radovan Skolnik," "Odoo Community Association (OCA)",
```
yajo:
```suggestion
    "author": "Radovan Skolnik, Odoo Community Association (OCA)",
```

website_menu_icon/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    "version": "13.0.1.0.0",
+    "summary": """
+Provides possibility to define icons/images for menus.
+These can be then rendered in a custom theme.
+It also brings back option for opening new window.
+    """,
```
yajo, line 15: A good rule of thumb is to write summaries of 1 line. If you can't then try to summarize more!

Not a blocker.

website_menu_icon/__manifest__.py
```diff
@@ -0,0 +1,19 @@
...
+    "depends": ["website", "web_editor"],
```
yajo:
```suggestion
    "depends": ["website"],
```

web_editor is implicit.

website_menu_icon/models/website_menu.py
```diff
@@ -0,0 +1,55 @@
...
+            for child in node.child_id:
+                menu_node["children"].append(make_tree(child))
+            return menu_node
+
+        menu = (
+            menu_id
+            and self.browse(menu_id)
+            or self.env["website"].browse(website_id).menu_id
+        )
+        return make_tree(menu)
```
yajo: You must call `super()`. If you need a different structure, instead of re-building it, use a different approach, such as getting super's result and traversing all elements to modify their structure as needed.

website_menu_icon/readme/CONFIGURE.rst
```diff
@@ -0,0 +1,20 @@
+Go to each `Website`, click `Edit` and click on menu item. The confirmation
```
yajo: This is RST and code is double-backticked:

```
Go to each ``Website``, click ``Edit`` and click on menu item. The confirmation
```



However... This isn't code, right? Then you better use italics:

```suggestion
Go to each *Website*, click *Edit* and click on menu item. The confirmation
```

The same in the rest of the fragments.

yajo (changes requested):

Rad0van:
@Yajo finally got to finish this one. Was easier than I thought year ago when I was starting with Odoo/Python. Please re-review. Who else could provide review? Thanx.

Rad0van:
@yajo, @ypapouin, @pedrobaeza, @chienandalu - would any of you be so kind and review please? Thanx.

Rad0van:
Happy New year to everyone. Trying to revive this. @Yajo @Tardo @chienandalu @pedrobaeza @victoralmau @ernestotejeda would you please care to have a look? Thank you.

max3903 (approved):

chienandalu (commented):
Can you rebase to main branch to refresh checks?

Rad0van:
> Can you rebase to main branch to refresh checks?

Done hopefully. Also squashed commits...

chienandalu (approved):
Tested :+1: Thanks!

Rad0van:
@max3903 , @chienandalu , @yajo is this good to go? Can anyone merger this please?  Thank you.

yajo (approved):
/ocabot merge nobump

Rad0van:
> /ocabot merge nobump

Hi @yajo ! Was it supposed to merge it? Because it seems nothing happened...

PR #747, opened 2020-08-22
[12.0][ADD]website_hide_invoice

hpatelserpentcs:
**Added** 

>> website_hide_invoice


pedrobaeza:
The invoices in the portal doesn't require website (only portal), so you shouldn't call it this way and don't be hosted here.

marcelsavegnago:
Considering the non-dependency of the website module, it may be better to host the module at https://github.com/OCA/account-invoicing

dreispt:
Makes sense.
What about naming it `account_portal_hide_invoice`?

dreispt:
@hpatelserpentcs Can you please rename and move to https://github.com/OCA/account-invoicing ?

HaraldPanten:
> Makes sense.
> What about naming it `account_portal_hide_invoice`?

I like your `account_portal_hide_invoice` module name suggestion. It suits well 👍

PR #750, opened 2020-09-18
[FIX] website_megamenu: Don't show default submenu

Tardo:
This is a 'hidden' problem... only affects if you use themes that can reveal it.

cc @tecnativa


pedrobaeza (approved):
Is this happening in official v13?

Tardo:
> Is this happening in official v13?

No, 13.0 works fine.

chienandalu (approved):

PR #782, opened 2020-10-07
[FIX] website_legal_page: better visualization as required checkbox label

ernestotejeda:
cc @Tecnativa TT25963
@pedrobaeza @Tardo Could you review?

When this template is used as a label for a required checkbox and the user doesn't check it before the form is submitted, then this is how it currently looks:

image

 and this is how I propose it to look in those cases:

image

### DIFF
--- a/website_legal_page/views/reusable_templates.xml
+++ b/website_legal_page/views/reusable_templates.xml
@@ -3,9 +3,9 @@
     <!-- Base templates for submodules -->
     <template id="acceptance_full">
         <span class="acceptance_full">
-            <a
+            I accept <a
                 href="/legal"
-            >I accept the legal terms, the privacy policy &amp; conditions of this website.</a>
+            >the legal terms, the privacy policy &amp; conditions</a> of this website.
         </span>
     </template>
 </odoo>


pedrobaeza:
Can whole text be in red on have a red background color?

ernestotejeda:
> Can whole text be in red on have a red background color?

Then it would look the same as text that is not a link to another page. Perhaps it is good that the user sees that he must check it, but that he can also click on it and go to the page of legal terms.

pedrobaeza (approved):
OK, good for me

yostashiro:
I guess this will split up the string and make the translation harder for some languages (ref https://github.com/odoo/odoo/issues/29104)? Not an objection to the proposed change but just trying to provide a point of view.

marcelsavegnago (approved):

pedrobaeza:
@yostashiro good point, but I see no other way to make this highlighted. Do you have any alternative solution?

yostashiro:
> Do you have any alternative solution?

@pedrobaeza No, I do not have one myself. :sweat_smile: Therefore, I am personally fine if this PR is processed as is.

pedrobaeza:
Thanks for your comments. @ernestotejeda please amend the translations for all languages with this change before merging.

yajo:
Sorry! I'm late to the party.

The purpose of this string being un-split in v13 refactor was precisely to avoid the problem @yostashiro is mentioning. So, in general I'm :-1: with this change.

Just look at how broken it was for v12:

https://github.com/OCA/website/blob/862ba28642fb85f3c96d6dddaf74abd0ea8838ca/website_legal_page/i18n/es.po#L58-L66
https://github.com/OCA/website/blob/862ba28642fb85f3c96d6dddaf74abd0ea8838ca/website_legal_page/i18n/es.po#L163-L166
https://github.com/OCA/website/blob/862ba28642fb85f3c96d6dddaf74abd0ea8838ca/website_legal_page/i18n/es.po#L540-L558

Translating that is... :nauseated_face: 

So let me propose 2 alternative approaches:

1. Add a SCSS rule such as this pseudocode: 

   ```scss
   .form-control.has-error {
       a {
           @include text('error')
       }
   }
   ```

1.  Add an argument to the template:

    ```xml
    <t t-set="extra_classes" t-value="" />
    <a t-att-class="extra_classes">I accept the...</a>
    ```

Tardo:
Is the 'required' property not enough?

Simple demo: https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_checkbox_required2
Useful pseudo-classes: https://www.digitalocean.com/community/tutorials/css-styling-form-input-validity

yajo:
Probably not, because Odoo uses its own form input validation, not browser's native one, AFAICR. But I encourage you to test that, it might have changed over time 🤷🏼‍♂️

PR #788, opened 2020-10-26
[12.0][ADD] website_no_index module

ecino:
See readme of module to see how useful this module can be. Add option in routes to easily remove them for search engine results.


pedrobaeza:
Isn't this the same as web_no_crawler (https://odoo-community.org/shop/product/4392)?

Issue #803, opened 2020-11-17
[13.0] website_crm_privacy_policy bug xpath


vanderperre:
installation : XML parser don't find parent

suggestion

replace in templates/contactus.xml 

expr="//form/div/div/span[@id='o_website_form_result']/../.."
by
expr="//form//div/div/span[@id='o_website_form_result']/../.."

wytzejan:
Please see my comment in OCA#804.
Your suggestion will also work but with a base website_form installation it's not really needed in my opinion.

PR #801, opened 2020-11-13
[12.0][ADD] website_google_analytics_anonymize_ip

wytzejan:
This module adds the ability to enable the Google Analytics IP Anonymization feature.
The feature adds 'ga('set', 'anonymizeIp', true);' to the tracking-code script once enabled.

PR #804, opened 2020-11-18
[13.0] website_crm_privacy_policy bug xpath #803

vanderperre:
installation : XML parser don't find parent

suggestion

replace in templates/contactus.xml

expr="//form/div/div/span[@id='o_website_form_result']/../.."
by
expr="//form//div/div/span[@id='o_website_form_result']/../.."

### DIFF
--- a/website_crm_privacy_policy/templates/contactus.xml
+++ b/website_crm_privacy_policy/templates/contactus.xml
@@ -6,7 +6,7 @@
         name="Add privacy policy checkbox"
     >
         <xpath
-            expr="//form/div/div/span[@id='o_website_form_result']/../.."
+            expr="//form//div/div/span[@id='o_website_form_result']/../.."
             position="before"
         >
             <div class="form-group row form-field o_website_form_required_custom">


wytzejan (commented):
I don't think that it is really necessary. In runbot it matches the exact div with only one slash:
image

Maybe you have a custom form layout somewhere that causes the 'XML parser don't find parent' error?

vanderperre:
> I don't think that it is really necessary. In runbot it matches the exact div with only one slash:
> image
> 
> Maybe you have a custom form layout somewhere that causes the 'XML parser don't find parent' error?

ok BUT :

Ideally, it would simply be necessary to match the identifier but referencing to the two division is not useless
It's hard to explain to the client that they can't include a super division inside the form ...

wytzejan:
I understand. I'm curious about why you would like to include a super division inside the form. But anyway, your requested changes does not break anything so I will approve :)

And may I ask you to review my merge request (#801)? I would very much like that.

wytzejan (approved):

HaraldPanten (approved):

PR #808, opened 2020-12-02
[12.0][FIX] website_lazy_load_image: reintroduce missing doctype when recreating html

ypapouin:
This fix reintroduce the missing doctype when the html is recreated by the `lxml` parser.
When `website_lazy_load_image` module is installed, the `<!DOCTYPE html>` is no more set as a header.
It was reported in https://github.com/odoo/odoo/issues/44612

This issue can only by seen by a non-logged user, probably because when logged, an editor module is fixing this issue as a side-effect.

A backport fix for 11.0 will be necessary.


simahawk:
Thanks for your contrib! However, it has been fixed here already https://github.com/OCA/website/pull/719 but the are some issues w/ tests.
Maybe you can help to get that one finished?

PR #818, opened 2021-01-31
Right spanish cookie message

dftossem:
Actual message does not contain the right information and punctuation marks.


Abranes:
@dftossem translations are managed via [weblate](https://translation.odoo-community.org/)

PR #820, opened 2021-02-05
[FIX] website_no_crawler: robots.txt should not be translated

angelmoya:
robots.txt should not be translated

It was translated from weblate.

@claudiagn check it

Is there any mechanism to prevent weblate from translating it?


pedrobaeza:
Use t-translation="off"

angelmoya:
Thanks @pedrobaeza I will do

chienandalu:
I didn't see this. It can be closed since it's already patched in c75cf895d941365e4bdf7dd3a6e1dc4e412aacdf

pedrobaeza:
Yes, and the fix wasn't correct.

PR #822, opened 2021-02-08
14.0 mig website portal contact

chandni299:
Migration of module "website_portal_contact" to v14

PR #825, opened 2021-02-23
[14.0] [ADD] new module website_portal_receipt

primes2h:
This module adds sale and purchase receipts in the customer portal.

It depends on https://github.com/OCA/account-invoicing/pull/849


primes2h:
As per https://github.com/OCA/website/pull/822#issuecomment-775148281, this is the wrong repository. 

Reopened in https://github.com/OCA/account-invoicing/pull/873

PR #833, opened 2021-04-15
[FIX] website_no_crawler: don't translate

chienandalu:
robots.txt should not be translated in order to have any effect

fw port from c75cf895d941365e4bdf7dd3a6e1dc4e412aacdf

cc @Tecnativa TT29159


pedrobaeza (approved):

pedrobaeza:
Fast-tracking for fixing this annoying bug:

/ocabot merge patch

PR #839, opened 2021-06-02
[12.0] FIX website_lazy_load_image in case of XHTML

eLBati:
E.g. going to /blog/1/feed you would get

ValueError: Unicode strings with encoding declaration are not supported. Please use bytes input or XML fragments without declaration


simahawk (changes requested):
can you update tests to proof this? :)

PR #853, opened 2021-10-24
[ADD][13.0] Web Core Vitals: Critical CSS (for FCP/LCP improvement)

thomaspaulb:
This module can improve the Google PageSpeed score of individual website pages.

In particular, it improves the score for First Contentful Paint (FCP). It does so
by allowing to configure a blob of 'Critical CSS' for each page, which is the only
CSS that will be loaded before first paint. Any other CSS is delayed.

To use this module:

  * Go to Website -> Pages and select a page to optimize
  * Find out the public facing URL of this page
  * Generate Critical CSS for this URL
  * Paste the blob in the 'Critical CSS' field.

To generate critical CSS you have a couple of options:

  * Use a free online Critical CSS generator. At the time of writing
    there are: Sitelocity, Pegasaas, web.dev, Corewebvitals.io, ...
  * Install the npm [critical](https://github.com/addyosmani/critical)
    package and generate it on your local.

To test improvement of FCP score, use for example Lighthouse, which is
build into Chrome browser.

### DIFF
--- a/website_critical_css/models/ir_qweb.py
+++ b/website_critical_css/models/ir_qweb.py
@@ -1,0 +1,77 @@
+from odoo import models
+
+from odoo.addons.website.models.ir_qweb import AssetsBundleMultiWebsite
+
+
+class AssetsBundleCriticalCSS(AssetsBundleMultiWebsite):
+    def to_node(
+        self,
+        css=True,
+        js=True,
+        debug=False,
+        async_load=False,
+        defer_load=False,
+        lazy_load=False,
+    ):
+        response = super(AssetsBundleCriticalCSS, self).to_node(
+            css=css,
+            js=js,
+            debug=debug,
+            async_load=async_load,
+            defer_load=defer_load,
+            lazy_load=lazy_load,
+        )
+        if not (css and defer_load):
+            return response
+        new_response = []
+        for item in response:
+            if not (item and item[0] == "link"):
+                new_response.append(item)
+            else:
+                attr = item[1]
+                attr["media"] = "print"
+                attr["onload"] = "this.media='all'"
+                new_response.append(("link", attr, None))
+        return new_response
+
+
+class IrQWeb(models.AbstractModel):
+    _inherit = "ir.qweb"
+
+    def get_asset_bundle(self, xmlid, files, env=None):
+        return AssetsBundleCriticalCSS(xmlid, files, env=env)
+
+    def _get_asset_nodes(
+        self,
+        xmlid,
+        options,
+        css=True,
+        js=True,
+        debug=False,
+        async_load=False,
+        defer_load=False,
+        lazy_load=False,
+        values=None,
+    ):
+        has_critical_css = bool(
+            css
+            and isinstance(values, dict)
+            and values.get("main_object")
+            and values["main_object"]._name == "website.page"
+            and values["main_object"].critical_css
+        )
+        # toggle defer_load to True in case of css=True and has_critical_css.
+        # In core, defer_load is only in use for JS currently,
+        # so we can use it for CSS in order to bust tools.ormcache cache, instead
+        # of trying to override the tools.ormcache dependency.
+        return super(IrQWeb, self)._get_asset_nodes(
+            xmlid,
+            options,
+            css=css,
+            js=js,
+            debug=debug,
+            async_load=async_load,
+            defer_load=defer_load or has_critical_css,
+            lazy_load=lazy_load,
+            values=values,
+        )
--- a/website_critical_css/models/ir_qweb.py
+++ b/website_critical_css/models/ir_qweb.py
@@ -59,6 +59,7 @@ class IrQWeb(models.AbstractModel):
             and values.get("main_object")
             and values["main_object"]._name == "website.page"
             and values["main_object"].critical_css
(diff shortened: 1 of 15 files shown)


website_critical_css/views/templates.xml
```diff
@@ -0,0 +1,10 @@
+<odoo>
+    <template id="add_critical_css" inherit_id="website.user_navbar" name="Add Critical CSS">
+
+        <xpath expr="//head/*[1]" position="before">
+            <t t-if="main_object and main_object._name == 'website.page' and main_object.critical_css">
+                <t t-raw="main_object.critical_css"/>
```
thomaspaulb: Perhaps include the `style` tag here.

website_critical_css/models/website_page.py
```diff
@@ -0,0 +1,9 @@
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class WebsitePage(models.Model):
+    _inherit = "website.page"
+
+    critical_css = fields.Text("Critical CSS")
```
thomaspaulb, line 10: And remove the `style` tag here on saving, if people are including one.

website_critical_css/readme/USAGE.rst
```diff
@@ -0,0 +1,16 @@
+To use this module:
+
+  * Go to Website -> Pages and select a page to optimize
```
thomaspaulb, line 3: Clear out any current contents of the 'Critical CSS' field.

website_critical_css/readme/USAGE.rst
```diff
@@ -0,0 +1,16 @@
...
+  * Find out the public facing URL of this page
+  * Generate Critical CSS for this URL
```
thomaspaulb: While selecting the right widths for Desktop and Mobile views.

website_critical_css/views/website_page.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
+    <record id="view_website_pages_form" model="ir.ui.view">
+        <field name="model">website.page</field>
+        <field name="inherit_id" ref="website.website_pages_form_view" />
+        <field name="arch" type="xml">
+            <field name="menu_ids" position="after">
+                <group>
+                    <field name="critical_css"/>
+                </group>
```
thomaspaulb, line 10: It would be nice if this is also editable in the frontend "Pages" manager - less clicking.

Kiplangatdan:
+1
Nice feature. Tested on v13 CE.

Kiplangatdan (approved):
Nice feature.
Tested on v13 CE.

website_critical_css/static/src/xml/page_properties.xml
```diff
@@ -0,0 +1,31 @@
...
+        <t t-jquery="#advances_page_info" t-operation="after">
+            <div role="tabpanel" id="critical_css_tab" class="tab-pane fade">
+                <div class="form-group row">
+                    <textarea
+                        class="form-control"
+                        id="critical_css"
+                        style="min-height: 200px"
+                    >
+                        <t t-if="widget.page.critical_css">
+                            <t t-esc="widget.page.critical_css" />
```
tarteo: You can shorten this by using `widget.page.critical_css or ''`

There is always some spaces / tabs in the textarea:
image
If you save it, it will always enable critical css (unless you manually remove the spaces everytime you want to change something like Page Name). You need to strip them when saving the page properties, or do a check in Python.

website_critical_css/models/ir_qweb.py
```diff
@@ -0,0 +1,77 @@
...
+        defer_load=False,
+        lazy_load=False,
+        values=None,
+    ):
+        has_critical_css = bool(
+            css
+            and isinstance(values, dict)
+            and values.get("main_object")
+            and values["main_object"]._name == "website.page"
+            and values["main_object"].critical_css
```
tarteo, line 61: You should also check if the user is an editor, otherwise the editor toolbar will flicker.

tarteo (changes requested):

website_critical_css/static/src/xml/page_properties.xml
thomaspaulb: Great comments, fixed and tested both.

tarteo (approved):
Thanks, LGTM!

thomaspaulb:
@gfcapalbo Note that `major` means "increase the major version by 1" so for example when you merged `website_prevent_cls" it became version `13.0.2.0.0`:

image

So, for next time:

- In case of a new module, you can use `/ocabot merge nobump`
- In case of a change whereby the whole concept of the module is refactored and manual migration is needed, `/ocabot merge major` is the one to use
- In case of a change whereby an upgrade is needed to apply changes in the module, `/ocabot merge minor` is the one to use
- In case of a change whereby just a Python restart is needed to apply changes in the module, `/ocabot merge patch` is the one to use

Then, if this is a version-to-version migration, you can add: `/ocabot migration` in order to do some specific processing

thomaspaulb:
Oh, oops :-) 

Well, and definitely don't do something like this haha

PR #847, opened 2021-10-04
14.0 mig website crm privacy policy

pedroguirao:

PR #857, opened 2021-11-08
[14.0][MIG] website_crm_privacy_policy: Migration to 14.0

victoralmau:
Migration to 14.0

Superseed: https://github.com/OCA/website/pull/847

Please @pedrobaeza and @joao-p-marques can you review it?

@Tecnativa TT30314


website_crm_privacy_policy/templates/contactus.xml
```diff
@@ -6,18 +6,18 @@
...
     >
         <xpath
-            expr="//form//div/div/span[@id='o_website_form_result']/../.."
+            expr="//form//div/div/span[@id='s_website_form_result']/.."
             position="before"
         >
-            <div class="form-group row form-field o_website_form_required_custom">
-                <div class="offset-lg-3 offset-md-4 col-md-8 col-lg-7">
+            <div class="form-group row form-field s_website_form_required_custom">
+                <div class="col-sm" style="margin-left: 200px;">
```
joao-p-marques: Shouldn't the different `col-(sm|md|lg)-X` attributes be kept to ensure responsiveness? Could you give more insight on why you remove this?

joao-p-marques (commented):
Code review

website_crm_privacy_policy/templates/contactus.xml
victoralmau: I don't really like this change.
Now, in v14 without making this change it would look like this:
old


In v13 the odoo code
https://github.com/odoo/odoo/blob/21eb4d1afdd0e45f3d9081a2ca7317a6eabeb392/addons/website_crm/views/website_crm_templates.xml#L44-L47
allowed to make that change.

Now in v14 the code is the following:
https://github.com/odoo/odoo/blob/d43b9923291dd103a80f552a38c395229a84d8db/addons/website_crm/views/website_crm_templates.xml#L83-L87

Any suggestions to improve this?

website_crm_privacy_policy/templates/contactus.xml
```diff
@@ -0,0 +1,31 @@
...
+                            name="privacy_policy"
+                            class="s_website_form_input form-check-input"
+                            required="required"
+                        />
+                        <label class="form-check-label" for="privacy_policy">
+                            <t t-call="website_legal_page.acceptance_full" />
+                        </label>
+                    </div>
+                </div>
+            </div>
```
Tardo: it looks acceptable to me putting it like this:
```suggestion
          <div class="form-group col-12 s_website_form_field s_website_form_required_custom">
                <div class="form-check">
                    <input
                        type="checkbox"
                        value="accepted"
                        id="privacy_policy"
                        name="privacy_policy"
                        class="s_website_form_input form-check-input"
                        required="required"
                    />
                    <label class="form-check-label" for="privacy_policy">
                        <t t-call="website_legal_page.acceptance_full" />
                    </label>
                </div>
            </div>
```

victoralmau: Thank you very much, it is perfect now with this change.

website_crm_privacy_policy/templates/contactus.xml
joao-p-marques: Now it looks much better with @Tardo's suggestion. Thanks :+1:

joao-p-marques (approved):

PR #866, opened 2021-12-30
[ADD]website_search_blog_bar: Search bar for blog entries

manuelregidor:
Bar to search blog entries by title, content and/or manual excerpt.

### DIFF
--- a/website_search_blog_bar/views/search_blog.xml
+++ b/website_search_blog_bar/views/search_blog.xml
@@ -1,0 +1,67 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2021 Manuel Regidor <manuel.regidor@sygel.es>
+     License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html). -->
+<odoo>
+    <template id="website_search_blog_bar">
+        <form
+            t-att-action="'/blog/%s/search_content' % slug(blog) if blog else '/blog/search_content'"
+            method="get"
+        >
+            <div class="input-group">
+                <input
+                    type="text"
+                    name="search_val"
+                    class="search-query form-control oe_search_box"
+                    placeholder="Search in blog"
+                    t-att-value="search_val"
+                />
+                <div class="input-group-append">
+                    <button
+                        type='submit'
+                        title="Search"
+                        class="btn btn-primary oe_search_button"
+                    >
+                        <i class="fa fa-search" />
+                    </button>
+                </div>
+            </div>
+        </form>
+    </template>
+    <template
+        id="website_blog_search_bar_blogs_nav"
+        inherit_id="website_blog.blogs_nav"
+    >
+        <xpath expr="//div[@class='container']" position="inside">
+            <t
+                t-if="website.show_search_bar"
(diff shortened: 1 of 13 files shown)


website_search_blog_bar/models/res_config_settings.py
```diff
@@ -0,0 +1,23 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+from odoo import fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    blog_search_title = fields.Boolean(
```
legalsylvain: I dont understand Why it is an option on res company. I'd see better as an option of the end user making the research. Don't you think ?

manuelregidor: It is just a first approach. We wanted it to be as simple as possible. Of course, it can be improved in the future. Appart from that, what do you think of the module in general? Thank you

manuelregidor:
@tarteo @victoralmau @eLBati Could you review, please? Thank you.

website_search_blog_bar/models/res_config_settings.py
legalsylvain: Not a precise point of view. I'm really not a website expert.
I let other oca members make reviews.

HaraldPanten (approved):
Tested in our production environment. Works like a charm 👍

website_search_blog_bar/models/res_config_settings.py
```diff
@@ -0,0 +1,23 @@
...
+        string="Search by Title", related="company_id.blog_search_title", readonly=False
+    )
+    blog_search_content = fields.Boolean(
```
tarteo: These should be True by default. If all of these are False the search bar shouldn't be visible, or you can add another bool something like "Enable search bar".

website_search_blog_bar/models/res_config_settings.py
tarteo: For me it also makes more sense to let the visitor decide. It's OK this way for now it can be improved upon later, nonetheless the fields should be saved in `website` instead of `res.company`.

website_search_blog_bar/views/search_blog.xml
```diff
@@ -0,0 +1,62 @@
...
+        >
+            <div class="input-group">
+                <input
+                    type="text"
+                    name="search_val"
+                    class="search-query form-control oe_search_box"
+                    placeholder="Search in blog"
+                    t-att-value="search_val"
+                />
+                <button
```
tarteo: You should wrap this in a `.input-group-append` so the button extends the input.

Before:
image

After:
image

tarteo (changes requested):
Nice functionality. Code LGTM. I have some functional comments which are pretty easy to fix no big issues. Thanks.

ValentinVinagre (approved):
LGTM 👍🏻

tarteo (approved):
Thanks!

gfcapalbo:
@manuelregidor looks good :+1 , you are just  missing some tests for having a merge.,
I will add the "help wanted" tag  
Perhaps someone wants to help out with the tests with an MR to this branch?

PR #873, opened 2022-01-17
[15.0][ADD] Matomo Tag Manager 

reneeduijzers:
This module introduces the Matomo Tag Manager JS script on all the website pages.
For all websites selected by the user settings/website (front-end) or website/configuration/websites.
The JS script gets inserted in the head of each page as first script, as desired by Matomo for analysis.

### DIFF
--- a/website_matomo_tag_manager/__init__.py
+++ b/website_matomo_tag_manager/__init__.py
@@ -1,0 +1,1 @@
+from . import models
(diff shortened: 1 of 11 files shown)


netfxtech:
Nice~ Thank you for this. I tested this in Odoo e2022-02-07 today and generated this in the logs

`2022-03-23 21:35:02,118 31 WARNING netfxtech.com odoo.addons.base.models.qweb: Found deprecated directive @t-raw='website._matomo_tag_manager_script()' in template 722. Replace by @t-out, and explicitely wrap content in Markup if necessary (which likely is not the case) `

steven-bouillon:
Nice~ Thank you for this. I generated this in the logs today when testing

`2022-03-23 21:35:02,118 31 WARNING odoo15e-db odoo.addons.base.models.qweb: Found deprecated directive @t-raw='website._matomo_tag_manager_script()' in template 722. Replace by @t-out, and explicitely wrap content in Markup if necessary (which likely is not the case) `

PR #879, opened 2022-02-23
[13.0][FIX] website_form_recaptcha: permissions

chienandalu:
fw of #876 

If the user doesn't have read permissions on the model, the form will
be blocked. It shouldn't interfere.

cc @Tecnativa TT34392 TT34402

ping @sergio-teruel @victoralmau

### DIFF
--- a/website_form_recaptcha/controllers/main.py
+++ b/website_form_recaptcha/controllers/main.py
@@ -27,7 +27,7 @@ class WebsiteForm(WebsiteForm):
     def extract_data(self, model, values):
         """ Inject ReCaptcha validation into pre-existing data extraction """
         res = super(WebsiteForm, self).extract_data(model, values)
-        if model.website_form_recaptcha:
+        if model.sudo().website_form_recaptcha:
             recaptcha_model = request.env["website.form.recaptcha"].sudo()
             recaptcha_model._validate_request(request, values)
         return res


website_form_recaptcha/controllers/main.py
```diff
@@ -27,7 +27,7 @@ def recaptcha_public(self):
     def extract_data(self, model, values):
         """ Inject ReCaptcha validation into pre-existing data extraction """
         res = super(WebsiteForm, self).extract_data(model, values)
-        if model.website_form_recaptcha:
+        if model.sudo().website_form_recaptcha:
```
ioans73, line 30:
```suggestion
        if model.with_user(SUPERUSER_ID).website_form_recaptcha:
```

It should be like that?

victoralmau (approved):

website_form_recaptcha/controllers/main.py
chienandalu: I don't think so :sweat_smile: `sudo` is only replaced by `with_user` when the access is made in behalf of such user. That doesn't matter here as there's no more context to give.

pedrobaeza: Yes, `sudo` is the way to do it, as no impersonation is needed, just bypass security rules.

pedrobaeza (approved):

pedrobaeza:
Fast-tracking as already approved on previous version:

/ocabot merge patch

PR #881, opened 2022-03-01
[ADD]website_membership_gamification: Show user badges in website

manuelregidor:
New module that allows to show gamification badges on the Members page in website.


manuelregidor:
Hi @pedrobaeza 

While developing this module, we found something. There is a module named 'hr_gamification' which, as well as the one in this PR, inherits the module 'gamification'. The module 'hr_gamification' applies important changes in the way gamification features work. When 'hr_gamification' module is installed, only an employee can be selected to get a badge, and it is not permitted to select a user anymore. This feature involves a number of requirements and restrictions that generate incompatibilities when it is used along with the module in this PR. That is why we have develped an additional module (user_hr_gamification) that sorts those incompatibilities, so 'user_hr_gamification' module should be installed in case we need to use both hr_gamification and website_membership_gamification. The only thing is that we don't know in which repo we should include the module that solves the incompatibilities. Any idea? Thank you.

website_membership_gamification/models/gamification_badge.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2022 Manuel Regidor <manuel.regidor@sygel.es>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class GamificationBadge(models.Model):
+    _inherit = "gamification.badge"
```
tarteo, line 8: Why don't you use `website.published.multi.mixin` here?

website_membership_gamification/views/res_partner_views.xml
```diff
@@ -0,0 +1,43 @@
...
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="base.view_partner_form" />
+        <field name="arch" type="xml">
+            <xpath expr="//notebook" position="inside">
+                <page
+                    string="Badges"
+                    attrs="{'invisible': [('has_related_users', '=', False)]}"
+                >
+                    <field name="has_related_users" invisible="1" />
+                    <button
```
tarteo, line 33: Is this meant to be a shortcut for this:
image

website_membership_gamification/wizard/gamification_badge_user_wizard.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright 2022 Manuel Regidor <manuel.regidor@sygel.es>
+     License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl). -->
+<odoo>
+    <record id="partner_view_badge_wizard_reward" model="ir.ui.view">
+        <field name="name">partner.gamification.badge.user.wizard.form</field>
+        <field name="model">gamification.badge.user.wizard</field>
+        <field name="arch" type="xml">
+            <form string="Reward with">
```
tarteo, line 9: Can't you inherit the view instead of overriding it? If you want a sperate form from the already existing you can also use `primary="True"`

website_membership_gamification/views/res_partner_views.xml
manuelregidor: It is a shortcut to grant a badge to a contact but it also shows all the badges assigned to a contact.

website_membership_gamification/wizard/gamification_badge_user_wizard.xml
manuelregidor: It doesn't override another view. Please note that this is the view used to assign a badge to a contact from the contact form view, which is different from the view used to assign a badge from a badge itself.

angelmoya:
How we can rebuild runboat? runboat for this branch was not launched, also is not on main runboat page...

manuelregidor:
> How we can rebuild runboat? runboat for this branch was not launched, also is not on main runboat page...

Done

HaraldPanten:
@pedrobaeza Could you have a look at this? We need some feedback to continue.

THX!

pedrobaeza:
Followed in #938 for a newer version.

PR #885, opened 2022-03-25
[13.0][FIX] website_form_recaptcha: Refactor to allow working with any form enabled model

chienandalu:
TLDR:

- [x] Every form enabled model has now recaptcha checked and injected and no further extensions are needed.
- [x] `website_crm_recaptcha` or any similar extension is now useless and can be uninstalled.

The implementation of website_form_recaptcha always checks the recaptcha validation either the recaptcha template code is in the form or not. This is problematic now as the form editor allows us to modify any form, even the main `website_crm` one and this clashes with that rigid design.

As the recaptcha is always checked we'll inject it allways for every model enabled for website forms.

This way, the recaptcha will be injected automatically without the need of defining new extensions to the forms.

Thus the module `website_crm_recaptcha` has no effect and can be uninstalled. Any similar custom extension won't have any effect as we dinamically remove any recaptcha code from the form and inject the right one dinamically after it's loaded.

cc @Tecnativa TT35359

please review @Tardo @ernestotejeda


pedrobaeza (approved):

chienandalu:
@Tardo Finally I took this approach which is more flexible I think.

ernestotejeda (approved):

website_form_recaptcha/controllers/main.py
```diff
@@ -27,7 +27,25 @@ def recaptcha_public(self):
...
+        multilang=False,
+    )
+    def has_website_form_access(self, model_name):
+        """Public method to decide whether to hook recaptcha or not"""
+        return json.dumps(
+            model_name
+            in request.env["ir.model"]
+            .sudo()
+            .search([("website_form_access", "=", True)])
+            .mapped("model")
```
pedrobaeza, line 50: I think you can save one iteration

```suggestion
            request.env["ir.model"]
            .sudo()
            .search([("website_form_access", "=", True)])
            .mapped("model")
```

chienandalu: Well, I want to check if model name is in this results list. As this is a public method I didn't want to directly use the variable from the post into the domain. Maybe too paranoid?

pedrobaeza: OK, right, I understand. Let's move on.

PR #886, opened 2022-04-28
[15.0][MIG]website_crm_quick_answer

manuelregidor:
Standard migration.


manuelregidor:
@HaraldPanten @ValentinVinagre

HaraldPanten (approved):
Functional review. LGTM

@rafaelbn

tarteo (changes requested):
The template doesn't get parsed correctly:
image

image

Am I doing something wrong?

tarteo:
Ah I think in 15 you should use qweb like this ```<t t-out="object.name or ''">```

manuelregidor:
> Ah I think in 15 you should use qweb like this `<t t-out="object.name or ''">`

Done

ValentinVinagre (approved):
LGTM 👍🏻

tarteo (changes requested):
The body is parsed correctly, but ```email_from``` is still not parsed correctly I get and i guess language neither.```${object.company_id.email}``` it should be ```{{ object.company_id.email }}``` looking at other templates and odoo/tools/rendering_tools.py

tarteo (approved):
Thanks!

rafaelbn (approved):

PR #890, opened 2022-05-11
[14.0][ADD] - website_info_hide: initial commit

AshishHirapara:
Initial commit of the module
Functionality: Hides the website info ('/website/info') page and redirects users to 404 not found page.


pedrobaeza:
This can be done with an option in the Odoo config file or in the website config (I don't remember well)

ioans73:
@AshishHirapara @pedrobaeza Yes, it can be configured from the website. If you access the `/website/info` page, you can disable it from `Customize - Show Odoo Information` menu

AshishHirapara:
> @AshishHirapara @pedrobaeza Yes, it can be configured from the web. If you access the `/website/info` page, you can disable it from the `Customize - Show Odoo Information` menu

@ioans73 you are right, but the goal is to disable the '/website/info' page by default when the website installs instead of updating the customization setting. So users don't have to remember to inactive the info page. If you think there is a better way to achieve this functionality, let me know.

And the reason why I redirect the users to 404 rather than just disabling the page by configuration is that I have to inherit multiple xml templates and a controller to achieve that, instead I have just redirected the user to 404, so in a way it reduces the extra code.

francesco-ooops (approved):
functionally it's ok

francesco-ooops:
@OCA/website-maintainers opinions on this implementation?

SicurSam (approved):
LGTM

ioans73:
> > @AshishHirapara @pedrobaeza Yes, it can be configured from the web. If you access the `/website/info` page, you can disable it from the `Customize - Show Odoo Information` menu
> 
> @ioans73 you are right, but the goal is to disable the '/website/info' page by default when the website installs instead of updating the customization setting. So users don't have to remember to inactive the info page. If you think there is a better way to achieve this functionality, let me know.
> 
> And the reason why I redirect the users to 404 rather than just disabling the page by configuration is that I have to inherit multiple xml templates and a controller to achieve that, instead I have just redirected the user to 404, so in a way it reduces the extra code.


But you would have to do a manual step to install it, right? In the end you have to also remember that you have to install that module

Maybe `"auto_install": True` should be set?

francesco-ooops:
> But you would have to do a manual step to install it, right? In the end you have to also remember that you have to install that module
> 
> Maybe `"auto_install": True` should be set?

@ioans73 we thought about this, but then anyone updating the website repo would find this module automatically installed on every instance where website app is installed (am I wrong?) and thought the solution too invasive (also considering that we're not simply disabling the feature, but applying a 404 in its place)

we also know there are modules allowing to set auto_install according to the environment, so it can be set according to needs

francesco-ooops:
@tarteo any opinion on this? :)

tarteo:
This is already a functionality of core Odoo. So I don't think this module is needed. Thanks for your effort and open source mindedness though!

image

francesco-ooops:
> This is already a functionality of core Odoo. So I don't think this module is needed. Thanks for your effort and open source mindedness though!

Hi Dennis, we're aware of the core functionality. 

The issue we're facing is that displaying /website/info page is **not** an opt-in but the default, so for every new customer website we create we have to remember to switch it off, in order to avoid indexing to the whole web which apps or themes are installed, like it happens [here](https://school-uniform.ovs.it/it/website/info), [here](https://italia.rapsodoo.com/website/info), [here](http://vansystem.eu/it_IT/website/info), [here](http://uivco.vb.it/website/info), [here](https://www.corporacionmayprod.com/website/info), etc

This module allows this page to be unreachable by default; if the customer wants to inform the world which modules they're using in their environment (I can't really see anyone asking for that) all it takes is uninstalling the module. 

On our machines we will set this module with autoinstall = 1, so we'll not have to worry anymore about this page (hopefully) :)

I ask you kindly to reassess the usefulness of this module.

francesco-ooops:
> And the reason why I redirect the users to 404 rather than just disabling the page by configuration is that I have to inherit multiple xml templates and a controller to achieve that, instead I have just redirected the user to 404, so in a way it reduces the extra code.

@tarteo this is the reason why the solution in this PR was adopted, if you think it should be implemented differently please let us know! thanks

tarteo:
@francesco-ooops I agree that it should be an opt-in, but I would rather see a PR on the https://github.com/odoo/odoo repo instead. I don't really understand why the /website/info page even exists, we always turn it off.

francesco-ooops:
> @francesco-ooops I agree that it should be an opt-in, but I would rather see a PR on the https://github.com/odoo/odoo repo instead. I don't really understand why the /website/info page even exists, we always turn it off.

@tarteo we could open a PR there, but even if someone would take it into consideration (unlikely in my experience so far), being this an odoo promotional page with the purpose of creating SEO backlinks (see for reference [this PR](https://github.com/odoo/odoo/pull/91131/commits/1e53fcb20e5c9c4061dcb15c01f5b1e508a11cae) from a couple weeks ago) I don't think they have any intention of turning it off by default.

It would be good to have a solution on OCA and leave the module behind if odoo changes its approach to /website/info page in future versions.

francesco-ooops:
@OCA/website-maintainers what do you think?

website_info_hide/controllers/main.py
```diff
@@ -0,0 +1,10 @@
+from odoo import http
+from odoo.http import request
+
+from odoo.addons.website.controllers.main import Website
+
+
+class WebsiteInfo(Website):
+    @http.route("/website/info", type="http", auth="public", website=True, sitemap=True)
```
leemannd, line 8:
```suggestion
    @http.route(auth="user")
```

website_info_hide/controllers/main.py
```diff
@@ -0,0 +1,10 @@
...
+    def website_info(self, **kwargs):
+        return request.render("website.page_404")
```
leemannd, line 10:
```suggestion
        return request.not_found()
```

website_info_hide/__manifest__.py
```diff
@@ -0,0 +1,18 @@
...
+    "version": "14.0.1.0.0",
+    "author": "Ashish Hirpara, Ooops, Odoo Community Association (OCA)",
+    "contributors": ["Ashish Hirpara"],
+    "maintainers": ["AshishHirapara"],
+    "website": "https://github.com/OCA/website",
+    "license": "AGPL-3",
+    "category": "Website",
+    "depends": ["website"],
+    "installable": True,
+    "auto_install": False,
```
leemannd, line 17:
```suggestion
    "auto_install": True,
```

leemannd:
This module can be helpfull, but IMO only if it auto installs himself.

With the actual code, you will face a minor issue. The odoo standard view allows to activate/deactivate the page. It's useless now and should be hidden.

francesco-ooops:
> This module can be helpfull, but IMO only if it auto installs himself.
> 
> With the actual code, you will face a minor issue. The odoo standard view allows to activate/deactivate the page. It's useless now and should be hidden.

Hi we discussed about this [here](https://github.com/OCA/website/pull/890#issuecomment-1128675144), we have no issues with setting it to auto (we also think it's a useless and annoying page) but this will mean forcing the 404 page on everyone with OCA/website repo modules in their instance and website installed, and IMHO that goes against OCA best practices - leaving user in control about which modules to install and which features to enable/disable

Eg. if I had an hypothetical customer who is really happy to display that page this module would auto install itself and disable the page, OCA would have created a problem for me with the customer - unlikely scenario but I don't think anybody likes to have features forced in their environment

leemannd:
@francesco-ooops You're right. Back in the time we faced also issues with modules that have the auto-install. So that's legit not to enforce this into the OCA. It can bring unwanted behaviour.

I was suggesting it because intalling the module manually does not provide anything better than using GUI to deactivate the page.

francesco-ooops:
> I was suggesting it because intalling the module manually does not provide anything better than using GUI to deactivate the page.

Yes, that is true. I don't have right now the name of the specific OCA module, but our sysadmin uses environment settings to change according to our needs whether a module should auto-install or not, so we're planning to use that feature for this module in order to have it on all "website" instances. 

We can add instructions for that process in module description in order to inform users about how to use this module.

website_info_hide/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1 @@
+Hides the website info ('/website/info') page and redirects users to 404 not found page.
```
francesco-ooops, line 1: If you want to have this module autoinstalling in your environment, we suggest using https://github.com/OCA/server-tools/tree/14.0/module_change_auto_install or creating a custom auto-installing module depending on this one.

gfcapalbo:
@AshishHirapara Thank you for your contribution! 
I think that auto install by default in the repo is not a good idea,
Installing it manually , would not provide much more functionality than the standard core option, agreed with @leemannd .


I will close this MR. but Thank you again! and please post your next idea!

francesco-ooops:
> 

Hi @gfcapalbo , I left many comments in this PR but looks like I haven't been able to explain convincingly what's the goal behind this module. 

In our environment, we have several customer DBs. On ALL of those dbs, I don't want the /website/info to be displayed, and I don't want to have someone manually disable it every time a new db is created, as the action can be easily automated.

The solution we've found is to create a module disabling the feature (this PR). 

Now, we agreed not to use auto-install for this module as we'd be forcing the disable to anyone syncing OCA website repo. Our proposal is to merge the module with auto-install = false, and indicating in description possible solutions to make it auto-installable locally, eg: 

- create an auto-installing module (eg: "francesco-ooops-initial-setup") depending on this one
- use a customization solution such as https://github.com/OCA/server-tools/tree/14.0/module_change_auto_install

hope this is clearer now :)

PR #891, opened 2022-05-16
[14.0][ADD] website_no_crawler_environment: initial commit

AshishHirapara:
By default, /robots.txt on an Odoo installation with a website module installed will allow indexing by web crawlers.
This module will overwrite the view that generates /robots.txt and won't allow index based on the server environment.


website_no_crawler_environment/views/disable_robots.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<odoo>
+    <template id="robots" inherit_id="website.robots">
+        <xpath expr="t" position="replace">
+            <t t-translation="off">
+                User-agent: *
+                <t t-if="server_env in ('preprod', 'stage')">Disallow: /</t>
```
GSLabIt, line 7: In this way you're limiting user to have `preprod` and `stage` env only. Better to generalize it, having `test` as default one.

GSLabIt (changes requested):

GSLabIt:
@AshishHirapara  Better to move to https://github.com/OCA/server-env repo

website_no_crawler_environment/views/disable_robots.xml
```diff
@@ -0,0 +1,14 @@
...
+                Sitemap:
+                <t t-esc="url_root" />
+                sitemap.xml
```
Tardo, line 10: where is sitemap.xml?

What are the advantages of this module over using X-Robots-Tag?

AshishHirapara: @Tardo I think the main advantage of this module is that users don't need to have knowledge of 'developer-specific' and complicated things like the X-Robots-Tag, and many users don't even know how to use the X-Robots-Tag within the odoo, so the module will be useful for them.
Also, this module provides a way to overwrite the view that generates /robots.txt and won't allow to index based on the server environment, so it will be easy to use if users have multiple environments and just restrict some of the environments from crawlers. So this is the main purpose of this module.

And if you think this can be done in a better way with X-Robots-Tag, then I am open to your suggestions. Let me know.
Thanks.

PR #895, opened 2022-05-19
[MIG] website_crm_privacy_policy: Migration to 15.0

ioans73:
Standard migration to 15.0


tarteo (approved):
Tested locally works perfectly. Only suggestion I would have is to open the privacy page on a new  tab instead, but I guess that's not part of the migration.

HaraldPanten (approved):

PR #902, opened 2022-06-10
[ADD] website_sendinblue: Sendinblue integration

ioans73:
This module integrates Odoo website with the marketing platform [Sendinblue](https://www.sendinblue.com)


ioans73:
In the end we are going to cancel it because we are going to use the Odoo functionality of inserting scripts in the head/body elements

PR #903, opened 2022-06-14
[13.0][FIX] website_cookiefirst: resource loading

chienandalu:
Depending on the assets loading we could be loading incorrectly the
cookies banner and even block Google Analytics scripts with no chance to
unlock it. Moving the cookiefirst snippet to the bottom we prevent such
situations.

cc @Tecnativa TT37488 TT37473

ping @ioans73


ioans73 (approved):
LGTM 👍🏻

pedrobaeza (approved):

PR #912, opened 2022-09-08
[15.0][ADD] website_menu_url_translate

nileshsheliya:


website_menu_url_translate/controllers/main.py
```diff
@@ -0,0 +1,56 @@
...
+
+class Website(Home):
+    @http.route(
+        "/website/lang/<lang>",
+        type="http",
+        auth="public",
+        website=True,
+        multilang=False,
+    )
+    def change_lang(self, lang, r="/", **kwargs):
```
tarteo, line 15: Is it possible to not completely override the super?

tarteo (changes requested):
I tested it locally but encountered some functional problems:

1. I created a page: "Hats"
2. Inserted a block on the page with the text "Hats"
3. Translated the url of the page in Deutsch to "/mutze" in the backend
4. Go to the "/hats" page
5. Changed language to Deutsch using the selector in the footer and I get redirected to: http://localhost:8069/de/hats instead of "/mutze" so it gives a 404
6. I try to use the menu item which is translated correctly to Deutsch
7. but then it switches back to English and I again get a 404.

Also it would be nice to be able to make the transalations from the frontend. Can you add this to roadmap (readme)?

website_menu_url_translate/controllers/main.py
dsolanki-initos: @tarteo, Yes, I've added the fix that is mentioned above and did not overwrite the change_lang function.
Could you please review it?

Thank you

PR #914, opened 2022-10-05
[12.0][ADD] portal_recaptcha

huguesdk:
move `portal_recaptcha` module from [coopiteasy/vertical-cooperative](https://github.com/coopiteasy/vertical-cooperative) to the oca.

PR #928, opened 2022-12-09
[14.0][ADD] portal_recaptcha

huguesdk:
move `portal_recaptcha` module from [coopiteasy/vertical-cooperative](https://github.com/coopiteasy/vertical-cooperative) to the oca.

this is the 14.0 version of #914.


carmenbianca (approved):
LGTM.

Small comment: 40b85be2bf769860055facc659d60aba53de0cf6 is a huge commit. I recommend doing smaller ones.

carmenbianca:
I bump :)

portal_recaptcha/views/portal_mixin.xml
```diff
@@ -0,0 +1,24 @@
...
+    License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+-->
+<odoo>
+    <template
+        id="assets_editor_recaptcha"
+        name="Google reCAPTCHA"
+        inherit_id="portal.frontend_layout"
+    >
+        <xpath expr="//main" position="inside">
+            <script src="https://www.google.com/recaptcha/api.js" />
```
pedrobaeza: Why loading reCaptcha JS in all the pages? I also don't see where is placed for the specific login page.

victor-champonnois: Thank you for the review. 

>Why loading reCaptcha JS in all the pages?

I fixed this, now the JS is only called when the widget template is called. 

>I also don't see where is placed for the specific login page.

I don't understand what you mean  Could you explain ?

portal_recaptcha/views/portal_mixin.xml
```diff
@@ -4,17 +4,9 @@
...
-        name="Google reCAPTCHA"
-        inherit_id="portal.frontend_layout"
-    >
-        <xpath expr="//main" position="inside">
-            <script src="https://www.google.com/recaptcha/api.js" />
-        </xpath>
-    </template>
 
     <template id="recaptcha_widget" name="reCAPTCHA widget">
+        <script src="https://www.google.com/recaptcha/api.js" />
```
huguesdk, line 9: is it allowed to have a `script` element there?

victor-champonnois: I don't know if it's allowed but it works ;)

victor-champonnois:
@huguesdk Actually I wonder if this module is still useful now that there is https://github.com/odoo/odoo/tree/14.0/addons/google_recaptcha
What do you think ?

huguesdk:
@victor-champonnois good point. i’ve compared the modules and have also taken a look into the history of this module and into the `website_form_recaptcha` module.

the main difference between the `google_recaptcha` standard module and this module is that `google_recaptcha` uses recaptcha v3 while this module uses recaptcha v2. as explained in [the docs](https://developers.google.com/recaptcha/docs/versions), the v2 allows to display a widget and provide a challenge to allow the user to prove that they are human, while the v3 does not display anything and only returns a score between 0 and 1 representing the likeliness that the user is a human.

this module comes originally from `website_recaptcha_reloaded` (probably ported from 8.0) from tech receptives. the original github account is not accessible anymore, but i’ve found a copy [here](https://github.com/GauravSahu/website_recaptcha/tree/8.0/website_recaptcha_reloaded). `website_recaptcha_reloaded` itself seems to be coming from [`website_recaptcha` from elico-corp](https://github.com/Elico-Corp/odoo-addons/tree/8.0/website_recaptcha).

i think that `website_recaptcha_reloaded` was used instead of `website_form_recaptcha` (which also uses recaptcha v2) because it is simpler and does not depend on `website_form` (only on `website`).

`website_form_recaptcha` has not been ported to 14.0. [the migration issue](https://github.com/OCA/website/issues/781) says to use `google_recaptcha` instead.

where to go from here? i don’t think using recaptcha v3 should be the only option. v2 and v3 cover different use-cases. only recaptcha v2 can provide a challenge to the user. moreover, `google_recaptcha` stores the recaptcha keys in `ir.config_parameter`s; they are thus global and prevent using multiple websites with different domains.

here is what i think we should do:

*   change this module to store the settings back in the website (like it used to do before being renamed to `portal_recaptcha`)
*   rename it to `website_recaptcha_v2` (`_v2` to make it clear that it handles only recaptcha v2)
*   improve it by taking inspiration from `google_recaptcha`, mainly async loading of recaptcha library and using recaptcha domain name instead of google to avoid cookies

huguesdk:
created #1030 as a follow-up to this.

PR #931, opened 2022-12-16
[14.0][FIX] website_no_crawler, config param id

GSLabIt:


GSLabIt:
@pedrobaeza quick fix for param id, sorry my fault 😢

website_no_crawler/data/data.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version='1.0' encoding='utf-8' ?>
 <odoo>
     <!-- Add add crawler configuration parameter -->
-    <record id="set_ribbon_color" model="ir.config_parameter">
+    <record id="set_crawler_disable" model="ir.config_parameter">
         <field name="key">webiste.crawler.disable</field>
```
pedrobaeza: Shouldn't this be `website.crawler.disable` instead of `webiste.crawler.disable`?

pedrobaeza: And why adding a system parameter?

GSLabIt: sorry not getting the point 😬

GSLabIt: got it, I'm blind ..

GSLabIt: > And why adding a system parameter?

Added to allow setting based on environment using server env stuff

pedrobaeza: I hate such options that overcomplicate all modules. Just don't install the module...

GSLabIt: you're right, but if you have a restore from a db coming from another env, let's say from production to stage, you've to remember to uninstall it each time ..or vice versa

pedrobaeza: Yes, but you automate that thing, not the other. On sysadmin side, bloating everything through environment variables makes difficult to follow a propagation problem or to trace things, and on Odoo side, to complicate code or to happen this kind of things.

GSLabIt: well, as u prefer ..not going to argue ..do whatever u think is the best.

pedrobaeza: Well, the module is already merged and approve by others, so I'm not going to undo it, but please take it into account for future "new features".

pedrobaeza (approved):
/ocabot merge minor

website_no_crawler/data/data.xml
GSLabIt: I'll surely do. Thanks

PR #936, opened 2023-01-06
[15.0][ADD] website_analytics_matomo

nileshsheliya:


astirpe (changes requested):
You should preserve the commit history. You can get it from this PR: https://github.com/OCA/website/pull/944

Moreover you should not replace the original author with your name.

astirpe:
Superseded by https://github.com/OCA/website/pull/950

PR #941, opened 2023-01-17
[15.0][MIG] website_snippet_marginless_gallery: Migration to version 15.0

pilarvargas-tecnativa:
cc @Tecnativa TT41034

@chienandalu @CarlosRoca13 please review


website_snippet_marginless_gallery/views/snippets.xml
```diff
@@ -0,0 +1,138 @@
...
+                        style="background-image: url('/website/static/src/img/library/bridge.jpg');"
+                    >
+                        <div>
+                            <h4 class="o_default_snippet_text">Lorem ipsum</h4>
+                            <p class="o_default_snippet_text mb0">
+                                Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do.
+                            </p>
+                        </div>
+                    </section>
+                    <section class="col-lg-3 col-12 pt32 pb32 bg-alpha">
```
chienandalu:
```suggestion
                    <section class="col-lg-3 col-12 pt32 pb32 bg-primary">
```

chienandalu (changes requested):
bg-alpha|beta|delta|gamma et al doesn't exist anymore so the colors in the snippet initial state doesn't show up anymore as they were in the previous version:

image

Yo can use any of the available current classes anyway:

https://github.com/odoo/odoo/blob/15.0/addons/web/static/lib/bootstrap/css/bootstrap.css#L6510-L6596

website_snippet_marginless_gallery/views/snippets.xml
pilarvargas-tecnativa: Class have been updated, thank you :)

CarlosRoca13 (approved):
Code and functional review :+1:

website_snippet_marginless_gallery/views/snippets.xml
CarlosRoca13: Will we need any migration script for this cases or this migration script will be set on a more generic module?

pedrobaeza: This is being handled in https://github.com/OCA/openupgradelib/pull/315 and https://github.com/OCA/OpenUpgrade/pull/3707

chienandalu (approved):
:+1:

PR #706, opened 2020-03-25
migrated website_analytics_piwik to 12.0

jans23:

PR #944, opened 2023-01-19
14.0 - [MIG] website_analytics_piwik

victor-champonnois:
- Migrating from 11.0 to 14.0
- Renaming piwik to matomo (`[since the service has changed name](https://matomo.org/piwik-pro-vs-matomo/)). Linked to this issue https://github.com/OCA/website/issues/417

@hbrunn @astirpe FYI

Warning : not fully functionally tested


astirpe:
@victor-champonnois nice, thanks!

I didn't test it yet, however there was a comment by @kwoot here https://github.com/OCA/website/pull/706#issuecomment-1007454971 saying that field `matomo_analytics_id` should be a Char and not an Integer, to allow compatibility with the Matomo Protect Track ID plugin.
When changing the type of the field from Integer to Char, for example when upgrading the module from an old version, you should ensure that an old integer value of `matomo_analytics_id` is automatically converted to a Char string. In case that doesn't happen automatically, a migration script would be required.

Moreover: when renaming fields (like you did) a best practice requires to create a migration script invoking the openupgrade `rename_fields()` method.
And probably also renaming the entire module requires some action in the openupgrade repo as well.

Not sure if a migration script is really required for this particular module. I think in most of the cases, when there's a major version upgrade, the Matomo configuration can be updated manually. But let's see what others say. Maybe @hbrunn has a different opinion about it.

setup/website_analytics_piwik/odoo/addons/website_analytics_piwik
```diff
@@ -0,0 +1 @@
+../../../../website_analytics_piwik
```
hbrunn: this seems a leftover from before the rename

website_analytics_matomo/models/website.py
```diff
@@ -0,0 +1,17 @@
+# Copyright 2015 Therp BV <http://therp.nl>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import models, fields
+
+
+class Website(models.Model):
+    _inherit = 'website'
+
+    has_matomo_analytics = fields.Boolean("Matomo Analytics")
```
hbrunn, line 10: for renamed fields, please include a migration script like https://github.com/OCA/e-commerce/blob/13.0/website_snippet_product_category/migrations/13.0.2.0.0/pre-migration.py, cf https://oca.github.io/OpenUpgrade/API.html#openupgradelib.openupgrade.rename_fields

hbrunn (changes requested):
please also add a PR to https://github.com/OCA/OpenUpgrade/blob/14.0/openupgrade_scripts/apriori.py#L57 to make the rename known to openupgrade.

And [pre-commit fails](https://github.com/OCA/website/actions/runs/3967154932/jobs/6798789081#step:7:73), looks like you should regenerate the readme

victor-champonnois:
>Not sure if a migration script is really required for this particular module. I think in most of the cases, when there's a major version upgrade, the Matomo configuration can be updated manually. But let's see what others say. Maybe @hbrunn has a different opinion about it.

That was what I was thinking. I think it's not necessary to write migration scripts, since the configuration can be easily done manually and the service provider has changed. Don't you agree @hbrunn ?

victor-champonnois:
>I didn't test it yet, however there was a comment by @kwoot here https://github.com/OCA/website/pull/706#issuecomment-1007454971 saying that field matomo_analytics_id should be a Char and not an Integer, to allow compatibility with the Matomo Protect Track ID plugin.

Done

victor-champonnois:
>And [pre-commit fails](https://github.com/OCA/website/actions/runs/3967154932/jobs/6798789081#step:7:73), looks like you should regenerate the readme

done

MiquelRForgeFlow:
> Don't you agree?

The problem is that if you have already configured it in v13, and you migrate, then later you don't want to configure it again, instead you should want the previous configuration still being valid.

website_analytics_matomo/views/templates.xml
```diff
@@ -0,0 +1,30 @@
...
+        <xpath expr="//script[@id='tracking_code']" position="after">
+            <t
+                t-if="website and website.has_matomo_analytics and website.matomo_analytics_host"
+            >
+                <script type="text/javascript">
+                    var _paq = _paq || [];
+                    _paq.push(['trackPageView']);
+                    _paq.push(['enableLinkTracking']);
+                    (function() {
+                    var u="//<t t-raw="website.matomo_analytics_host" />/";
```
astirpe, line 14: I think this line should be this way:

```suggestion
                    var u="<t t-raw="website.matomo_analytics_host" />";
```

website_analytics_matomo/views/templates.xml
```diff
@@ -0,0 +1,30 @@
...
+                    _paq.push(['setTrackerUrl', u+'matomo.php']);
+                    _paq.push(['setSiteId', <t t-raw="website.matomo_analytics_id" />]);
```
astirpe, line 16: Since you converted the `matomo_analytics_id` to Char, now the apex are probably needed:

```suggestion
                    _paq.push(['setSiteId', '<t t-raw="website.matomo_analytics_id" />']);
```

astirpe (changes requested):

website_analytics_matomo/views/templates.xml
astirpe: Sorry @victor-champonnois I just noticed this [note in the field](https://github.com/coopiteasy/website/blob/14.0-mig-website_analytics_piwik/website_analytics_matomo/models/website.py#L19-L20) definition and [in the readme](https://github.com/coopiteasy/website/blob/14.0-mig-website_analytics_piwik/website_analytics_matomo/readme/CONFIGURE.rst) and I realized that what I wrote in my comment was wrong. The line I asked you to change was actually correct and with the last change it's not working properly anymore. My bad. So the last change should be reverted back as it was before. I apologize for that.

I'm thinking of a way to improve the UI for this configuration process, especially for distracted users like me. But that's not part for this PR. So for that I will probably open a new PR in a few days.

victor-champonnois:
@astirpe Noted, thank you ! So I reverted the two changes that you proposed.

astirpe:
Thank you @victor-champonnois. Do you need help with the migration scripts?

astirpe (approved):
Thanks. Just squash your commits.

victor-champonnois:
@astirpe Thanks, I've pushed my script but I fail to test it. Here's what I do:
- I install the package on a DB with Odoo11
- I checkout the repos on Odoo14
- I run update all the modules. 

Right after the start of the update, I have an error :

```
````2023-01-30 12:06:03,960 171159 CRITICAL test_website_analytics_piwik odoo.service.server: Failed to initialize database `test_website_analytics_piwik`. 
Traceback (most recent call last):
  File "/home/victor/cie/odoo14/src/odoo/odoo/tools/convert.py", line 677, in _tag_root
    f(rec)
  File "/home/victor/cie/odoo14/src/odoo/odoo/tools/convert.py", line 580, in _tag_record
    record = model._load_records([data], self.mode == 'update')
  File "/home/victor/cie/odoo14/src/odoo/odoo/models.py", line 4231, in _load_records
    records = self._load_records_create([data['values'] for data in to_create])
  File "/home/victor/cie/odoo14/src/odoo/odoo/models.py", line 4152, in _load_records_create
    return self.create(values)
  File "<decorator-gen-108>", line 2, in create
  File "/home/victor/cie/odoo14/src/odoo/odoo/api.py", line 348, in _model_create_multi
    return create(self, arg)
  File "/home/victor/cie/odoo14/src/odoo/odoo/addons/base/models/res_users.py", line 1329, in create
    users = super(UsersView, self).create(new_vals_list)
  File "<decorator-gen-106>", line 2, in create
  File "/home/victor/cie/odoo14/src/odoo/odoo/api.py", line 348, in _model_create_multi
    return create(self, arg)
  File "/home/victor/cie/odoo14/src/odoo/odoo/addons/base/models/res_users.py", line 1082, in create
    return super(UsersImplied, self).create(vals_list)
  File "<decorator-gen-100>", line 2, in create
  File "/home/victor/cie/odoo14/src/odoo/odoo/api.py", line 348, in _model_create_multi
    return create(self, arg)
  File "/home/victor/cie/odoo14/src/odoo/odoo/addons/base/models/res_users.py", line 543, in create
    users = super(Users, self).create(vals_list)
  File "<decorator-gen-65>", line 2, in create
  File "/home/victor/cie/odoo14/src/odoo/odoo/api.py", line 348, in _model_create_multi
    return create(self, arg)
  File "/home/victor/cie/odoo14/src/odoo/odoo/addons/base/models/ir_fields.py", line 534, in create
    recs = super().create(vals_list)
  File "<decorator-gen-13>", line 2, in create
  File "/home/victor/cie/odoo14/src/odoo/odoo/api.py", line 348, in _model_create_multi
    return create(self, arg)
  File "/home/victor/cie/odoo14/src/odoo/odoo/models.py", line 3876, in create
    records = self._create(data_list)
  File "/home/victor/cie/odoo14/src/odoo/odoo/models.py", line 3982, in _create
    cr.execute(query, params)
  File "<decorator-gen-3>", line 2, in execute
  File "/home/victor/cie/odoo14/src/odoo/odoo/sql_db.py", line 101, in check
    return f(self, *args, **kwargs)
  File "/home/victor/cie/odoo14/src/odoo/odoo/sql_db.py", line 300, in execute
    res = self._obj.execute(query, params)
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "res_users_login_key"
DETAIL:  Key (login)=(admin) already exists.
```
This seems related to https://github.com/OCA/OpenUpgrade/issues/183, however I cannot just delete the admin user and move on, since so many tables depends on the res_users table.

MiquelRForgeFlow (approved):
Migration scripts are correct. Please, squash your commits.

astirpe:
You should try from V13 to V14 since openupgrade can only work one version at a time

victor-champonnois:
commits squashed

MiquelRForgeFlow:
ping @hbrunn

hbrunn (approved):
just fwi: you don't need to migrate renamed fields in the settings wizard because they don't carry data, but it also doesn't hurt

PR #946, opened 2023-01-26
[16.0][MIG] website_legal_page: migration to 16.0

miguel-S73:
Standard migration


AungKoKoLin1997 (approved):

website_legal_page/README.rst
```diff
@@ -0,0 +1,140 @@
...
+
+Usage
+=====
+
+Website editor can change easily any text of these pages using website builder.
+
+If you install this module after 'website_sale', it will be merge the content
+of 'website_sale.terms' into a new page.
+
+This module hard-redirect '/shop/terms' to '/legal'.
```
astirpe: I didn't test the full functionality but I just tried it on runbot and I noticed that page `/shop/terms` shows a 404 Page Not Found error.

Page `/terms` seems working instead, but is displaying the standard Odoo terms page and is not redirecting to `/legal`.

The standard Odoo `/terms` page can be changed using website builder, as well as the `/legal` page of this module. 
So either I don't understand what is this module needed for or this module is not needed in Odoo V16 anymore.

astirpe (changes requested):

website_legal_page/README.rst
NICO-SOLUTIONS: From my POV these should be two different sections... the terms page shows legal terms and conditions for dealing with the company. The legal page is for imprint... 
THis is how i should interprete the sections.

On the other hand i can confirm, that the hard redirection from /shop/terms -> legal is not working properly.
functionally tested...
But that would counter speak to my conclusion.

astirpe: If terms and legal are two different concepts, then why does this module try to redirect from  `/shop/terms` to `/legal`? I'm even more confused now... :grin:

pedrobaeza: This module unifies all the terms and legal stuff in the same page (that you have redacted when installing `website_legal_page`), and enforces to accept it when doing the checkout.

astirpe: Tested on runbot:

- in the webshop, when viewing a product there's a link to the standard `/terms` page but not to the `/legal` page
- when processing the checkout, the `/legal` page does not appear and there is no enforcing to accept it
- standard Odoo has the checkbox "I agree to the terms & conditions" on checkout, it can be easily enabled with website builder
 
So at the end the standard Odoo covers all the functionalities of this module.

pedrobaeza: @chienandalu @pilarvargas-tecnativa you dealt with this in v15. Can you refresh why is needed?

pilarvargas-tecnativa:
ping @CarlosRoca13

website_legal_page/README.rst
CarlosRoca13: Try the test without installing account and you will see that this terms page does not exist. So I think this module is needed... 

Imagine that you are managing a blog and you want to make accept the legal terms to your users, you will not be able if you don't install account too, but you don't need all the features added on account...

NICO-SOLUTIONS: > If terms and legal are two different concepts, then why does this module try to redirect from  `/shop/terms` to `/legal`? I'm even more confused now... :grin: 

That's what I meant. For me it would make more sense to the redirect from /shop/terms to /terms.. and have the legal page separate. 
But i guess the approach in this module is different.

miguel-S73:
I have updated my PR with the last changes in this module in odoo 15.0

- https://github.com/OCA/website/pull/955

miguel-S73:
@CarlosRoca13 
How do you see last changes?

CarlosRoca13 (approved):

pedrobaeza:
Finally this was polished in v15, and transferred here, so I think this can cover a specific use case. Merging then:

/ocabot migration website_legal_page
/ocabot merge nobump

PR #948, opened 2023-01-28
[15.0][ADD] website_snippet_dynamic_follow

espo-tony:
This module introduces a "Dynamic Follow" snippet to the Website building blocks. This snippet will be dynamically populated with the social media URLs and icons configured in the website settings.


pedrobaeza:
What do you think about calling it `website_snippet_dynamic_social_media` for following the current nomenclature on other places of Odoo?

espo-tony:
@pedrobaeza 
I've thought about it, but it feels a little bit misleading for two reasons:

1. Across Odoo, social media icons are displayed for two different reasons: there are the "follow-us" icons, which point to a URL defined by the website owner (likely to a social media account main page) and the "share-on" icons, which allow the website visitor to share a specific website content (e.g.: a product, an event, a job offer, etc.) on its own social media. This module improves the former and has no impact on the latter.
2. Even if I feel the most natural use of this module is to share Social Media URL, the functionality proposed is, in fact, generic enough to be adapted to any kind of URL: I can see a website builder using this snippet for sharing generic URL (including mailto URLs) as well as Social Media URLs.

However, I'm of course open to rename this module, if there is a feeling that a different name better describes it or is more consistent with other modules' names.

pedrobaeza:
OK, understood. Then according what you say and seeing the code, it can be renamed to `website_snippet_dynamic_link`, as you can put any URL on the records.

espo-tony:
@pedrobaeza 
I will do it. Do you also suggest I rename the new model introduced in this module: `website.social.media.follow` into something more generic? `website.dynamic.link.item` maybe ?

pedrobaeza:
Yeah, it makes sense

espo-tony:
@pedrobaeza 
Module and model renamed

pedrobaeza (approved):
Code and functional review

astirpe (approved):

pedrobaeza:
Can you please squash all the commits into one?

espo-tony:
@pedrobaeza 
All commits squashed into one.

PR #386, opened 2017-10-02
[MIG][10.0] website_openstreetmap module

mFlayyan:

PR #949, opened 2023-01-29
[15.0][ADD] website_openstreetmap

espo-tony:
This module introduces a new website snippet "OpenStretMap" to drag and drop in your website.

It also implements a method on "res.partner" to return the OpenStreetMap link computed from the partner address. This method is a technical feature: it doesn't introduce any functionality by itself, but it can be used by other modules.


pedrobaeza:
We have in OCA the module `partner_external_map` that does the function you mention, so I think it's better to restrict this module to the snippet, and call it `website_snippet_openstreetmap`. Note that there were an old one in #386

espo-tony:
@pedrobaeza 
I removed the duplicated functionality and renamed the module as you suggested

website_snippet_openstreetmap/__manifest__.py
```diff
@@ -0,0 +1,28 @@
+# Copyright 2023 Onestein (<http://www.onestein.eu>)
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+{
+    "name": "Website Snippet OpenStreetMap",
+    "summary": "Website Snippet OpenStreetMap",
```
pedrobaeza: Don't put this key if the contain is the same as the name.

website_snippet_openstreetmap/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+This module introduces a new website snippet "OpenStretMap" to drag and drop in
+your website.
+
+It also implements a method on "res.partner" to return the OpenStreetMap link
```
pedrobaeza: This is outdated I suppose.

website_snippet_openstreetmap/readme/CONFIGURE.rst
```diff
@@ -0,0 +1 @@
+This module doesn't require any configuration.
```
pedrobaeza: Don't create this file if there's nothing to configure.

espo-tony:
@pedrobaeza 
Thanks for your review. I removed the unneeded texts.

pedrobaeza:
Trying it functionally, several comments:

- Initially, the map is not rendered with the default address provided. Trying with other addresses, it may be the default provided that doesn't locate anything.
- The block is not an inner one, and it's not possible to change the width. It always takes full width.
- It doesn't allow as well to put a top/bottom padding.

espo-tony:
@pedrobaeza 

This snippet takes, by default, the website company address. Indeed, the issue with it is that the company address provided by default from Odoo in a new DB: `250 Executive Park Blvd, Suite 3400 San Francisco California (US)` doesn't return any result in OpenStreetMap. As you can check here: https://www.openstreetmap.org/search?query=250%20Executive%20Park%20Blvd%2C%20Suite%203400%20San%20Francisco%20California%20(US)#map=19/45.47938/4.36428

If you set a different address to your company, the block should work correctly.

About the block, I tried to stick as much as possible to the standard Odoo "Map" snippet. Compare this (standard Map block):

Screenshot from 2023-02-01 21-32-51

and this (OpenStreetMap block):

Screenshot from 2023-02-01 21-32-57

Also, it should be possible to put it into other blocks:

Screenshot from 2023-02-01 21-39-05

pedrobaeza (approved):
What do you think about including demo data for modifying such company address removing `, Suite` that is the part that is breaking the search?

espo-tony:
@pedrobaeza 
I changed the demo data as you suggested.

tarteo (approved):

astirpe (approved):

espo-tony:
@pedrobaeza can I do anything to push this PR?

PR #950, opened 2023-01-30
[15.0][MIG/IMP] website_analytics_matomo

astirpe:
- [x] standard migration
- [x] add schema (https://) to matomo host if missing (to simplify configuration)
- [x] add "User ID" feature
- [x] add "Heartbeat" feature
- [x] add note regarding Event Tracking
- [x] add tests


astirpe:
@hbrunn I'm adding new stuff to this module, would you give a look?

tarteo (approved):
LGTM!

hbrunn (approved):
i did not however check the js with the matomo docs and just trust that works

website_analytics_matomo/models/website.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        for website in self.filtered(
+            lambda w: not w.matomo_analytics_host or not w.matomo_enable_userid
+        ):
+            website.matomo_get_userid = ""
+        for website in self.filtered(
+            lambda w: w.matomo_analytics_host and w.matomo_enable_userid
+        ):
+            if self.env.user != website.user_id:  # current user is logged in
+                website.matomo_get_userid = self.env.user.login
```
hbrunn: this is more a question which is why i put it out of the review: does matomo itself do some masquerading of userids? and is the value relevant for something or does it just need to be unique? given it's very probable those will be email addresses, and they might go somewhere else, we needlessly leak potentially sensitive data, and think of gdpr. why not by default: 

```suggestion
                website.matomo_get_userid = str(self.env.user.id)
```

astirpe: Matomo doesn't do any masquerading of userids, so the userid passed here is what is shown in Matomo. I think you are right and I'm going to adopt your suggestion. Thank you!

PR #951, opened 2023-01-31
[16.0][MIG] website_analytics_matomo

astirpe:


website_analytics_matomo/models/website.py
```diff
@@ -61,6 +62,6 @@ def _compute_matomo_userid(self):
             lambda w: w.matomo_analytics_host and w.matomo_enable_userid
         ):
             if self.env.user != website.user_id:  # current user is logged in
-                website.matomo_get_userid = self.env.user.login
+                website.matomo_get_userid = str(self.env.user.id)
```
chienandalu, line 65: Wouldn't this break past trackings?

astirpe: This is just the porting of the code from V15

chienandalu: I see. Then maybe rebasing your migration to that branch would remove any doubt, otherwise it looks like a change that came out of nowhere :smile:

astirpe:
Continues in https://github.com/OCA/website/pull/991

website_analytics_matomo/models/website.py
astirpe: Right. I opened https://github.com/OCA/website/pull/991 so it should be more clear

PR #956, opened 2023-02-23
[16.0][ADD] website_security

tarteo:
By default Odoo has two groups: *Restricted Editor*, and *Editor and Designer*.
This module adds an extra group on top of it: `Full Access`.
This group has some additional rights to the aforementioned groups:

* Full access to the website settings (Website > Configuration > Settings)
* Installing and updating themes
* Creating new websites
* Install languages

These rights are (in core Odoo) only accessible for users in the `Administration / Settings` group
which also grants access to critical models e.g. views, users, groups, and system parameters.

screen1


pedrobaeza:
A bit off-topic, but I would love to have a permission that only allows to manipulate `oe_structure` parts or custom pages, not standard templates (with XML-ID).

tarteo:
Hmm pre-commit check doesn't seem to be working.

PR #957, opened 2023-02-27
[ADD] website_url_sitemap_exclusion

SirTakobi:


SirTakobi:
CI should be fixed after https://github.com/OCA/website/pull/958

eLBati (approved):
Thanks

website_url_sitemap_exclusion/models/website.py
```diff
@@ -0,0 +1,29 @@
...
+        comodel_name="website.forbidden.url",
+        inverse_name="website_id",
+        string="Forbidden URLs regex",
+    )
+
+    @api.multi
+    def enumerate_pages(self, query_string=None, force=False):
+        pages = super().enumerate_pages(query_string=query_string, force=force)
+        forbidden_url_regexes = self.mapped("forbidden_url_ids.regex")
+        forbidden_url_compiled_regexes = list(map(re.compile, forbidden_url_regexes))
```
simahawk, line 21:
```suggestion
        forbidden_url_compiled_regexes = map(re.compile, forbidden_url_regexes)
```
no need to convert to list

website_url_sitemap_exclusion/models/website.py
```diff
@@ -0,0 +1,29 @@
...
+        for page in pages:
+            page_url = page["loc"]
+            for forbidden_url_compiled_regex in forbidden_url_compiled_regexes:
+                is_url_forbidden = forbidden_url_compiled_regex.match(page_url)
+                if is_url_forbidden:
+                    break
+            else:
+                yield page
```
simahawk:
```suggestion
            if not self._enumerate_pages_forbidden(page):
                yield page

    def _enumerate_pages_forbidden(self, page):
        page_url = page["loc"]
        for forbidden_url_compiled_regex in forbidden_url_compiled_regexes:
            if forbidden_url_compiled_regex.match(page_url):
                return True
```

website_url_sitemap_exclusion/models/website_forbidden_url.py
```diff
@@ -0,0 +1,41 @@
...
+#  License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+import re
+
+from odoo import _, api, fields, models
+from odoo.exceptions import ValidationError
+
+
+class WebsiteForbiddenUrls(models.Model):
+    _name = "website.forbidden.url"
```
simahawk, line 11: Maybe matter of taste but I would use "blacklist(ed)" instead of "forbidden".

simahawk (commented):
Non blocking remarks

website_url_sitemap_exclusion/models/website.py
SirTakobi: It is actually needed to convert to a `list` because otherwise the `map` generator is exhausted after the first loop.
See how the variable `m` is empty after the first loop:
```python
>>> l = [1, 2, 3]
>>> m = map(print, l)
>>> for p in m:
...     pass
... 
1
2
3
>>> for p in m:
...     pass
... 
>>>
```

website_url_sitemap_exclusion/models/website.py
SirTakobi: I am not really a fan of `return`ing in a loop so I have done something similar to what you suggested, please have a look at the new changes.

website_url_sitemap_exclusion/models/website_forbidden_url.py
SirTakobi: I wish we had thought of it from the beginning :)
Since this would require a migration and some renaming throughout the whole module, I'd rather leave it as is

SirTakobi (commented):
Thanks for the review!

website_url_sitemap_exclusion/models/website_forbidden_url.py
simahawk: maybe good for the roadmap? :wink:

simahawk (approved):

website_url_sitemap_exclusion/models/website_forbidden_url.py
SirTakobi: 👍🏻 added to the roadmap

OpenCode (approved):

PR #963, opened 2023-03-09
[15.0][ADD] website_form_require_legal: New module to add terms acceptance at website form level

CarlosRoca13:
cc @Tecnativa TT41895

please @pilarvargas-tecnativa @chienandalu review this


website_form_require_legal/tests/test_ui.py
```diff
@@ -0,0 +1,13 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html).
+
+import odoo.tests
+from odoo.tests import tagged
+
+
+@tagged("post_install", "-at_install")
+class TestUi(odoo.tests.HttpCase):
+    def test_admin_tour_marginless_gallery(self):
+        self.start_tour("/", "marginless_gallery", login="admin")
```
chienandalu: ?

website_form_require_legal/readme/USAGE.rst
```diff
@@ -0,0 +1,10 @@
+To use this module, you need to:
+
+* Edit any page in your website.
+* Add a form from the snippets.
+* Modify the parameters.
+* You will find the option "Legal terms acceptance".
+* If you select it you will be able to modify the text shown and the link to the
+  page from the text editor.
+
+Note: Remember to change the terms and conditions link to redirect to the desired page.
```
chienandalu, line 10: Should we add a ROADMAP declaring that this module is aimed to deprecate website_form_legal and its descendants?

chienandalu (changes requested):

website_form_require_legal/tests/test_ui.py
CarlosRoca13: Ooops copypaste problems :sweat_smile:

website_form_require_legal/readme/USAGE.rst
CarlosRoca13: Added :smile:

rvalyi:
Hello, I have nothing against such a module, but I would like to mention we had to requirement to present such legal terms in the bare Odoo sale portal, like when customer answers a quote email to come and sign the online order in the portal. I see you don't depend on sale here so there is probably no better way than what you did. Anyway just mentioning the use case in case it gives you some idea...

chienandalu:
@rvalyi That looks like a different module to me. The aim of this one is quite simple: to add this custom field to the website_form options. The one you're thinking would be a different extension for `portal` with no need of website.

pilarvargas-tecnativa (approved):
Tested locally, thanks

chienandalu (approved):
Code and functional review :+1:

pedrobaeza:
Great, now we need to change `crm_privacy_policy` to use this one, isn't it?

/ocabot merge nobump

CarlosRoca13:
> Great, now we need to change `crm_privacy_policy` to use this one, isn't it?
> 
> /ocabot merge nobump

Not need to do any change on `website_crm_privacy_policy` because now we can add the legal terms acceptance manually. So that module should be marked as deprecated, IDK if exists any way to do that

pedrobaeza:
Then the steps are:

- On OpenUpgrade, if `website_crm_privacy_policy` is installed, mark it to uninstall it, and mark this one to be installed.
- On the contact us view, inject the content to make this to work.
- On the README of `website_crm_privacy_policy`, indicate the deprecation content, making reference to this new one.

CarlosRoca13:
:+1: Thanks

PR #967, opened 2023-03-24
[15.0][FIX] website_crm_privacy_policy: Fixed error calling a template that does not exist

pilarvargas-tecnativa:
In the contact us form there was a call to the website_legal_page.acceptance_full template that no longer exists, it has been corrected so that it does not give error replicating that template directly.

cc @Tecnativa TT42194

@CarlosRoca13 @pedrobaeza please review :)


pedrobaeza (approved):

ioans73 (approved):

DucTruongKomit (approved):
Look good to me!

Issue #968, opened 2023-03-31
[14.0] [website_cookiefirst]


matteoopenf:
## Module
[website_cookiefirst](https://github.com/OCA/website/tree/14.0/website_cookiefirst)
## Describe the bug
try to translate page and obtain :
load could not load template
ValueError: L'elemento '<xpath expr="//footer//div/span[hasclass('o_footer_copyright_name')]">' non può esser localizzato nella vista genitore

View name: footer_custom
Error context:

## To Reproduce
**Affected versions**:

- [ ] 14.0  https://github.com/OCA/website/pull/969
- [ ] 16 ?

1. install  module
2. configure module
3. try to translate website , and have the error

**Expected behavior**
it's possibile translate

**Additional context**
Add any other context about the problem here. (e.g. OS, Python version, ...)

matteoopenf:
https://github.com/odoo/odoo/blob/9cc56fc3d3cdb43e393df0975e10192c3b10f455/addons/website/views/website_templates.xml#L1351 in this view not exist this class o_footer_copyright_name so the xpath is not exact

pedrobaeza:
Don't open an issue if you are going to open a PR.

PR #969, opened 2023-03-31
[14.0][FIX] website_cookiefirst: Bad inherited view

matteoopenf:
load could not load template
ValueError: L'elemento '' non può esser localizzato nella vista genitore https://github.com/OCA/website/issues/968

### DIFF
--- a/website_cookiefirst/views/portal_template.xml
+++ b/website_cookiefirst/views/portal_template.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
-    <template id="footer_custom" inherit_id="website.footer_custom">
+    <template id="footer_custom" inherit_id="website.layout">
         <xpath
-            expr="//footer//div/span[hasclass('o_footer_copyright_name')]"
+            expr="//footer//span[hasclass('o_footer_copyright_name','mr-2')]"
             position="after"
         >
             <span>-


pedrobaeza (commented):
I think you should change the inherited template:

https://github.com/odoo/odoo/blob/9cc56fc3d3cdb43e393df0975e10192c3b10f455/addons/website/views/website_templates.xml#L179

matteoopenf:
> I think you should change the inherited template:
> 
> https://github.com/odoo/odoo/blob/9cc56fc3d3cdb43e393df0975e10192c3b10f455/addons/website/views/website_templates.xml#L179

I check and I will update.
@pedrobaeza  thank for quick response

matteoopenf:
> I think you should change the inherited template:
> 
> https://github.com/odoo/odoo/blob/9cc56fc3d3cdb43e393df0975e10192c3b10f455/addons/website/views/website_templates.xml#L179

confirm I updated the Pr, thanks again

website_cookiefirst/views/portal_template.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
-    <template id="footer_custom" inherit_id="website.footer_custom">
+    <template id="layout" inherit_id="portal.frontend_layout">
```
pedrobaeza: For not losing the change when installing website, it should be:

```suggestion
    <template id="layout" inherit_id="website.layout">
```

pedrobaeza: Ah, and better to keep the same XML-ID for avoiding problems on migrations:

```suggestion
    <template id="footer_custom" inherit_id="website.layout">
```

matteoopenf: if I do that I have the previus error

pedrobaeza: Try with a less strict xpath `//footer//span...` instead of `//footer//div/span`

matteoopenf:
I squash the commit

matteoopenf:
@pedrobaeza  I See when I squash the commit I lost the coauthor of the commit, is a problem for you?

pedrobaeza:
Don't worry. No problem in this authorship. It's not a big deal in this case :smiley:

matteoopenf:
> Don't worry. No problem in this authorship. It's not a big deal in this case smiley

I readd the coauthor.
Thanks again for the help and suggestion and the patience

website_cookiefirst/views/portal_template.xml
```diff
@@ -1,10 +1,7 @@
 <?xml version="1.0" encoding="utf-8" ?>
 <odoo>
-    <template id="footer_custom" inherit_id="website.footer_custom">
-        <xpath
-            expr="//footer//div/span[hasclass('o_footer_copyright_name')]"
-            position="after"
-        >
+    <template id="footer_custom" inherit_id="website.layout">
+        <xpath expr="//footer//span/.." position="after">
```
pedrobaeza: When I put the `...` (3 dots), it was for not writing the rest of the expression, but I think you still need to use the `hasclass` part to select the exact element.

matteoopenf: ok I check I will come back

matteoopenf: i specify the xpath now

pedrobaeza (approved):

lorenzoallegrucci (approved):

pedrobaeza:
/ocabot merge minor

Please check if needed in upper versions.

PR #994, opened 2023-07-21
[16.0][ADD]website_rewrite_absolute_url: Module to add absolute url redirects

odooNextev:
By default Odoo doesn't allow to add absolute url redirects. I added "Is absolute url" field, if it's checked, then it will be possible to add an absolute url redirects. 
In check_url_to method I replicated default behaviours, but using routing_map.bind(rewrite.url_to) if it's an absolute url.

Example: If I want to redirect my page to google.com by default Odoo add a redirect to www.mywebsite.com/google.com
With this module you can correctly redirect to google.com


odooNextev:
@pedrobaeza What is your opinion about this pr?

pedrobaeza:
I can't say about the functionality, sorry. Reopening for now.

yajo (changes requested):
This module is unnecessary AFAICS.

Just make sure you add `http://` or `https://` at the beginning of the redirected URL and base odoo will work out of the box.

Something like this:
imagen

It will perform the expected absolute redirection, without needing this module.

odooNextev:
Thanks @yajo, It works properly without this module.

PR #996, opened 2023-07-25
[13.0][FIX] website_google_analytics_4 mime type

chienandalu:
In some casesGoogle won't perform the tracking. Looks like the mainoffender was thee mimetype. The recommendation is to omit the attribute in the script element when is js code though:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type

cc @Tecnativa TT44128

please review @pedrobaeza @miguel-S73

### DIFF
--- a/website_google_analytics_4/views/website_templates.xml
+++ b/website_google_analytics_4/views/website_templates.xml
@@ -8,17 +8,11 @@
         <xpath expr="//script[@id='tracking_code']" position="replace">
             <script
                 id="google_analytics_4_code"
-                type="text/plain"
-                data-cookiefirst-category="performance"
                 t-if="website and website.google_analytics_key and not editable"
                 async="1"
                 t-attf-src="https://www.googletagmanager.com/gtag/js?id={{ website.google_analytics_key }}"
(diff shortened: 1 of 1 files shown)


miguel-S73:
@chienandalu Following the European cookie regulations. By default I understand that any tracking must be disabled and only when the user accepts this tracking,  it should be activated. I use Cookiefirst/Cookiebot . It changes mimetype of scripts when user accepts tracking cookies.

chienandalu:
IMO this module should only replace the old tracking snippet (which has no mimetype). Anything beyond that should be addressed in the proper cookie module. Indeed, the cookiebot references should go away. They're harmless, but have no use if the customer doesn't use that service or use other like cookiefirst

miguel-S73:
@chienandalu I go along with you. Can you remove this attributes from script?

chienandalu:
Oks. Removed and commit message changed as well :)

miguel-S73 (approved):

pedrobaeza (approved):
/ocabot merge minor

PR #1003, opened 2023-08-08
[16.0][MIG] website_form_require_legal: Migration to version 16.0

pilarvargas-tecnativa:
cc @Tecnativa TT44376

@chienandalu @CarlosRoca13 please review


CarlosRoca13 (changes requested):
Something is wrong...

error_include

pilarvargas-tecnativa:
> Something is wrong...

I tried again locally, it should no longer happen

CarlosRoca13:
I have found another error: error_classlist

pilarvargas-tecnativa:
> I have found another error:

is a error in the website_crm_privacy_policy module, it is corrected at https://github.com/OCA/website/pull/1014

pilarvargas-tecnativa:
@CarlosRoca13 The error you were experiencing in runbot in the contact form is already fixed in the other module, however the functionality test has to be done by inserting a new form block as the other module already has the contact form customised and only allows editing the link. In local tests where the two modules do not coexist, the functionality is also applied to the contact form, so to check that it really works you would have to do it without the other module marked as "deprecated" installed.

AungKoKoLin1997 (approved):
Functional Review: LGTM

CarlosRoca13 (approved):

PR #1007, opened 2023-08-23
[15.0][ADD]website_sale_partner_firstname: New module to set lastname from portal

manuelregidor:
New module to set and edit lastname from portal.
This field is mandatory when the contact is B2C.


manuelregidor:
@HaraldPanten

HaraldPanten:
@manuelregidor  Only some comments:

- Could you add the version of the module in the PR title? (the commit message is OK).
- Could you add some more information in the USAGE document? Such as how portal users should do that.
- I'm not sure if `oca_dependencies.txt` has to be changed. Maybe @etobella could shed some light here. I don't really uderstand how does pip work here.

etobella:
No need to change oca-dependecies.txt, as it is generated automatically with pip :wink:

manuelregidor:
> @manuelregidor Only some comments:
> 
> * Could you add the version of the module in the PR title? (the commit message is OK).
> * Could you add some more information in the USAGE document? Such as how portal users should do that.
> * I'm not sure if `oca_dependencies.txt` has to be changed. Maybe @etobella could shed some light here. I don't really uderstand how does pip work here.

Done

HaraldPanten (approved):
Functional review. Tested in runboat. Works as expected.

@chienandalu could you add a quick review? Any suggestion?

THX in advance.

website_contact_lastname/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "version": "15.0.1.0.0",
+    "category": "Website",
+    "website": "https://github.com/OCA/website",
+    "author": "Sygel Technology," "Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "auth_signup",
+        "website_account_fiscal_position_partner_type",
```
chienandalu: Isn't this too specific for the task?

website_contact_lastname/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+        "partner_firstname",
```
chienandalu: I'd expect this module to be named as a derivative of this :) (maybe website_sale_partner_firstname)

website_contact_lastname/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+    "name": "Website Contact Lastname",
+    "version": "15.0.1.0.0",
+    "category": "Website",
+    "website": "https://github.com/OCA/website",
+    "author": "Sygel Technology," "Odoo Community Association (OCA)",
+    "license": "AGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "auth_signup",
```
chienandalu: You don't need to declare this one. It comes with website already
```suggestion
```

website_contact_lastname/static/description/icon.png
chienandalu, line 1: Better use the OCA's default or use https://ivantodorovich.github.io/odoo-icon/ to quickly make a fancy one ;)

chienandalu (commented):
Some comments in the formal part

website_contact_lastname/__manifest__.py
manuelregidor: Sorry, but I don't think I get what you mean in this point.

website_contact_lastname/__manifest__.py
manuelregidor: done

website_contact_lastname/static/description/icon.png
manuelregidor: done

website_contact_lastname/__manifest__.py
manuelregidor: done

website_contact_lastname/__manifest__.py
chienandalu: That this dependency is too specific and not everyone will be interested in pulling it to enjoy this new feature

ValentinVinagre: then we would have to carry out 3 separate modules. Since we have made the extension for the secondname. I don't see any harm in indicating on the website whether the client is a company or an individual for the correct treatment of the fields (the addition of the dependency).

HaraldPanten: I agree. The module has been designed considering that a customer can be B2B or B2C. Considering this fiscal position type, the dependency is a must

chienandalu (commented):
Oh, it looks like the module already existed!: https://github.com/OCA/e-commerce/pull/804

ValentinVinagre:
> Oh, it looks like the module already existed!: [OCA/e-commerce#804](https://github.com/OCA/e-commerce/pull/804)

We have several modules to upload:
* website_sale_partner_firstname
* website_sale_partner_second_lastname

both are with dependencies to the 'website_account_fiscal_position_partner_type' module because it makes sense to indicate whether certain fields are required or not depending on whether the user is a company or an individual user.
If we were to separate the modules we would have to make 2 more modules to hook the modules together:
* website_sale_partner_firstname
* website_sale_partner_second_lastname

with the 'website_account_fiscal_position_partner_type'.
What do you propose to do?
IMO: I think 4 modules is exaggerated. I would say that it is a good dependence on the type of client.

pedrobaeza:
The problem is that you are using the partner type in fiscal position to determine if individual or company, which is not correct. If core doesn't do that distinction, you should build another base module for that (although I think you can infer it through the input data).

PR #1008, opened 2023-09-04
[15.0][FIX] website_snippet_big_button: default links

pilarvargas-tecnativa:
Default links should not be set so that a user can edit where each button should redirect to so that in case of not editing the link there are no default redirects.

cc @Tecnativa TT41033

@chienandalu @CarlosRoca13 please review :)


chienandalu (approved):
:+1:

pedrobaeza (approved):
/ocabot merge minor

PR #964, opened 2023-03-16
[15.0][IMP] website_crm_privacy_policy: Mark it as deprecated

CarlosRoca13:
cc @Tecnativa TT41895

reference: https://github.com/OCA/OpenUpgrade/pull/3786

ping @chienandalu @pedrobaeza

PR #1014, opened 2023-11-07
[16.0][IMP] website_crm_privacy_policy: correct editor errors + Forward port

pilarvargas-tecnativa:
First of all, this module was deprecated in v15, so a port of the commit forwarding has been made indicating that it is deprecated. In order to avoid any major problems for potential users, it will remain in v16.
Forward port of: https://github.com/OCA/website/pull/964
On the other hand this module had problems with the editor because of some classes defined in the container of the checkbox, which was causing a conflict and did not allow editing correctly, the problem has been solved by removing the conflicting and unnecessary classes to edit both the text and the link, although this functionality is already included in the website_form_require_legal module and hence website_crm_privacy_policy is obsolete.
The error:
Peek 2023-11-07 13-09

cc @Tecnativa TT45834

@pedrobaeza @CarlosRoca13 please review


CarlosRoca13 (approved):

website_crm_privacy_policy/templates/contactus.xml
```diff
@@ -9,9 +9,7 @@
             expr="//form//div/div/span[@id='s_website_form_result']/.."
             position="before"
         >
-            <div
-                class="form-group col-12 s_website_form_field s_website_form_required_custom"
-            >
+            <div class="form-group col-12">
                 <div class="form-check">
```
pedrobaeza, line 13: Is all this stuff needed? Can this be replaced by enabling such check that you mention?

pilarvargas-tecnativa: After analysing it, if the option is enabled in the editor, there is no way to do it specifically only for this form, it would be enabled for all the forms that are inserted from the snippet, in addition we would add the dependency to this module. 
The other option is to recreate the behaviour of the website_form_require_legal module which involves defining new js code to add the terms acceptance checkbox in addition to the dependency on website_form_require_legal.
This modification has the least impact on this module and also does not imply incompatibility with website_form_require_legal as it will eventually be replaced by website_form_require_legal.

chienandalu (commented):
The main issue I'm seeing here is that the inheritance in the middle of the form seems to be spoiling the edition of the contact page. I think there's no much sense to invest in more time in this module which should go away

pedrobaeza (approved):
OK, let's continue, as this one is the less bad solution:

/ocabot merge major

PR #1017, opened 2023-11-12
[17.0][MIG] website_odoo_debranding Migration to 17.0

NICO-SOLUTIONS:
Standard migration to 17.0


website_odoo_debranding/static/description/index.html
```diff
@@ -0,0 +1,441 @@
+<?xml version="1.0" encoding="utf-8"?>
```
tate11, line 1: <?xml version="1.0" encoding="utf-8"?>
It gives an error, there is no need

NICO-SOLUTIONS: the line in the index file is added by the precommit run, that i perform before commiting the migration part. Should i delete this manually? or maybe an issue for the pre commit setup?

NICO-SOLUTIONS: @pedrobaeza 
What do you think concerning the mentioned line and how to get rid of it. Manually or adjusting pre commit? Or do I miss something?

pedrobaeza: It's just a warning, so let's continue with it for now. The README generator should be changed to not generate this line, but that's not done yet. If someone wants to contribute with that improvement (only for 17.0).

NICO-SOLUTIONS: Alright.. thank for having a look and the feedback.

website_odoo_debranding/templates/disable_odoo.xml
```diff
@@ -1,9 +1,8 @@
-<?xml version="1.0" encoding="UTF-8" ?>
 <odoo>
     <!-- Remove "Create a free website with Odoo" -->
     <template
         id="layout_footer_copyright"
-        inherit_id="website.brand_promotion"
+        inherit_id="web.brand_promotion"
```
astirpe: Why did you change this? It seems not necessary to me.

NICO-SOLUTIONS: you are right! not necessary. happens, as this promotion message comes from web. my bad...

website_odoo_debranding/templates/disable_odoo.xml
```diff
@@ -1,4 +1,3 @@
-<?xml version="1.0" encoding="UTF-8" ?>
```
astirpe: It seems to me that this line should be not removed. I'm checking the tech documentation and I don't see any indication for removing this.

NICO-SOLUTIONS: In general XML declarations in descriptions are deprecated since Odoo 17. I think this is related primarly to module descriptions. Shouldn´t this apply to views as well, to have a straight handling of the encoding?

astirpe: Yes, the deprecation it's related to the [index.html](https://github.com/OCA/website/blob/61179fb2b5275e5ab91dd8746b87945d927d4f65/website_odoo_debranding/static/description/index.html) only.
But you don't have to worry to remove that line in the index.html file because the bot will do it automatically.

astirpe: So it would be better to leave the disable_odoo.xml untouched IMO.

NICO-SOLUTIONS: ok i will return this.
The bot will only take care of the index, when the dotfiles are updated on each repo. so not yet...

astirpe (approved):
Tested in runboat and code review ok

AaronHForgeFlow (approved):
Functional

astirpe (changes requested):
There's an error when installing this module and afterwards installing enterprise module `planning`:

`Element '<xpath expr="//t[@t-call=&#39;web.brand_promotion_message&#39;]">' cannot be located in parent view`

NICO-SOLUTIONS:
> There's an error when installing this module and afterwards installing enterprise module `planning`:
> 
> `Element '<xpath expr="//t[@t-call=&#39;web.brand_promotion_message&#39;]">' cannot be located in parent view`

good catch! 
when installing planning, web.brand_promotion will be replaced by planning.brand_promotion. I´ll check xpathing another element...

NICO-SOLUTIONS:
@astirpe 
i think in this case switching back to inherit the web module is the better approach.
As we need to keep...
t t-call="web.brand_promotion_message"
in order to get it working with planning... 

inheriting the web.brand_promotion_message and xpathing the fields seems to be the better choice. 
As this would loose the character of a website module... What do you think?

astirpe:
@NICO-SOLUTIONS 

> inheriting the web.brand_promotion_message and xpathing the fields seems to be the better choice. As this would loose the character of a website module... What do you think?

Yes, you are right: inheriting the `web.brand_promotion_message` would solve the conflicts.
I made a quick try locally with this xpath:
```xml
        <xpath expr="//t[@t-out]" position="replace">
            <t t-out=""/>
        </xpath>
```
but there are still some dirty characters left. So it seems not enough.

website_odoo_debranding/templates/disable_odoo.xml
```diff
@@ -0,0 +1,21 @@
...
+    <!-- Remove "Create a free website with Odoo" -->
+    <template
+        id="layout_footer_copyright"
+        inherit_id="web.brand_promotion_message"
+        customize_show="True"
+        name="Remove Odoo Promotional Link"
+        priority="999"
+    >
+        <xpath expr="//t[@t-set='odoo_logo']" position="replace">
+            <t t-set="odoo_logo" />
```
astirpe: Not sure if it's necessary to replace the t-set

astirpe (approved):
Tested and works fine. Just don't know if the `t-set` needs to be replaced or not but it doesn't matter for the result.

NICO-SOLUTIONS:
> Tested and works fine. Just don't know if the `t-set` needs to be replaced or not but it doesn't matter for the result.

I added this because I had some leftovers when this is not set. But I'll cross check later on

website_odoo_debranding/templates/disable_odoo.xml
NICO-SOLUTIONS: its not necessary... removed xpath on odoo_logo

sanderlienaerts:
Status for getting this one merged?

pedrobaeza:
@sanderlienaerts you can collaborate reviewing the PR: https://odoo-community.org/page/review

But in this case, the existing reviewers are reputed ones,so merging:

/ocabot merge nobump

NICO-SOLUTIONS:
thanks! @pedrobaeza

PR #1018, opened 2023-11-13
[16.0][FIX] website_legal_page: footer xpath

danielduqma:
In some cases, Odoo wraps original span inside another span for translation making view fail. This change allows it to be found eitherwise.

image

Steps to reproduce:
1. Add a new website with one language
2. In this website, add another language
3. From main page of website, switch to the second language
4. Enter translate menu

image

Reproduced on [runboat](http://oca-website-16-0-0115ca85d00b.runboat.odoo-community.org/), website *My Website - 1018*

### DIFF
--- a/website_legal_page/views/website_legal_main_page.xml
+++ b/website_legal_page/views/website_legal_main_page.xml
@@ -8,7 +8,7 @@
         customize_show="True"
     >
         <xpath
-            expr="//footer//div/span[hasclass('o_footer_copyright_name')]"
+            expr="//footer//span[hasclass('o_footer_copyright_name')]"
             position="after"
         >
             <span class="legal_page">-


papulo79:
LGTM!

website_legal_page/views/website_legal_main_page.xml
```diff
@@ -8,7 +8,7 @@
         customize_show="True"
     >
         <xpath
-            expr="//footer//div/span[hasclass('o_footer_copyright_name')]"
+            expr="//footer//div//span[hasclass('o_footer_copyright_name')]"
```
yostashiro: A non-blocking suggestion.
```suggestion
            expr="//footer//span[hasclass('o_footer_copyright_name')]"
```

yostashiro (approved):
LGTM.

website_legal_page/views/website_legal_main_page.xml
danielduqma: I see that well, I tried to reduce diff to ease merge, but the important thing here is an HTML class inside element `footer`. Thanks!

pedrobaeza (approved):
/ocabot merge minor

PR #1024, opened 2023-11-29
[ADD] website_cookieconsent

KKamaa:


website_cookieconsent/views/assets.xml
```diff
@@ -0,0 +1,28 @@
...
+                      href="/website_cookieconsent/static/src/css/cookieconsent/css-components/light-scheme.css"/>
+                <link rel="stylesheet"
+                      href="/website_cookieconsent/static/src/css/cookieconsent/css-components/preferences-modal.css"/>
+                <link rel="stylesheet"
+                      href="/website_cookieconsent/static/src/css/cookieconsent/cookieconsent.css"/>
+                <link rel="stylesheet"
+                      href="/website_cookieconsent/static/src/css/website_cookieconsent.css"/>
+                <link rel="stylesheet" href="/website_cookieconsent/static/src/js/iframemanager/css/iframemanager.css"/>    
+                <script src="/website_cookieconsent/static/src/js/cookieconsent/cookieconsent.js"></script>
+                                <script src="/website_cookieconsent/static/src/js/iframemanager/js/iframemanager.js"></script>
```
thomaspaulb: When including libraries, could you add the unminified ones instead? Odoo has [its own minifier](https://github.com/OCA/OCB/blob/8.0/openerp/addons/base/ir/ir_qweb.py#L1568) and in this way we can unminify by going into debug=assets mode if we want to set some breakpoints or do debugging.

thomaspaulb (changes requested):

website_cookieconsent/controllers/main.py
```diff
@@ -0,0 +1,17 @@
...
+    @http.route("/website_cookieconsent", type='json', auth="public")
+    def get_config_cookie_config(self, **kw):
+        cookie_config = request.env['ir.config_parameter'].sudo().get_param(
+            'website.cookie_consent_config')
+        iframe_config = request.env['ir.config_parameter'].sudo().get_param(
+            'website.iframemanager_config')
+        return {
+            'cookie_config': cookie_config,
+            'iframemanager_config': iframe_config
+            }
```
thomaspaulb: This should be somewhat to the left I think?

website_cookieconsent/static/src/js/website_cookieconsent.js
```diff
@@ -0,0 +1,68 @@
...
+            return false;
+        }
+    }
+
+    $.ajax({
+        type: 'POST',
+        url: '/website_cookieconsent',
+        dataType: 'json',
+        beforeSend: function(xhr) {
+            xhr.setRequestHeader('Content-Type', 'application/json');
```
thomaspaulb: I think this is not needed if you use `this.rpc` as for example in [here](https://github.com/OCA/web/blob/8.0/help_online/static/src/js/help_online.js#L79).

website_cookieconsent/static/src/js/website_cookieconsent.js
```diff
@@ -0,0 +1,68 @@
+$(function() {
+    "use strict";
```
thomaspaulb, line 2: I think instead of using this start, and then call `$.ajax` from the main, the following is more in line with Odoo:

https://github.com/OCA/website/blob/8.0/website_backend_views/static/src/js/website_backend_views.js#L23

thomaspaulb (changes requested):

website_cookieconsent/views/assets.xml
KKamaa: @thomaspaulb I added this for the js libs.

website_cookieconsent/controllers/main.py
KKamaa: fixed!

website_cookieconsent/static/src/js/website_cookieconsent.js
KKamaa: This should be the one for website js in v 8.0

website_cookieconsent/static/src/js/website_cookieconsent.js
KKamaa: I think rpc is when you are on the web side. For website I used [openerp.jsonrpc](https://github.com/OCA/OCB/blob/279529c088d28b8ad58da3314a829b0b9421dff0/addons/website/static/src/js/website.js#L309)

website_cookieconsent/static/src/js/cookieconsent/cookieconsent.js
```diff
@@ -0,0 +1,1143 @@
+/*!
+ * CookieConsent 3.0.0-rc.17
+ * https://github.com/orestbida/cookieconsent/tree/v3.0-beta
+ * Author Orest Bida
+ * Released under the MIT License
+ */
+var e, t;
+e = this, t = function(e) {
```
thomaspaulb: This looks like it's still minified.

website_cookieconsent/data/website_cookieconsent_data.xml
```diff
@@ -0,0 +1,15 @@
...
+    <data noupdate="1">
+        <record id="cookie_consent_config" model="ir.config_parameter">
+            <field name="display_name">Cookie Consent Config</field>
+            <field name="key">website.cookie_consent_config</field>
+            <field name="value">{}</field>
+        </record>
+        <record id="cookie_iframemanager_config" model="ir.config_parameter">
+            <field name="display_name">Cookie Consent Iframemanager Config</field>
+            <field name="key">website.iframemanager_config</field>
+            <field name="value">{}</field>
```
thomaspaulb, line 12: Shouldn't we put some defaults here?

website_cookieconsent/static/src/js/cookieconsent/cookieconsent.js
KKamaa: Hey Tom, I only see minified version of the lib from the repo here => https://github.com/orestbida/cookieconsent/tree/master/dist

website_cookieconsent/data/website_cookieconsent_data.xml
KKamaa: Adding defaults leads to xml syntax errors I saw its better to leave it empty and add usage example to the github repo https://github.com/orestbida/iframemanager#how-to-set-attributes-on-the-iframe-element

thomaspaulb: Okay. XML syntax errors I think can be avoided by using `CDATA`

website_cookieconsent/static/src/js/cookieconsent/cookieconsent.js
```diff
@@ -0,0 +1,7 @@
+/*!
```
thomaspaulb, line 1: Hmm, this still looks minified.

website_cookieconsent/static/src/js/cookieconsent/cookieconsent.js
thomaspaulb: Ah, you're right. Nevermind then.

website_cookieconsent/static/src/js/cookieconsent/cookieconsent.js
thomaspaulb: And nevermind this either.

thomaspaulb (approved):

PR #1025, opened 2023-12-13
[15.0][IMP] website_form_require_legal: Set link to terms and conditions

pilarvargas-tecnativa:
Improve the template by allowing to set the URL when rendering the qweb template by setting a value for the termsURL variable by context.

cc @Tecnativa TT40305

@pedrobaeza @chienandalu please review

### DIFF
--- a/website_form_require_legal/static/src/js/options.js
+++ b/website_form_require_legal/static/src/js/options.js
@@ -35,7 +35,7 @@ odoo.define("website_form_require_legal.form_editor", function (require) {
                 $(template).html(
                     qweb.render("website_form_require_legal.s_website_form_legal", {
                         labelWidth: labelWidth,
-                        termsURL: "terms",
+                        termsURL: "/terms",
                     })
                 );
                 const legal = template.content.firstElementChild;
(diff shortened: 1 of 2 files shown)


website_form_require_legal/static/src/js/options.js
```diff
@@ -35,7 +35,6 @@ odoo.define("website_form_require_legal.form_editor", function (require) {
                 $(template).html(
                     qweb.render("website_form_require_legal.s_website_form_legal", {
                         labelWidth: labelWidth,
-                        termsURL: "terms",
```
pedrobaeza, line 38: Why removing this default value?

pedrobaeza: I confirm that this is working with a sane default `/terms`, so please restore it with that value.

pilarvargas-tecnativa: done!

pedrobaeza (approved):

CarlosRoca13 (approved):

chienandalu (approved):

PR #1028, opened 2024-01-09
[MIG] website_local_font: Migration to 16.0

Reyes4711-S73:
Standard migration to 16.0
Reopen from https://github.com/OCA/website/pull/999


website_local_font/__manifest__.py
```diff
@@ -25,5 +25,8 @@
                 "website_local_font/static/src/scss/secondary_variables.scss",
             ),
         ],
+        "web.assets_backend": [
+            "website_local_font/static/src/xml/website.editor.xml",
```
ByteMeAsap, line 29: Can you add this xml file to website.assets_wysiwyg asset instead of website_assets_backend? That way you don't need to define this in the snippets.options.js as xmldependency

website_local_font/static/src/js/snippets.options.js
ByteMeAsap, line 1: [snippets.options.js.txt](https://github.com/OCA/website/files/14405362/snippets.options.js.txt)
Modify the code as shared in the attached file. I have removed xmlDependencies, inherited the start method in a way that it can be inherited by other modules in future, removed widgetObj in _saveLocalFont function and replaced by this and made use of "const" instead of "var"

ByteMeAsap (changes requested):

Reyes4711-S73:
@ByteMeAsap I put your changes but, I tested in a local environment and in a runboat and I have this error.
image
When you modify the snippets.options.js, there are some errors in pre-commit. Please review it. I'm going to return to previous changes

PR #1030, opened 2024-01-15
[14.0][ADD] website_recaptcha_v2

huguesdk:
this module allows to use recaptcha v2 on website forms.

it is a helper module that only provides the widget and the validation logic which can be used by other modules to actually display it on website forms and check whether the user entry is valid.

it is derived from `website_recaptcha` (for odoo 8) from elico corp, `website_recaptcha_reloaded` (for odoo 9) from tech receptives, with also some code coming from `website_form_recaptcha` (for odoo 12) from the oca.

the standard module `google_recaptcha` that appeared in odoo 14 only allows to use recaptcha v3, which uses a different (non-interactive) validation mechanism. both versions are supported by google as they consider that they serve different purposes.

this is a follow-up to #928, as explained in https://github.com/OCA/website/pull/928#issuecomment-1677290413.


huguesdk:
note that pre-commit is failing only because #1031 is not merged yet.

robinkeunen (approved):
LGTM, same comments as #1032 

precommit will be fixed with an update of copier on branch 14.0

PR #1031, opened 2024-01-15
[14.0] update repository template to 1.20

huguesdk:
this is to make pre-commit pass on new pull requests.

note that i disabled an error in some files to make it pass, as i don't know the modules enough to fix the error properly.


huguesdk:
superseded by #1037.

PR #1032, opened 2024-01-16
[16.0][ADD] website_recaptcha_v2

huguesdk:
this module allows to use recaptcha v2 on website forms.

it is a helper module that only provides the widget and the validation logic which can be used by other modules to actually display it on website forms and check whether the user entry is valid.

it is derived from `website_recaptcha` (for odoo 8) from elico corp, `website_recaptcha_reloaded` (for odoo 9) from tech receptives, with also some code coming from `website_form_recaptcha` (for odoo 12) from the oca.

the standard module `google_recaptcha` that appeared in odoo 14 only allows to use recaptcha v3, which uses a different (non-interactive) validation mechanism. both versions are supported by google as they consider that they serve different purposes.

this is a follow-up to #928, as explained in https://github.com/OCA/website/pull/928#issuecomment-1677290413.

this is the 16.0 version of #1030.


website_recaptcha_v2/models/website.py
```diff
@@ -0,0 +1,66 @@
...
+    def _get_error_message(self, errorcode=None):
+        mapping = {
+            "missing-input-secret": _("The secret parameter is missing."),
+            "invalid-input-secret": _("The secret parameter is invalid or malformed."),
+            "missing-input-response": _("The response parameter is missing."),
+            "invalid-input-response": _(
+                "The response parameter is invalid or malformed."
+            ),
+        }
+        return mapping.get(errorcode, _("There was a problem with the captcha entry."))
```
robinkeunen: Would it be helpful to log the errorcode if it is not in the mapping above ?

website_recaptcha_v2/models/website.py
```diff
@@ -0,0 +1,66 @@
...
+# SPDX-FileCopyrightText: 2019 Simone Orsi - Camptocamp SA
+# SPDX-FileCopyrightText: 2019 Coop IT Easy SC
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
+import requests
+
+from odoo import _, api, fields, models
+
+URL = "https://www.recaptcha.net/recaptcha/api/siteverify"
```
robinkeunen: Would this be a better naming ?

```suggestion
RECAPTCHA_API_URL = "https://www.recaptcha.net/recaptcha/api/siteverify"
```

website_recaptcha_v2/models/website.py
```diff
@@ -0,0 +1,66 @@
...
+
+        res = requests.post(URL, data=get_res, timeout=RECAPTCHA_API_TIMEOUT).json()
+
+        error_msg = "\n".join(
+            self._get_error_message(error) for error in res.get("error-codes", [])
+        )
+        if error_msg:
+            return (False, error_msg)
+
+        if not res.get("success"):
```
robinkeunen, line 75: In this case, you get no "error-codes" nor "success" in the response : would another error message than "There was a problem with the captcha entry" be more helpful ?

robinkeunen (approved):
LGTM, thx for the comprehensive unit tests.

I made a few non-blocking suggestions.

remytms (approved):
LGTM.

remytms:
Can a PSC merge this PR? I think that it's ready to merge. Thanks :)

edescalona (commented):
LGTM

huguesdk:
@robinkeunen thank you for the insightful review! i finally added your suggestions to the code (but the changes don’t show up in this pull request, probably because it is currently closed).

@edescalona thank you for the review.

huguesdk:
@OCA/website-maintainers could someone please reopen this and review or merge this?

jelenapoblet:
Great team work @edescalona @robinkeunen @huguesdk @pedrobaeza @remytms !

huguesdk:
@pedrobaeza many thanks!

PR #1039, opened 2024-03-21
[17.0][MIG] website_require_login: Migration to 17.0

vvro:
website_require_login module works as is from version 16.0, just updated the version to 17.


pedrobaeza:
Now formally seems OK. Thanks.

/ocabot migration website_require_login

pedrobaeza:
Ouch, there's another PR already.

vvro:
It seems already fixed, but pending for almost half year. How should we proceed in this case? @pedrobaeza

pedrobaeza:
You should review it https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#L1345

PR #1042, opened 2024-05-08
[16.0][ADD] website_geoip_language

yibudak:
While experimenting with Odoo's GeoIP capabilities, I noticed that the website did not have automatic language selection based on the customer's IP address. I searched for modules that did this, but I could not find an example and decided to make this module.

Technically, I monkey-patching the “best_lang” property in the odoo.http.Request class so that the client's IP address takes priority in language selection. Maybe there is a better way, but this is how I was able to make the module without breaking the overall functioning of the system.

There is no need to make any additional configuration for the module, it works if the module is installed and the GeoIP database is defined in Odoo. Maybe I need to explain this in a better way in the “Usage” section, I am waiting for your help.


yibudak:
@pedrobaeza @remytms @robinkeunen @chienandalu

pedrobaeza:
Interesting addition. Maybe you should use Odoo patch method instead of direct Python monkey-patching. Apart from that, it seems good.

yibudak:
> Interesting addition. Maybe you should use Odoo patch method instead of direct Python monkey-patching. Apart from that, it seems good.

Thanks for your attention Pedro. I don't know what you mean by using Odoo's patch method, is there such a method for Python like there is for Javascript?

pedrobaeza:
I was talking about https://github.com/odoo/odoo/blob/819705cb9418854d9c85248214ce54d5528f7ef2/odoo/models.py#L5063

PR #1049, opened 2024-07-30
[16.0][MIG] website_menu_by_user_status: Migration to 16.0

alejandro-vallejoFL:
This PR has two objectives: to make a standard migration of the website_menu_by_user_status module to v16.0 and fix a bug related to the visibility of website menus:

The _compute_visible method of the core calculates the is_visible field based on certain conditions, such as that the linked page is published.
This method calls the super of this method, and if after doing so the menu entry is not visible, it is saved as not visible, but then it is made visible if it is visible to logged in or non-logged in users.

Let's see an example:

This page is not published:

PageNotPublished

We added a menu that links to the previous page. By default, the visible checks are marked for logged in and non-logged in users:

ConfigVisible

With these settings, the menu is displayed and directed to the page, giving an error as it is not published:

Error

This PR  is responsible for fixing that error.

FL-556-1335


danielduqma (approved):

danielduqma:
cc @OCA/website-maintainers

tarteo (approved):

tarteo:
LGTM! Thanks (tested locally)

PR #1052, opened 2024-08-06
[FIX] website_google_tag_manager: fix hidden menu behavior with tag manager

rolandojduartem:
Adding script before the wrapwrap element affects style [1] and hide the nav bar behind of the main menu, so it is required adding it after the wrapwrap element, the same way as [2] does it.

Reference:

- [1] https://github.com/odoo/odoo/blob/32a07bf4/addons/website/static/src/scss/website.scss#L200
- [2] https://github.com/odoo/odoo/blob/32a07bf4/addons/website/views/website_templates.xml#L151

**Before:**

Screenshot from 2024-08-06 08-03-59

**After:**

Screenshot from 2024-08-06 08-04-22

PR #1053, opened 2024-08-06
[FIX] website_google_tag_manager: fix hidden menu behavior with tag manager

rolandojduartem:
Adding script before the wrapwrap element affects style [1] and hide the nav bar behind of the main menu, so it is required adding it after the wrapwrap element, the same way as [2] does it.

Reference:

- [1] https://github.com/odoo/odoo/blob/32a07bf4/addons/website/static/src/scss/website.scss#L200
- [2] https://github.com/odoo/odoo/blob/32a07bf4/addons/website/views/website_templates.xml#L151

This is a forward-port from https://github.com/OCA/website/pull/1052

### DIFF
--- a/website_google_tag_manager/views/website_templates.xml
+++ b/website_google_tag_manager/views/website_templates.xml
@@ -2,7 +2,7 @@
      License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl). -->
 <odoo>
     <template id="layout" inherit_id="website.layout">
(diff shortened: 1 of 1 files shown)


rolandojduartem:
Hi, @luisg123v, @hugho-ad and @imanie383, please, could you review this PR? T#76185

luisg123v (approved):
@moylop260

Could you merge, please?

This is a forward-port.

PR #1055, opened 2024-08-14
[16.0][FIX] website_cookiebot: disable for sure

chienandalu:
If we don't remove the key the assets will keep trying to load, which isn't what we'd expect.

cc @Tecnativa TT50520

please review @pedrobaeza @CarlosRoca13

### DIFF
--- a/website_cookiebot/wizards/res_config_settings.py
+++ b/website_cookiebot/wizards/res_config_settings.py
@@ -15,14 +15,9 @@ class ResConfigSettings(models.TransientModel):
     )
     cookiebot_enabled = fields.Boolean(
         string="Cookiebot",
-        compute="_compute_cookiebot_enabled",
+        related="website_id.cookiebot_enabled",
         readonly=False,
     )
-
-    @api.depends("website_id.cookiebot_dgid")
-    def _compute_cookiebot_enabled(self):
-        for record in self:
-            record.cookiebot_enabled = bool(record.cookiebot_dgid)
 
     @api.onchange("website_cookies_bar")
     def _onchange_website_cookies_bar(self):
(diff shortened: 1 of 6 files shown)


pedrobaeza:
Erasing the identifier seems a bit aggressive. Can we convert the boolean into a regular field and change the template to not load it if not checked?

chienandalu:
We could do it but that will require more changes and a migration script

pedrobaeza:
Well, I think that the way to go. If not, you lose the ID if you want to disable it temporarily or make some tests.

chienandalu:
Ok, done :)

website_cookiebot/wizards/res_config_settings.py
```diff
@@ -15,6 +15,7 @@ class ResConfigSettings(models.TransientModel):
     )
     cookiebot_enabled = fields.Boolean(
         string="Cookiebot",
+        related="website_id.cookiebot_enabled",
         compute="_compute_cookiebot_enabled",
```
pedrobaeza: Remove this attribute

chienandalu: It's used as an onchange to set it off when the Odoo's own cookies setting is on

pedrobaeza: Then just remove the related and put an inverse for clearer understanding.

chienandalu: Sorry, I missread it. It was useless...

CarlosRoca13 (approved):

pedrobaeza (approved):
/ocabot merge nobump

PR #1056, opened 2024-08-26
Portal Email Readonly

kumarinie:


kumarinie:
This module is an addition to Odoo's portal module, and makes sure that the email field on the 'My Account' page in the portal / website / webshop is not editable by the portal user.
This can be very useful when making use of an extended authentication tooling that uses the email address as the user id, and you do not want this to be changed.

PR #1071, opened 2024-11-28
[17.0][MIG] website search header

dnplkndll:
@kobros-tech


dnplkndll:
@kobros-tech should we review and get this one merged to 17? not sure is worth the trouble or not.

PR #1073, opened 2024-11-30
[ADD] no_modal_search_header: search in page view not in modal view

kobros-tech:
Screenshot from 2024-11-30 23-03-15

Screenshot from 2024-12-01 01-23-45


kobros-tech:
@dnplkndll

dnplkndll:
@kobros-tech do we still want to use this on the CMS sites like FC?

kobros-tech:
we can keep it and migrate if we need, no need for merging.

PR #1092, opened 2025-03-07
[16.0][FIX] website_whatsapp: messages shown should be translatable.

JordiBForgeFlow:
When a user starts the chat a message is proposed. This message should be translatable, so that it will appear in the language used by the user in the website.

### DIFF
--- a/website_whatsapp/models/res_config_settings.py
+++ b/website_whatsapp/models/res_config_settings.py
@@ -24,7 +24,6 @@ class ResConfigSettings(models.TransientModel):
     whatsapp_text = fields.Char(
         related="website_id.whatsapp_text",
         readonly=False,
-        translate=True,
     )
     whatsapp_track_url = fields.Boolean(
         related="website_id.whatsapp_track_url",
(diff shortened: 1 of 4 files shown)


ioans73 (approved):
LGTM

JordiBForgeFlow:
@pedrobaeza Can you also merge this one plz?

website_whatsapp/models/res_config_settings.py
```diff
@@ -24,6 +24,7 @@ def _inverse_whatsapp_enabled(self):
     whatsapp_text = fields.Char(
         related="website_id.whatsapp_text",
         readonly=False,
+        translate=True,
```
pedrobaeza: I think this is not needed being a related.

JordiBForgeFlow: Fixed!

JordiBForgeFlow:
@pedrobaeza @ioans73  I have changed a bit the style so that odoo's livechat and this whatsapp button can co-exist and follow the same style.

After the fix:
image

pedrobaeza (approved):
@ioans73 do you agree with the style change?

ioans73:
@pedrobaeza It's ok for me! Thanks

PR #1097, opened 2025-03-18
[16.0][IMP] website_whatsapp: restrict the visibility of the whatsapp button by country

JordiBForgeFlow:
Introduces a whitelisted countries and blacklisted countries, so that the icon can optionally appear only in selected countries.

image


JordiBForgeFlow:
In order to test this it's necessary to set up your server to use the geo_ip https://www.odoo.com/documentation/16.0/administration/on_premise/geo_ip.html

PR #1099, opened 2025-03-25
[16.0][FIX] website_require_login: Error response

Alexgars73:
Error in response in first load:
image

### DIFF
--- a/website_require_login/models/ir_http.py
+++ b/website_require_login/models/ir_http.py
@@ -44,5 +44,7 @@ class IrHttp(models.AbstractModel):
         path = request.httprequest.path
         for auth_path in auth_paths:
             if auth_path == path or Path(auth_path) in Path(path).parents:
+                if request.httprequest.content_type == "application/json":
+                    return False
                 redirect_path = "/web/login?redirect=%s" % path
                 return request.redirect(redirect_path, code=302)
--- a/website_require_login/tests/test_ir_http.py
+++ b/website_require_login/tests/test_ir_http.py
@@ -1,3 +1,5 @@
+import json
+
 from odoo.tests import HttpCase
 
 
@@ -34,4 +36,19 @@ class TestIrHttp(HttpCase):
             response.status_code,
             200,
             "Expected the response status code to be 200 which means no redirection",
+        )
+
+    def test_dispatch_json_no_redirect(self):
+        """With Content-Type application/json there should be no redirection."""
+        self.authenticate(None, None)
+        response = self.url_open(
+            self.path,
+            data=json.dumps({}),
+            headers={"Content-Type": "application/json"},
+            allow_redirects=False,
+        )
+        self.assertNotEqual(
+            response.status_code,
+            302,
+            "JSON request should not redirect even if user is not logged in",
         )


Alexgars73:
@pedrobaeza could you remove the stale tag please?

pedrobaeza:
OK, but please enroll people to review it.

Alexgars73:
@yvaucher @NICO-SOLUTIONS could you review please?

NICO-SOLUTIONS (approved):
Havent seen this effect up to now but it makes sense!
in generel i would place the check for Content-Type == "application/json" before assigning the redirect_path. but thats academic.

LGTM!

website_require_login/tests/test_ir_http.py
```diff
@@ -35,3 +37,18 @@ def test_dispatch_authorized(self):
             200,
             "Expected the response status code to be 200 which means no redirection",
         )
+
+    def test_dispatch_json_no_redirect(self):
+        """Con Content-Type application/json no debe haber redirección."""
```
tarteo: Please use English

tarteo (changes requested):

website_require_login/tests/test_ir_http.py
Alexgars73: @tarteo done!

website_require_login/models/ir_http.py
```diff
@@ -44,5 +44,7 @@ def _check_require_auth(cls):
         path = request.httprequest.path
         for auth_path in auth_paths:
             if auth_path == path or Path(auth_path) in Path(path).parents:
+                if request.httprequest.content_type == "application/json":
```
SirPyTech, line 47: **suggestion**: I'd place this close to the checks at the beginning of the method so that if this fails then we can skip everything else, like the `search` method.

website_require_login/tests/test_ir_http.py
```diff
@@ -35,3 +37,18 @@ def test_dispatch_authorized(self):
             200,
             "Expected the response status code to be 200 which means no redirection",
         )
+
+    def test_dispatch_json_no_redirect(self):
```
SirPyTech, line 41: **praise**: Thanks for adding the test!

SirPyTech (approved):
Thanks for the PR!
I only reviewed the code: I just suggested an optimization.

Please have a look at https://github.com/OCA/website/pull/1160 that is quite similar but for backend requests 😁

website_require_login/tests/test_ir_http.py
francesco-ooops: @tarteo could you check again? thanks!

SirPyTech:
@tarteo please update your review :pray:

tarteo (approved):
Sorry for late response

website_require_login/models/ir_http.py
tarteo: Yes I think you're right, do want this change? Or can I go ahead with merging?

tarteo: I continued with merging because this PR is open for a long time already, and you approved the changes, so I assume you don't need the change you suggested

SirPyTech: That's right, I approved because this is not mandatory for merge in my opinion, thanks for merging!

PR #1100, opened 2025-04-07
[18.0][MIG] website_form_require_legal: Migration to 18.0

Alexgars73:
Standard migration to 18.0


Reyes4711-S73:
LGTM

Reyes4711-S73 (approved):

pilarvargas-tecnativa:
Please indicate the version in the PR title [18.0] and reduce the administrative commits.

pilarvargas-tecnativa:
The selector may need to be adjusted, it should be aligned to the left.
image

Alexgars73:
@pilarvargas-tecnativa Hello, I have tried to replicate what you told me in a new database but it works fine.
image

pilarvargas-tecnativa:
@Alexgars73  Can you push again to make runbot available for testing?

Alexgars73:
> @Alexgars73 Can you push again to make runbot available for testing?

Done

pilarvargas-tecnativa:
My first test had been directly on the contact form. Testing by inserting the form block replicates this behaviour when changing the position of the label but since it is intended for this block and is editable it is not necessary to change this behaviour, it just needs to be edited.
image
image

On the other hand, there must be some kind of field that when selecting the terms acceptance line to edit it, it detects it as a field and shows the editing tools of a field and it should not as it causes an error:

https://github.com/user-attachments/assets/aa3f3272-beeb-451b-84f6-398f5788810b

These tools should not be visible for the acceptance of terms.

image

You can take as reference id="website_sale_tc_checkbox" from website_sale

Alexgars73:
@pilarvargas-tecnativa Done.

pilarvargas-tecnativa:
I tried again in runbot, although the input is required, I let you submit the form without having accepted the terms.

image

Alexgars73:
@pilarvargas-tecnativa I don't think we're in the same situation as "website_sale_tc_checkbox." When I remove the s_website_form_field class, it doesn't interpret it as a field, but it doesn't perform the check. The other way around does. This problem has been going on since v15. Any ideas on how to fix it? Thanks.

pilarvargas-tecnativa:
It is clear that the field must be required before the form can be submitted, as that is how it works. Having a class that causes errors in the editor is not ideal and should also be avoided in previous versions. Perhaps the solution is to follow Odoo's approach, as it does at checkout. Until the legal terms have been checked, the button should remain inactive.

Alexgars73:
@pilarvargas-tecnativa i made some changes please if you can check it. Thanks

pilarvargas-tecnativa:
cc @Tecnativa TT54378

@CarlosRoca13 please review

pilarvargas-tecnativa (approved):
thanks!

CarlosRoca13 (approved):
Code review :+1:

carlos-lopez-tecnativa (approved):
LGTM

PR #1107, opened 2025-06-02
[18.0][MIG] website_analytics_matomo: Migration to 18.0

ByteMeAsap:


legalsylvain:
hi @ByteMeAsap. could you reopen ? 

I use this module in production, and it works like a charm.

@victor-champonnois : could you review this PR, as you are the original author ?

thanks !

ByteMeAsap:
hey @legalsylvain , can you please guide me on how to reopen this?

legalsylvain:
@OCA/website-maintainers could you reopen ?

hbrunn:
@legalsylvain if this works like a charm for you, you should review it

https://github.com/OCA/website/pull/1133 seems interesting to review+integrate too

legalsylvain:
> @legalsylvain if this works like a charm for you, you should review it

It is not possible, when it's closed. but now, it's possible.

legalsylvain (approved):
LGTM. Used in production. Thanks for porting this tool.

One thing : 

- this module implement the "integrate js file" way. As a result, we have full information in matomo, but such implementation is blocked by tools that preserves user privacy (Like uBlock Origin). As a result, all people using such tool will be ignored, in matomo, and user count will be wrong. That's a pity.

- Another way to implement is to use the "matomo API" way. (here Odoo calls directly matomo api). As a result, user count is correct. (but there will be less information in matomo like no duration of user in each page, etc...)

What do you think ?
Could you just add a ROADMAP.md file to mention the interest to implement the second way ? 

Thanks !

hbrunn (approved):
while I agree that @legalsylvain's comments would make this much more accurate, for a migration, this is good enough

PR #1110, opened 2025-06-26
[IMP] website_cookiefirst: Replace deprecated Cookiefirst functionality

NICO-SOLUTIONS:
- replace banner.js with new consent.js script using domain and identifier

### DIFF
--- a/website_cookiefirst/README.rst
+++ b/website_cookiefirst/README.rst
@@ -44,9 +44,10 @@ To configure this module, you need to:
 
 1. Go to **Website > Configuration > Settings**
 2. Search 'Cookiefirst' option.
-3. Fill in your 'Cookiefirst ID' (e.g.
-   '00000000-0000-0000-0000-000000000000').
-4. Click on "Save" button.
+3. Enable Use Cookiefirst checkbox
+4. Enter your Cookiefirst **API Key** into **Cookiefirst ID** field
+   (e.g. '00000000-0000-0000-0000-000000000000').
+5. Click on "Save" button.
 
 ⚠️ **Please note: if another cookie consent solution is installed (e.g.
 Cookiebot), the execution of the Cookiefirst script will be prevented.
--- a/website_cookiefirst/readme/CONFIGURE.md
+++ b/website_cookiefirst/readme/CONFIGURE.md
@@ -2,9 +2,10 @@ To configure this module, you need to:
 
 1.  Go to **Website \> Configuration \> Settings**
 2.  Search 'Cookiefirst' option.
-3.  Fill in your 'Cookiefirst ID' (e.g.
+3.  Enable Use Cookiefirst checkbox
+4.  Enter your Cookiefirst **API Key** into **Cookiefirst ID** field (e.g.
     '00000000-0000-0000-0000-000000000000').
-4.  Click on "Save" button.
+5.  Click on "Save" button.
 
 ⚠️ **Please note: if another cookie consent solution is
 installed (e.g. Cookiebot), the execution of the Cookiefirst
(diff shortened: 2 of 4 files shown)


NICO-SOLUTIONS:
@ioans73 

Would you mind reviewing the changes at your convenience?

ioans73:
@NICO-SOLUTIONS The old way is no longer valid?

NICO-SOLUTIONS:
@ioans73 
It will be phased out on july 24th

ioans73:
@NICO-SOLUTIONS  Oh, I didn't know that.
Can you give me the documentation or the blog where you read it?
Thanks.

ioans73 (approved):

ioans73:
@NICO-SOLUTIONS In the Cookiefirst settings (https://app.cookiefirst.com), do I need to change anything or can I just leave it with the current settings?

NICO-SOLUTIONS:
> @NICO-SOLUTIONS Oh, I didn't know that. Can you give me the documentation or the blog where you read it? Thanks.

@ioans73 
This is the relevant port of the mail coming from cookiefirst support...

cookiefirst

NICO-SOLUTIONS:
> @NICO-SOLUTIONS In the Cookiefirst settings (https://app.cookiefirst.com), do I need to change anything or can I just leave it with the current settings?

@ioans73 
i only tested it with a new database and new domain on cookiefirst.
i guess you can leave the settings as it is. If i understood correclty, it will just regenerate a new api key when you change settings in the cookiefirst portal.
It should be sufficient to add only the domain in res config settings defined in cookiefirst portal after updating the module. 
Just to make sure... i would go for a test upfront.

website_cookiefirst/models/website.py
```diff
@@ -11,3 +11,7 @@ class Website(models.Model):
         string="Cookiefirst ID",
         help="This field holds the ID, needed for Cookiefirst functionality.",
     )
+    cookiefirst_domain = fields.Char(
```
ioans73: @NICO-SOLUTIONS Could we set the website domain as the default value?

ioans73: @NICO-SOLUTIONS Or directly use the default Odoo domain?

NICO-SOLUTIONS: @ioans73 
i thought about this as well.. but decided to add a separate field as no http or https is allowed here. What do you think... better to fetch it from settings or let the user enter the domain when activating cookiefirst.

ioans73: I think it would be better to take it from the settings

NICO-SOLUTIONS: i will have a look to it later... should be done pretty easy.

NICO-SOLUTIONS: @ioans73 

done

ioans73 (approved):

NICO-SOLUTIONS:
@eugenios73 
@matteoopenf 
hey. sorry for bothering. 
But are you willing to review the changes ass well? Want to forward and back port it, before the old way is phased out.

matteoopenf:
> @eugenios73 @matteoopenf hey. sorry for bothering. But are you willing to review the changes ass well? Want to forward and back port it, before the old way is phased out.

I check, thank you for the tag

matteoopenf (approved):
LGTM

NICO-SOLUTIONS:
@pedrobaeza 
Thanks 🙏

ioans73:
@NICO-SOLUTIONS With these new changes, does the cookie banner work in your production environment?

ioans73:
@NICO-SOLUTIONS I've found the problem! When we insert the domain into the script, it shouldn't have www.
Can we use "replace" like we do for "http"? I can't modify the domain in the configuration, and this could affect more users.

NICO-SOLUTIONS:
> @NICO-SOLUTIONS I've found the problem! When we insert the domain into the script, it shouldn't have www.
> 
> Can we use "replace" like we do for "http"? I can't modify the domain in the configuration, and this could affect more users.

@ioans73 
hey. Yeah sure we can adapt it. You are right... we should cover that too.

PR #1111, opened 2025-06-27
[16.0][IMP] Add modularity for other reCAPTCHA v2-compatible APIs

Honeyxilia:
# Context

In Odoo v12.0, the OCA module `website_form_recaptcha` allowed for the use of CAPTCHA APIs that were compatible with reCAPTCHA v2 by changing some variables.
However, with `website_form_recaptcha` being fused into `website_form` (and itself into `website`), there are no reCAPTCHA v2 compatibility tool for various APIs.

So, we bring forward this development, to allow for customisation of the used CAPTCHA API for the widget introduced by  this module.

# Additions 

- We add 4 new Char fields to the `website` model, to offer customization of the used API : 
  - `recaptcha_v2_api_url` : the link to the JS CAPTCHA library file (default is https://www.recaptcha.net/recaptcha/api.js )
  - `recaptcha_v2_verify_url` :  the link to the verification endpoint of the API (default is https://www.recaptcha.net/recaptcha/api/siteverify )
  - `recaptcha_v2_html_class` : the CSS class to indicate in which element the widget is to be inserted (default is `g-recaptcha`)
  - `recaptcha_v2_resp_attr` : the value of the form field for hCaptcha validation (default is `g-recaptcha-response`)
  -> These fields are inserted into the verification code of the `website` model, as well as in the reCAPTCHA template.

- We added editable fields in the Settings section to allow them to be easily changed by frontend users.

# Other points

- As we were coding this addition, we noticed that, despite the stated goal for this module to "used by other modules to actually display it on website forms", the widget is actually not available in the Website Page Editor menu, in the Snippets Blocks - is it voluntary ? If not, would it be interesting to add the widget as a Snippet block in the Website Page Editor menu ?

- While the module offers the widget, its initialization and a method to verify the CAPTCHA, there are no controllers in the module for the `/website/form/` endpoint to actually verify it - would it be worth considering adding controllers in this module in order to offer reCAPTCHA verification for the `/website/form/` endpoint ?

### DIFF
--- a/website_recaptcha_v2/models/res_config_settings.py
+++ b/website_recaptcha_v2/models/res_config_settings.py
@@ -21,3 +21,15 @@ class ResConfigSettings(models.TransientModel):
     recaptcha_v2_secret_key = fields.Char(
         related="website_id.recaptcha_v2_secret_key", readonly=False
     )
+    recaptcha_v2_html_class = fields.Char(
+        related="website_id.recaptcha_v2_html_class", readonly=False
+    )
+    recaptcha_v2_api_url = fields.Char(
+        related="website_id.recaptcha_v2_api_url", readonly=False
+    )
+    recaptcha_v2_verify_url = fields.Char(
+        related="website_id.recaptcha_v2_verify_url", readonly=False
+    )
+    recaptcha_v2_resp_attr = fields.Char(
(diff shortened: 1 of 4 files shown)


fcayre (approved):
Code review + real life tests OK (with hCaptcha).

PR #1112, opened 2025-07-01
[18.0][MIG] website_legal_page: Migration to 18.0

victoralmau:
Migration to 18.0

Please @CarlosRoca13 and @david-banon-tecnativa can you review it?

@Tecnativa TT54870


victoralmau:
This module is no longer needed since **v15** because [account](https://github.com/odoo/odoo/blob/3a28e5b0adbb36bdb1155a6854cdfbe4e7f9b187/addons/account/controllers/terms.py#L18) already includes /terms, instead, the module is already in v16. I think it would be appropriate to add in OpenUpgrade 15.0 in apriori this module in `merged_modules`, what do you think @pedrobaeza  ?

CarlosRoca13:
The module is still necessary because not everyone has or wants to have the `account` module installed. This is why the page for this module was set to `/legal`, making it complementary to `/terms`.

david-banon-tecnativa (approved):

pedrobaeza (approved):
/ocabot migration website_legal_page
/ocabot merge nobump

PR #1117, opened 2025-07-10
[17.0][FIX] website_cookiefirst: remove www. prefix from domain in cookiefirst script URL

NICO-SOLUTIONS:
fix for domains with "www." subdomain as discussed here...https://github.com/OCA/website/pull/1110

### DIFF
--- a/website_cookiefirst/views/website_template.xml
+++ b/website_cookiefirst/views/website_template.xml
@@ -19,7 +19,7 @@
         <xpath expr="//script[last()]" position="after">
             <t t-if="website.cookiefirst_identifier and website.domain">
                 <script
-                    t-attf-src="https://consent.cookiefirst.com/sites/{{ (website.domain or '').replace('http://', '').replace('https://', '') }}-{{ website.cookiefirst_identifier }}/consent.js"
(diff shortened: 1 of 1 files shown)


ioans73 (approved):
LGTM

eugenios73 (approved):
LGTM

NICO-SOLUTIONS:
@pedrobaeza 
so can we merge this one? or is there something that needs to be adjusted here as well?

pedrobaeza:
Why cookiefirst doesn't have an URL based only on the identifier, and not in the domain?

/ocabot merge patch

NICO-SOLUTIONS:
> Why cookiefirst doesn't have an URL based only on the identifier, and not in the domain?
> 
> /ocabot merge patch

@pedrobaeza 
well, i gueess this is part of CookieFirst's internal design, that has changed. 
They stated, that this is to ensure future updates etc... 

Thanks a lot for your review, the hints and for merging!

NICO-SOLUTIONS:
@ioans73 
just to let you know. I noticed a few days ago, that only full classified cookies are listed in the declaration. In fact, after a scan only the cookiefirst cookies are listed.
I was in contact with the support the last days and now the issue should be solved, also for free subscription domais. All scanned cookies shoudl be displayed now, even if they are not fully classified (a missing "provider" e.g. means, not fully classified).
I Just wanted to let you know about this and maybe just double check the one or other domain to confirm this.

ioans73:
@NICO-SOLUTIONS Thanks for your message.
I've checked some cases and it seems to work as you say.

PR #1138, opened 2025-10-21
[14.0][ADD] website_user_login_redirect_custom: redirect website users to custom URL

Mohajiro:
Allows configuration of a relative URL where website/portal users will be redirected after login or signup.
Useful for cases when you want new users to land directly on `/shop` or a “thank you” page instead of `/my/home`.

Task: 5023


Marco-ooops (approved):
test funzionale ok

website_user_login_redirect_custom/controllers/main.py
```diff
@@ -0,0 +1,66 @@
...
+        """Override standard login route to apply a custom redirect after login."""
+        response = super().web_login(redirect=redirect, **kw)
+
+        # Only redirect if:
+        # 1. User is logged in and is portal/website user
+        # 2. Feature is enabled
+        # 3. We have a valid redirect response
+        # 4. Custom URL is valid and safe
+        if (
+            self._should_redirect_custom()
```
ivs-cetmix: I think we need to redirect custom only if current redirect is done to the "/my" location. Otherwise this may affect custom redirect scenarios, eg when user logs in during the checkout process.

ivs-cetmix (changes requested):

website_user_login_redirect_custom/controllers/main.py
Mohajiro: I see, good idea. I hope my implementation is as good as your proposal.

website_user_login_redirect_custom/controllers/main.py
```diff
@@ -0,0 +1,68 @@
...
+        # 1. User is logged in and is portal/website user
+        # 2. Feature is enabled
+        # 3. We have a valid redirect response
+        # 4. Redirect target is /my (default after login)
+        # 5. Custom URL is valid and safe
+        if (
+            self._should_redirect_custom()
+            and hasattr(response, "status_code")
+            and response.status_code in (303, 302)
+            and (redirect in (None, "/my", "/my/"))
```
ivs-cetmix: `not redirect or "/my" in redirect"`

ivs-cetmix (changes requested):

ivs-cetmix (approved):
LGTM

website_user_login_redirect_custom/models/res_config_settings.py
```diff
@@ -0,0 +1,84 @@
...
+        res["website_login_redirect_enabled"] = enabled_param == "1"
+        return res
+
+    def set_values(self):
+        """Save boolean value to ir.config_parameter."""
+        super().set_values()
+        enabled_value = "1" if self.website_login_redirect_enabled else "0"
+        self.env["ir.config_parameter"].sudo().set_param(
+            "website_user_login_redirect_custom.enabled", enabled_value
+        )
```
SirPyTech: **question**: Why are we getting/setting the field `website_login_redirect_enabled` with a configuration parameter if we could simply use the `config_parameter` field attribute?

You also have already used the `config_parameter` field attribute for `website_login_redirect_url`, so maybe it wasn't working for the other field?
Consider that usually boolean settings linked to configuration parameters work in a different way than you might expect: when they are set to `False` the configuration parameter is deleted.

website_user_login_redirect_custom/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,55 @@
...
+                    <div class="o_setting_right_pane">
+                        <label
+                            for="website_login_redirect_enabled"
+                            string="Redirect on Login/Signup​"
+                        />
+                        <span
+                            class="fa fa-lg fa-globe"
+                            title="Values set here are website-specific."
+                            groups="website.group_multi_website"
+                        />
```
SirPyTech: **suggestion**: The values are not website-specific: website-specifc values are related to `website_id`.
We are defining values that are system-wide so please either remove this node or adjust the implementation accordingly.

website_user_login_redirect_custom/readme/CONFIGURE.md
```diff
@@ -0,0 +1,4 @@
+- Go to Website → Configuration → Settings
+- Scroll to the Features section
+- Enable the checkbox “On Login/Signup Redirect”
```
SirPyTech: **typo**: 
```suggestion
- Enable the checkbox “Redirect on Login/Signup”
```

website_user_login_redirect_custom/data/ir_config_parameter_data.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo>
```
SirPyTech: **issue**: Please add `noupdate`: we don't want the configured values to be overwritten when the module is updated.

website_user_login_redirect_custom/models/res_config_settings.py
```diff
@@ -0,0 +1,84 @@
...
+    @api.model
+    def get_values(self):
+        """Load boolean value from ir.config_parameter."""
+        res = super().get_values()
+        enabled_param = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("website_user_login_redirect_custom.enabled", "0")
+        )
+        res["website_login_redirect_enabled"] = enabled_param == "1"
```
SirPyTech: **thought**: Here and wherever you are converting a string to a boolean, you might want to use `str2bool`; see for example https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/addons/mail/models/mail_thread.py#L896:
```py
bounce_alias_static = tools.str2bool(self.env['ir.config_parameter'].sudo().get_param("mail.bounce.alias.static", "False"))
```

website_user_login_redirect_custom/tests/test_website_login_redirect.py
```diff
@@ -0,0 +1,246 @@
...
+            "shop",
+            "http://evil.com",
+            "https://test.com",
+            "//phish.com",
+            "javascript:alert(1)",
+            "data:text/html;base64,xxx",
+            "vbscript:msgbox(1)",
+        ]
+        for url in invalid_urls:
+            with self.subTest(url=url):
```
SirPyTech, line 71: **praise**: Thanks for writing the tests, and this `subTest` usage is a nice touch 🤩

website_user_login_redirect_custom/tests/test_website_login_redirect.py
```diff
@@ -0,0 +1,246 @@
...
+                with self.assertRaises(ValidationError):
```
SirPyTech: **note**: Please check that the raised `ValidationError` is the one we are expecting, so we won't get any false positive.

website_user_login_redirect_custom/tests/test_website_login_redirect.py
```diff
@@ -0,0 +1,246 @@
+# Copyright (C) 2025 Cetmix OÜ
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo.exceptions import ValidationError
+from odoo.tests import common, tagged
+
+
+@tagged("post_install", "-at_install")
+class TestWebsiteLoginRedirect(common.TransactionCase):
```
SirPyTech: **suggestion**: Most of these tests check that Odoo is working properly, for instance creating an inherited view or getting/setting the configuration fields.
Many tests also check again and again the same behavior, for instance getting/setting configuration fields.

I'm a big fan of tests, but this is too much in my opinion: please only check that the behavior added by this module works as expected and let Odoo worry about testing its own behavior.

Moreover the probably most critical piece of code - the controller - is not being tested.

website_user_login_redirect_custom/controllers/main.py
```diff
@@ -0,0 +1,68 @@
...
+            return False
+        url = url.strip()
+        if request.env["res.config.settings"].is_valid_redirect_url(url):
+            return url
+
+        return False
+
+    @http.route("/web/login", type="http", auth="public", website=True, sitemap=False)
+    def web_login(self, redirect=None, **kw):
+        """Override standard login route to apply a custom redirect after login."""
```
SirPyTech: **thought**: The method
```py
def _get_login_redirect_url(uid, redirect=None):
    """ Decide if user requires a specific post-login redirect, e.g. for 2FA, or if they are
    fully logged and can proceed to the requested URL
    """
```
(from https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/addons/web/controllers/main.py#L299-L302)

looks like the perfect hook for the new behavior, like `_login_redirect` that is used by `portal` to redirect portal users to `/my` (https://github.com/odoo/odoo/blob/cc0060e889603eb2e47fa44a8a22a70d7d784185/addons/portal/controllers/web.py#L17-L20), have you tried those?

SirPyTech (changes requested):
Thanks for the PR!

website_user_login_redirect_custom/models/res_config_settings.py
Mohajiro: Thank you for pointing this out! You're absolutely right. I've replaced the manual get/set methods with the config_parameter attribute as suggested.

website_user_login_redirect_custom/views/res_config_settings_views.xml
Mohajiro: Good catch! Thank you! I've removed the globe icon.

website_user_login_redirect_custom/readme/CONFIGURE.md
Mohajiro: Thank you for the suggestion! I've updated the label text to 'Redirect on Login/Signup' as recommended."

website_user_login_redirect_custom/data/ir_config_parameter_data.xml
Mohajiro: Sure, thank you!

website_user_login_redirect_custom/models/res_config_settings.py
Mohajiro: Thank you for the suggestion and the reference! I've replaced the manual string comparison with str2bool() for more robust boolean conversion that handles multiple formats.

website_user_login_redirect_custom/tests/test_website_login_redirect.py
Mohajiro: You're absolutely right about avoiding false positives. I've updated the tests to verify the specific ValidationError message we expect to catch.

website_user_login_redirect_custom/tests/test_website_login_redirect.py
Mohajiro: Thanks you a lot for the motivation!

website_user_login_redirect_custom/tests/test_website_login_redirect.py
Mohajiro: Excellent point about focusing on our custom logic. I've removed the tests that were verifying base Odoo functionality and kept only those testing our specific business logic.

website_user_login_redirect_custom/controllers/main.py
Mohajiro: Thank you for the fantastic code review! I've done my best to implement all of your suggestions throughout the codebase. 

I hope I haven't missed anything and that the code is now significantly improved. Your expertise and thorough review have been incredibly valuable in helping me learn Odoo best practices.

Thank you again for sharing your knowledge and experience!

website_user_login_redirect_custom/controllers/main.py
```diff
@@ -0,0 +1,64 @@
...
+            return False
+
+        # Do not redirect backend/internal users
+        if request.env.user.has_group("base.group_user"):
+            return False
+
+        enabled = (
+            request.env["ir.config_parameter"]
+            .sudo()
+            .get_param("website_user_login_redirect_custom.enabled", False)
```
SirPyTech: **suggestion**: `str2bool` probably expects a string
```suggestion
            .get_param("website_user_login_redirect_custom.enabled", "False")
```
but it works fine already so you can also leave it as is.

website_user_login_redirect_custom/controllers/main.py
```diff
@@ -0,0 +1,64 @@
...
+
+        return False
+
+    def _login_redirect(self, uid, redirect=None):
+        """
+        Override login redirect behavior for portal users
+        Redirect to custom URL instead of /my when feature is enabled
+        """
+        # First call parent to get default redirect (e.g., portal users to /my)
+        redirect = super()._login_redirect(uid, redirect)
```
SirPyTech: **polish**: `redirect` is a _kwarg_ so it's better to use it as such
```suggestion
        redirect = super()._login_redirect(uid, redirect=redirect)
```
Otherwise if another _kwarg_ is added in `_login_redirect`'s signature before `redirect`, you'd be assigning this value to the wrong method argument.

SirPyTech: **polish**: We are overwriting the method argument with the returned value, please assign a different variable name to the returned value.

website_user_login_redirect_custom/tests/test_website_login_redirect.py
SirPyTech: I'm still a bit concerned that the controller isn't being tested but :+1:

website_user_login_redirect_custom/tests/test_website_login_redirect.py
```diff
@@ -0,0 +1,122 @@
...
+    """Test website login redirect functionality"""
+
+    def setUp(self):
+        super().setUp()
+        self.config_parameter = self.env["ir.config_parameter"].sudo()
+        # Reset to default state before each test
```
SirPyTech: **suggestion**: Resetting the environment is already done in the usual testing flow, so there's no need for this

website_user_login_redirect_custom/tests/test_website_login_redirect.py
```diff
@@ -0,0 +1,122 @@
...
+        self.config_parameter.set_param(
+            "website_user_login_redirect_custom.enabled", "0"
+        )
+        self.config_parameter.set_param("website_user_login_redirect_custom.url", "/")
+
+        # Create test users
```
SirPyTech: **suggestion**: The users are being recreated before each test and that's not really necessary, we could avoid that by using `SavepointCase` and moving this to a `setUpClass` override.

website_user_login_redirect_custom/models/res_config_settings.py
SirPyTech: :+1:

website_user_login_redirect_custom/views/res_config_settings_views.xml
SirPyTech: :+1:

website_user_login_redirect_custom/readme/CONFIGURE.md
SirPyTech: :+1:

website_user_login_redirect_custom/data/ir_config_parameter_data.xml
SirPyTech: :+1:

website_user_login_redirect_custom/models/res_config_settings.py
SirPyTech: :+1:

website_user_login_redirect_custom/tests/test_website_login_redirect.py
SirPyTech: :+1:

website_user_login_redirect_custom/tests/test_website_login_redirect.py
SirPyTech: :+1:

website_user_login_redirect_custom/controllers/main.py
SirPyTech: :+1:

website_user_login_redirect_custom/views/res_config_settings_views.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<odoo noupdate="1">
```
SirPyTech: **issue**: Usually the views are reset on module update, this is because if the user screws up a view then they are no more able to do anything so a module update would remove their edits.
In order to have permanent edits to a view, the user should create an inherited view.

`noupdate` is typically used on data, like the configuration parameters.

SirPyTech (changes requested):
Thanks for adapting the code

website_user_login_redirect_custom/controllers/main.py
Mohajiro: Thanks, fixed.

website_user_login_redirect_custom/controllers/main.py
Mohajiro: Thank you a lot, fixed too.

Mohajiro: Renamed, thank you!

website_user_login_redirect_custom/tests/test_website_login_redirect.py
Mohajiro: Thanks, done.

website_user_login_redirect_custom/tests/test_website_login_redirect.py
Mohajiro: I see, is it nice right now?

website_user_login_redirect_custom/controllers/main.py
SirPyTech: :+1:

website_user_login_redirect_custom/controllers/main.py
SirPyTech: :+1:

SirPyTech: :+1:

website_user_login_redirect_custom/tests/test_website_login_redirect.py
SirPyTech: :+1:

website_user_login_redirect_custom/tests/test_website_login_redirect.py
SirPyTech: :+1:

website_user_login_redirect_custom/views/res_config_settings_views.xml
SirPyTech: :+1:

SirPyTech (approved):
Thanks, I reviewed the code and it's fine by me.

You could also squash all the commits together.

francesco-ooops:
@ivs-cetmix can you merge?

ivs-cetmix:
> @ivs-cetmix can you merge?

Not yet, we need someone from the @OCA/website-maintainers to check this.

francesco-ooops:
@OCA/website-maintainers ping :)

ivs-cetmix:
Ok, looks like no one loves Odoo 14.0 any more, so ...
/ocabot merge nobump

PR #1159, opened 2026-01-21
[16.0][ADD] website_blog_scheduled_publication: add new module

marcelsavegnago:
## Summary

This PR adds a new module `website_blog_scheduled_publication` that allows users to schedule blog posts for automatic publication at a specific date and time.

### Features

- **Scheduled Publication Date**: Adds a datetime field to blog posts for scheduling publication
- **Quick Publish Button**: One-click immediate publication from the form view button box
- **Schedule Wizard**: User-friendly wizard dialog to set the publication date
- **Visual Feedback**: Information banner displays the scheduled date on the post form
- **Smart Publication Logic**: Automatically prevents immediate publication if a future date is set
- **Automatic Cron Job**: Runs every hour to check and publish scheduled posts
- **Notification System**: Sends notifications to blog followers when posts are published
- **Search Filters**: Filter posts by Published/Not Published status and group by Publication Date or Scheduled Publication Date

### Technical Details

- Extends `blog.post` model with new fields (`scheduled_publication_date`, `schedule_type`, `next_departure`) and methods
- Includes a transient model `blog.post.schedule.date` for the scheduling wizard
- Inherits and enhances form and search views from `website_blog`
- Handles edge cases: setting past dates, clearing scheduled dates, multiple record writes
- Comprehensive test coverage with 25+ test cases

### Dependencies

- `website_blog`

## Test Plan

- [x] Install the module on a clean database
- [x] Create a blog post with a future scheduled date and verify it's not published immediately
- [x] Create a blog post with a past scheduled date and verify it publishes immediately
- [x] Test the Schedule wizard dialog
- [x] Test the Publish Now button
- [x] Run the cron job manually and verify scheduled posts are published
- [x] Verify notifications are sent to blog followers
- [x] Test search filters and grouping options
- [x] Run all unit tests: `pytest website_blog_scheduled_publication/tests/`


yostashiro:
@marcelsavegnago I guess you could take a similar approach as https://github.com/OCA/e-commerce/blob/15.0/website_sale_product_publish_date/models/product_template.py? I think there could even be a generic module with a mixin model which can be inherited by regular models if it fits the core architecture (I haven't taken a thorough look on it).

marcelsavegnago:
> @marcelsavegnago I guess you could take a similar approach as https://github.com/OCA/e-commerce/blob/15.0/website_sale_product_publish_date/models/product_template.py? I think there could even be a generic module with a mixin model which can be inherited by regular models if it fits the core architecture (I haven't taken a thorough look on it).

@yostashiro You’re right — that approach looks like a solid reference point.

For now I’m handling a very specific case to solve the immediate need, but once I have a bit more time I’ll take a closer look and see if we can support a more generic solution (like a mixin / small reusable module) that fits well with the core architecture.

website_blog_scheduled_publication/models/blog_post.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+class BlogPost(models.Model):
+    _inherit = "blog.post"
+
+    scheduled_publication_date = fields.Datetime(
+        help="If set, the blog post will be automatically published on "
+        "this date and time. Leave empty to publish immediately.",
+    )
+    schedule_type = fields.Selection(
```
ivs-cetmix: You can use a boolean field if you are not expecting any other types besides "now" and "later".

website_blog_scheduled_publication/models/blog_post.py
```diff
@@ -0,0 +1,225 @@
...
+        [("now", "Publish now"), ("scheduled", "Publish on")],
+        string="Schedule",
+        default="now",
+        required=True,
+        help="Choose when to publish this blog post.",
+    )
+    next_departure = fields.Datetime(
```
ivs-cetmix: Why do we need another datetime field besides `scheduled_publication_date`?

website_blog_scheduled_publication/views/blog_post_views.xml
```diff
@@ -0,0 +1,104 @@
...
+                    role="alert"
+                    attrs="{'invisible': ['|', ('schedule_type', '!=', 'scheduled'), ('scheduled_publication_date', '=', False)]}"
+                >
+                    <div class="o_blog_post_scheduled">
+                        <strong>
+                            <span
+                                name="next_departure_text"
+                            >This post is scheduled for </span>
+                            <field
+                                name="next_departure"
```
ivs-cetmix: Can't we use the `scheduled_publication_date` here?

ivs-cetmix (changes requested):

website_blog_scheduled_publication/models/blog_post.py
marcelsavegnago: Done

website_blog_scheduled_publication/views/blog_post_views.xml
marcelsavegnago: Done

website_blog_scheduled_publication/models/blog_post.py
marcelsavegnago: DONE

website_blog_scheduled_publication/models/blog_post.py
```diff
@@ -0,0 +1,207 @@
...
+        )
+        if not posts_to_notify:
+            return False
+        return super(BlogPost, posts_to_notify.sudo())._check_for_publication(vals)
+
+    def _has_future_scheduled_date(self, post, vals, now):
+        """Check if post has a future scheduled date."""
+        if post.scheduled_publication_date and post.scheduled_publication_date > now:
+            return True
+        scheduled_date = _to_datetime(vals.get("scheduled_publication_date"))
```
ivs-cetmix, line 109: Question: isn't `scheduled_publication_date` a Python datetime object already?

ivs-cetmix (changes requested):
Thank you for you contribution! 
Some note - not an issue - how about adding a calendar view for the `blog.post` model so it would be easier to see the scheduled posts?

website_blog_scheduled_publication/models/blog_post.py
marcelsavegnago: Good point. When accessed from a record (post.scheduled_publication_date), it's already a datetime object. However, vals.get("scheduled_publication_date") can be a string when values come from forms, wizards, or API calls.
The _to_datetime() helper handles both cases safely — it converts strings to datetime and passes through datetime objects unchanged.
We're open to suggestions if you have a better approach.

ivs-cetmix (approved):
Code review LGTM

website_blog_scheduled_publication/models/blog_post.py
ivs-cetmix: Noted, let's leave as is for now. Would be good to add some comment in the code though to avoid same question in the future.

website_blog_scheduled_publication/models/blog_post.py
```diff
@@ -0,0 +1,224 @@
...
+            dialog_size="medium",
+        )
+        return action
+
+    def _check_for_publication(self, vals):
+        """Override to handle publication notifications properly.
+
+        Skip notifications for posts that are scheduled for the future.
+        Use sudo() to ensure notifications can be sent regardless of user
+        permissions, as this is a system action that should always succeed.
```
hbrunn: in which contexts does this happen?

hbrunn:
what's the point of this module given Odoo already supports setting `post_date` to the future?

website_blog_scheduled_publication/models/blog_post.py
```diff
@@ -0,0 +1,224 @@
...
+    post.scheduled_publication_date), Odoo automatically returns a Python
+    datetime object. However, when working with vals dictionaries (from forms,
+    wizards, or API calls), the value may come as a string and needs to be
+    converted to datetime for proper comparison operations.
+    """
+    if not value:
+        return False
+    if isinstance(value, str):
+        return fields.Datetime.from_string(value)
+    return value
```
StefanRijnhart, line 22: Perfect example of AI slop. This method has the exact same resilience as `from_string` itself! https://github.com/odoo/odoo/blob/16.0/odoo/fields.py#L2216-L2239

PR #1162, opened 2026-02-10
[16.0][ADD] web_editor_clean_url

hbrunn:
This module allows administrators to configure cleanup rules for links
inserted by their users, as inexperienced people tend to just copy&paste
whatever they find in their address bar, which might contain data you do
not want to see published on your website, be it tracking tokens or
session identifiers.

Configuration
=============

You find the configuration in Website / Configuration / Clean URLs.

For every configuration, you can fill in:

``Hostname``: allowing to restrict this configuration to some domain

``Remove Query Parameter``: removes parameters from the URL. Separate
with comma

``Replace Regex``: replaces a whole portion of the url

``Substitution``: if the regex contains brackets, you can refer to the
matched content with \\1, \\2 etc

Usage
=====

1. Review the configuration
2. Edit a website page, insert links and observe they are transformed as
   configured after saving


simahawk (approved):
Nice! Only code review 👍

PR #1169, opened 2026-03-08
[18.0][ADD] website_sale_partner_firstname: surface first and last name fields

aaronNgu:
- New module that replaces the single "Name" field with separate "First name" and "Last name" fields on the checkout (/shop/address) and portal             
  (/my/account) pages                                 
- Depends on partner_firstname for name computation                                                                                                         
- Both fields are mandatory; whitespace-only values are rejected


ivs-cetmix (approved):
Hi @aaronNgu , than you for your contribution! Not a blocker: code is fine, however docstings in functions always make it better 😄

PR #1177, opened 2026-04-17
[18.0][ADD] website_exclude_sitemap

edescalona:
@BinhexTeam

T21540

### Context

Odoo includes every public website URL in the generated sitemap. Some of those URLs are useful for direct navigation but add little or no SEO value when exposed in `/sitemap.xml` (`/livechat`, feeds, apply forms, user profiles...). This module keeps the sitemap focused on strategic public pages while leaving normal website access untouched.

### What it does

- Adds a per-website `sitemap_excluded_paths` text field, editable from **Website > Configuration > Settings > Sitemap Exclusions**.
- Overrides `website._enumerate_pages()` to drop the entries whose path matches one of the configured patterns.
- Clears the cached sitemap attachments when the exclusions change, when `website.page` records are created or deleted, and when their `url`, `website_published`, `website_indexed`, `date_publish` or `website_id` changes.
- Adds a **Reload Sitemap** button to clear the sitemap cache of the current website manually.

The module only affects sitemap generation: it does not unpublish pages, change access rights, or block direct access to the excluded URLs.

### Pattern syntax

Values are separated by line breaks, commas or semicolons; lines starting with `#` are comments.

| Pattern | Matches | Does not match |
| --- | --- | --- |
| `/livechat` | `/livechat` | `/livechat/room` |
| `/customers/` | `/customers`, `/customers/acme`, `/customers/acme/2024` | `/customers-list` |
| `/blog/*/feed` | `/blog/news/feed` | `/blog/a/b/feed`, `/blog/news` |
| `/solutions*` | `/solutions`, `/solutions-cloud` | `/solutions/cloud` |
| `/jobs/**` | `/jobs`, `/jobs/apply/1`, `/jobs/a/b/c` | `/jobs-list` |

- `*` matches inside a single path segment and never crosses a `/`.
- `**` crosses `/`, matching any number of segments.
- A pattern ending in `/` or `/**` is a prefix: it excludes that path and everything below it.
- Any other pattern matches that exact path; the trailing slash of the URL is ignored.

The default value only covers paths present in any Odoo website (`/customers/`, `/livechat`, `/blog/*/feed`, `/jobs/apply/`, `/profile/`); deployment-specific paths are meant to be added by the administrator.

### Implementation notes

- Patterns are compiled to regexes cached with `tools.ormcache` keyed on the raw exclusions text, so they are compiled once per configuration instead of once per sitemap entry.
- `write()` on `website` clears the registry cache when `sitemap_excluded_paths` changes, so the compiled patterns are invalidated too.
- Sitemap cache clearing removes the `/sitemap-<website_id>-%` binary attachments.
- Pages without `website_id` (shared across websites) invalidate the sitemap cache of every website.


website_exclude_sitemap/i18n/es_ES.po
rrebollo, line 1: Idem to es.po, so why? Get rid of it.

website_exclude_sitemap/wizards/res_config_settings.py
```diff
@@ -0,0 +1,30 @@
+# Copyright 2026 Binhex <https://www.binhex.cloud>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import _, fields, models
```
rrebollo:
```suggestion
from odoo import fields, models
```

website_exclude_sitemap/wizards/res_config_settings.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def action_reload_sitemap(self):
+        self.ensure_one()
+        website = self.website_id or self.env["website"].get_current_website()
+        website._clear_sitemap_cache()
+        return {
+            "type": "ir.actions.client",
+            "tag": "display_notification",
+            "params": {
+                "title": _("Sitemap reloaded"),
```
rrebollo:
```suggestion
                "title": self.env._("Sitemap reloaded"),
```

website_exclude_sitemap/wizards/res_config_settings.py
```diff
@@ -0,0 +1,30 @@
...
+                "message": _(
```
rrebollo:
```suggestion
                "message": self.env._(
```

rrebollo (commented):
please consider my suggestions

edescalona:
Hi @rrebollo , thanks for the review; the suggestions have been applied. @pedrobaeza

rrebollo (approved):
Code Review. LGTM!

pedrobaeza:
Odoo already allows to control which pages are published:

imagen

so I don't get the sense of this.

edescalona:
Hi @pedrobaeza , thanks for the review. That column only affects `website.page` records, but the goal is to also control controller routes, which cannot be managed via the UI.

For example, with `website_customer` and `website_blog` installed, `/sitemap.xml` contains 45 URLs, but only one (`/contactus`) originates from a page. The other 44 come from controllers: `/customers` and its 21 routes (e.g.,
`/customers/industry/...`, `/partners/...`), `/blog/<blog>/feed`, `/website/info`, etc.

Furthermore, the standard core toggle works on a record-by-record basis, whereas here, a pattern like `/customers/` or `/blog/*/feed` covers entire families of URLs, including those created later.

I’ve added a test (`test_enumerate_pages_skips_excluded_controller_urls`) that takes an enumerated URL lacking a `website.page` record and verifies that the pattern excludes it from the sitemap.

Does that make sense to you, or do you see another way to handle this case?

pedrobaeza:
Well, on blogs you can also mark to not track, but OK if you consider the accelerator for the main path. The problem I see with this is the UI, as if you go to the blog page/category and mark to track or not, will contradict this supra configuration.

edescalona:
@pedrobaeza  I could add a note in the settings explaining that the list takes precedence over the per-record markup—and also affects the `website.page` view—though this wouldn't apply to the blog in this module because it would introduce unwanted dependencies. Alternatively, I could mention in `description.md` that this module overrides the markup. Which option do you think is best?

pedrobaeza:
OK, just mention it in the ROADMAP.md file and let people to decide.

edescalona:
ready @pedrobaeza

pedrobaeza:
OK, a second review is needed.

rrebollo (approved):
Code Review. LGTM!

jelenapoblet (approved):
This module solves a problem we have been dealing with for a long time: customizing the sitemap without modifying Odoo’s standard files.
I wasn't even aware you where working on this, I was modifying the sitemap manually with the marketing team all the time.
LGTM

edescalona:
Ready @pedrobaeza Can we merge now?

pedrobaeza:
@OCA/pypi-support to reserve the name

sbidoul:
Name reserved

PR #1180, opened 2026-05-12
[17.0][ADD] portal_invitation_by_website: New Module

cristina-hidalgo-tecnativa:
This module extends the portal invitation wizard to allow restricting portal users to a specific website. The selected website_id is propagated to res.partner, which causes `get_base_url()` to return the correct domain in all outgoing Odoo emails for that contact.

@Tecnativa TT61856
@pedrobaeza @pilarvargas-tecnativa @christian-ramos-tecnativa


portal_invitation_by_website/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1 @@
+- Cristina Hidalgo \<cristina.hidalgo@tecnativa.com\>
```
pedrobaeza: Don't put emails here and use the 2 level structure.

portal_invitation_by_website/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,7 @@
+This module extends the portal invitation wizard to allow restricting portal
+users to a specific website in multi-website environments.
+
+When inviting contacts as portal users from the backend, you can now select a
+website at the wizard level (applied to all contacts) or override it
+per-contact. The invitation email URL will automatically use the selected
```
pedrobaeza, line 6: I don't see the part of the URL in the code.

portal_invitation_by_website/models/portal_wizard.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def action_invite_again(self):
+        self._apply_website_to_partner()
+        return super().action_invite_again()
+
+    def _apply_website_to_partner(self):
+        """Set website_id on the partner so the invitation URL uses the correct
+        website domain (get_base_url() reads partner.website_id.domain)."""
+        for user in self:
+            if user.website_id != user.partner_id.website_id:
```
pedrobaeza, line 50: Can't be done in the user creation vals dictionary preparation?

portal_invitation_by_website/readme/CONTRIBUTORS.md
cristina-hidalgo-tecnativa: Removed

portal_invitation_by_website/readme/DESCRIPTION.md
cristina-hidalgo-tecnativa: Modified

pedrobaeza: But the solution isn't to remove the description, but to make the code to behave like that, as we want to have the invitation URL with the proper website. I would say that's done passing `website_id` context to the code that makes such invitation.

portal_invitation_by_website/readme/CONTRIBUTORS.md
```diff
@@ -0,0 +1,3 @@
+* Tecnativa:
```
pedrobaeza: Missing URL. Please check the structure in other contributions.

portal_invitation_by_website/models/portal_wizard.py
cristina-hidalgo-tecnativa: There's no vals dict hook in the base method, so doing it that way would require duplicating the full `_create_user` body. 
The post-creation write inside `_create_user` avoids that risk while still keeping the assignment within the same method.

pedrobaeza: Yeah, but doing it in 2 steps (create it, and then update it), may lead to unexpected results. You can do the following:

- Inject in the methods a context `portal_invitation_website_id` in the corresponding methods (`action_grant_access`, etc).
- Override `_create_user_from_template` from `res.users`.
- If the context variable is there, inject the value in `values` argument.
- Call super with that new `values`.

And this may fix also the URL of the invitation mail template.

cristina-hidalgo-tecnativa: Done following your suggestion. `action_grant_access` now injects `portal_invitation_website_id` in the context, picked up by `_create_user_from_template` on `res.users` to set `website_id` in the creation vals — so `partner.website_id` is set before the email is sent and `get_base_url(`) returns the correct domain. 

Also description restored.

I hope It is ok now.

portal_invitation_by_website/readme/CONTRIBUTORS.md
cristina-hidalgo-tecnativa: Fixed now.

portal_invitation_by_website/models/portal_wizard.py
pedrobaeza: Don't you need to inject the context in other methods as well?

cristina-hidalgo-tecnativa: Yes, also injected in `action_invite_again`.

portal_invitation_by_website/models/portal_wizard.py
```diff
@@ -0,0 +1,54 @@
+from odoo import api, fields, models
+
+
+class PortalWizard(models.TransientModel):
+    _inherit = "portal.wizard"
+
+    website_id = fields.Many2one(
+        comodel_name="website",
+        string="Website",
+        help="Apply this website to all contacts below. "
```
pedrobaeza:
```suggestion
        help="Apply this website to all the contacts below. "
```

portal_invitation_by_website/models/portal_wizard.py
```diff
@@ -0,0 +1,54 @@
...
+        "The invitation email URL will match this website's domain. "
+        "Leave empty to keep each contact's current website.",
+    )
+
+    def action_apply_website(self):
+        """Update website_id on all partners without modifying portal access or sending
+        emails."""
```
pedrobaeza: According PEP257, dosctring should be put this way:

```suggestion
        """Update website_id on all partners without modifying portal access or sending
        emails.
        """
```

portal_invitation_by_website/models/portal_wizard.py
```diff
@@ -0,0 +1,54 @@
...
+    @api.depends("wizard_id.website_id", "partner_id.website_id")
+    def _compute_website_id(self):
+        for user in self:
+            user.website_id = user.wizard_id.website_id or user.partner_id.website_id
+
+    def action_grant_access(self):
+        return super(
+            PortalWizardUser,
+            self.with_context(portal_invitation_website_id=self.website_id.id),
+        ).action_grant_access()
```
pedrobaeza: For clearer code:

```suggestion
        self = self.with_context(portal_invitation_website_id=self.website_id.id)
        return super().action_grant_access()
```

portal_invitation_by_website/models/portal_wizard.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def action_invite_again(self):
+        self._apply_website_to_partner()
+        return super(
+            PortalWizardUser,
+            self.with_context(portal_invitation_website_id=self.website_id.id),
+        ).action_invite_again()
+
+    def _apply_website_to_partner(self):
```
pedrobaeza, line 48: Why keeping this method and the parent one?

cristina-hidalgo-tecnativa: `_apply_website_to_partner` is still needed in `action_invite_again` because no user is created in that flow — `_create_user_from_template` is never called, so the context has no effect there. 
Setting `partner.website_id` directly is the only way to ensure `get_base_url()` returns the correct domain for the invitation email. 
It's also used by `action_apply_website` for the bulk website update without granting access.

portal_invitation_by_website/models/portal_wizard.py
cristina-hidalgo-tecnativa: Fixed.

portal_invitation_by_website/models/portal_wizard.py
cristina-hidalgo-tecnativa: Fixed.

portal_invitation_by_website/models/portal_wizard.py
cristina-hidalgo-tecnativa: Modified.

pedrobaeza (approved):

marcelsavegnago (approved):
LGTM

PR #1185, opened 2026-05-18
[17.0][FIX] portal_invitation_by_website: use website name in portal invitation email

cristina-hidalgo-tecnativa:
Override the portal invitation email template to display the website name instead of the company name when a website is selected in the wizard. 

Falls back to the company name when no website is set.

@Tecnativa TT61856
@pedrobaeza @pilarvargas-tecnativa


pedrobaeza (changes requested):
This is not valid, as translations won't be available, being an XML-ID of another module.

One thing you can do is to intercept `def ref(..`) method to return a different template from this module when getting the reference of `portal.mail_template_data_portal_welcome`.

This is still not the best, as you have to re-translate the whole template, but at least it's translatable.

portal_invitation_by_website/models/ir_model_data.py
```diff
@@ -0,0 +1,18 @@
+from odoo import api, models
+
+PORTAL_WELCOME_XMLID = "portal.mail_template_data_portal_welcome"
+WEBSITE_WELCOME_XMLID = (
+    "portal_invitation_by_website.mail_template_data_portal_welcome_website"
```
pedrobaeza: This XML-ID is not the correct one.

pilarvargas-tecnativa: Perhaps there is a better approach; ideally, the website should be sent instead of the company name, that's true, but I'm not entirely convinced by this solution. Perhaps we should just leave it as it is.

cristina-hidalgo-tecnativa: > This XML-ID is not the correct one.

Modified now.

pedrobaeza (approved):
Why the template doesn't appear in translations?

pilarvargas-tecnativa:
> Why the template doesn't appear in translations?

ping @cristina-hidalgo-tecnativa

pedrobaeza (approved):
/ocabot merge minor

PR #1187, opened 2026-05-20
[17.0][IMP] portal_invitation_by_website: use website domain in portal invitation footer

pilarvargas-tecnativa:
Use the website domain in the portal invitation email footer instead of always using the company website.

A single company can manage multiple websites, so using the company website may display the default website URL instead of the website from which the portal invitation is being sent.

@Tecnativa TT61856

@pedrobaeza @cristina-hidalgo-tecnativa please review


portal_invitation_by_website/migrations/17.0.1.0.0/noupdate_changes.xml
```diff
@@ -0,0 +1,160 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
pedrobaeza: Don't do this migration script. It was just merged yesterday. We do it by hand.

pedrobaeza: Anyway, it won't work as is.

pilarvargas-tecnativa: removed

pedrobaeza (commented):
The translation change is missing.

pilarvargas-tecnativa:
> The translation change is missing.

done

pedrobaeza (approved):
/ocabot merge patch

PR #1193, opened 2026-06-09
[18.0][FIX] website_forum_subscription: avoid nested links in forum cards

pilarvargas-tecnativa:
Wrap the public subscription link in an <object> tag to avoid rendering an anchor inside the forum card anchor, which breaks the page layout for public users.
Like Odoo https://github.com/odoo/odoo/blob/e6a402933822adcd85f771159ab4abf58bda5d24/addons/website_forum/views/forum_forum_templates_forum_all.xml#L56-L61

Before:

image

After:

image

@Tecnativa 

@pedrobaeza @victoralmau please review

### DIFF
--- a/website_forum_subscription/templates/website_forum_templates.xml
+++ b/website_forum_subscription/templates/website_forum_templates.xml
@@ -14,9 +14,11 @@
                 t-value="request.env.user.has_group('base.group_public')"
             />
             <div t-if="user_public">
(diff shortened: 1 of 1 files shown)


pedrobaeza (approved):
What a hack...

Let's fast-track it to fix it.

/ocabot merge patch

PR #1197, opened 2026-06-12
[18.0][IMP] website_forum_subscription: bypass captcha on forum subscriptions

pilarvargas-tecnativa:
Pending from:
- [ ] https://github.com/odoo/odoo/pull/269697

Forum subscriptions rely on the `website_mail` follow widget and are expected to behave as an immediate subscribe/unsubscribe action.

When `website_cf_turnstile` is enabled, the captcha validation flow may introduce a noticeable delay before the subscription is processed, resulting in a degraded user experience.

Mark forum subscription widgets with
`o_website_mail_follow_no_recaptcha` to use the opt-out mechanism provided by `website_mail` and bypass captcha validation for these interactions.

TT62994

### DIFF
--- a/website_forum_subscription/templates/website_forum_templates.xml
+++ b/website_forum_subscription/templates/website_forum_templates.xml
@@ -3,7 +3,7 @@
     <!-- Uses the gears of website_mail for the UI enhancements -->
     <template id="follow">
         <div
-            t-attf-class="input-group js_follow #{div_class}"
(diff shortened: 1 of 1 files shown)


pilarvargas-tecnativa:
Closing this PR as it is no longer needed.

The solution has moved to website_mail, where captcha validation is skipped for authenticated users. This automatically covers website_forum_subscription, so no widget-specific change is required anymore.

PR #1199, opened 2026-06-12
[18.0][ADD] website_sale_vat_label: Introduction of country-specific …

jarcosmts:
…VAT labels

Create the new module to dynamically update the VAT field label in the frontend address form. Override _prepare_address_form_values in WebsiteSale controller to fetch country_sudo.vat_label. Add module documentation (description, context, usage, and history) following OCA guidelines.


jarcosmts:
Hi maintainers. I have opened this new PR with the history fully squashed into a single clean commit. The custom HttpCase tests cover both country-specific and default VAT label branches, and the CI suite is passing successfully. Codecov might still show the old coverage percentage due to cache from the previous PR, but the method is fully tested locally.

PR #1201, opened 2026-06-16
[16.0][ADD] portal_pager_size

halbtonjazz:
## Description

New module that adds a configurable page size selector to the portal
pager. Portal users can pick how many records are shown per page
(10/20/40/80/100 by default) via a dropdown next to the pagination
controls. The choice is applied through a `limit` query parameter and
kept while paginating, sorting, filtering and searching.

Allowed values are whitelisted through the
`website_portal_pager_limit.options` system parameter; any invalid or
out-of-list value falls back to the standard portal page size (80), so a
misconfiguration can never break pagination.

## How it works

- Controller exposes `_items_per_page` as a property reading the
  validated `limit`, so every `/my/*` route benefits without overriding
  each route.
- `portal.pager` is inherited to render the selector (kept visible even
  on single-page lists).
- Frontend JS propagates the limit to pager links.

Task: 0350


website_portal_pager_limit/controllers/portal.py
```diff
@@ -0,0 +1,44 @@
...
+    def _get_portal_pager_limit_options(self):
+        """Return the whitelist of allowed page sizes.
+
+        Misconfigured values (non digits, empty string) are discarded so a
+        broken system parameter can never disable the portal pagination.
+        """
+        param = (
+            request.env["ir.config_parameter"]
+            .sudo()
+            .get_param(OPTIONS_PARAM, default="10,20,40,80,100")
```
geomer198: Maybe better use default value=False and then use `DEFAULT_LIMIT_OPTIONS` const without split methods?
```py
param = request.env["ir.config_parameter].sudo().get_param(OPTIONS_PARAM, default=False)
if not param:
    return DEFAULT_LIMIT_OPTIONS
return [int(x.strip()) for x in param.split(",") if x.strip().isdigit()] or DEFAULT_LIMIT_OPTIONS

File: website_portal_pager_limit/tests/test_portal_pager_limit.py [hunk 2]
```diff
@@ -0,0 +1,207 @@
...
+            httprequest=SimpleNamespace(args=args or {}),
+        )
+        _request_stack.push(fake_request)
+        try:
+            yield fake_request
+        finally:
+            _request_stack.pop()
+
+
+class TestPagerLimitOptions(TestPortalPagerLimitCommon):
```
geomer198: Please split this file.
Use for each test class separate file for tests.

geomer198 (commented):
Could you please check code review result.

website_portal_pager_limit/controllers/portal.py
halbtonjazz: done

website_portal_pager_limit/tests/test_portal_pager_limit.py
halbtonjazz: done

website_portal_pager_limit/controllers/portal.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def _get_portal_pager_limit_options(self):
+        """Return the whitelist of allowed page sizes.
+
+        Misconfigured values (non digits, empty string) are discarded so a
+        broken system parameter can never disable the portal pagination.
+        """
+        param = request.env["ir.config_parameter"].sudo().get_param(OPTIONS_PARAM)
+        if not param:
+            return list(DEFAULT_LIMIT_OPTIONS)
```
geomer198, line 40: Why do you convert list to list?

website_portal_pager_limit/controllers/portal.py
```diff
@@ -0,0 +1,42 @@
...
+        options = [int(x.strip()) for x in param.split(",") if x.strip().isdigit()]
+        return options or list(DEFAULT_LIMIT_OPTIONS)
```
geomer198, line 42: The same

website_portal_pager_limit/controllers/portal.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def _prepare_portal_layout_values(self):
+        values = super()._prepare_portal_layout_values()
+        values.update(
```
geomer198, line 52: Please add comment this with describe about new functional.

website_portal_pager_limit/controllers/portal.py
```diff
@@ -0,0 +1,42 @@
...
+OPTIONS_PARAM = "website_portal_pager_limit.options"
+
+
+class CustomerPortal(portal.CustomerPortal):
+    # Defined as a property so that every portal route relying on
+    # ``self._items_per_page`` (sale, account, project, ...) transparently
+    # honors the ``limit`` query parameter without overriding each route.
+    @property
+    def _items_per_page(self):
+        limit = request.httprequest.args.get("limit", "")
```
geomer198, line 24: Docstring is missing.

geomer198 (commented):
Please check proposed changes.

website_portal_pager_limit/controllers/portal.py
halbtonjazz: it was meant as a defensive copy against append, sort, etc. I just didn't notice the constant was already a list, not a tuple. Reworked the constant into a tuple, now list() actually makes sense (a fresh mutable list from an immutable constant)

website_portal_pager_limit/controllers/portal.py
halbtonjazz: done

website_portal_pager_limit/controllers/portal.py
halbtonjazz: done

geomer198 (approved):
LGTM

Bearnard21 (approved):
LGTM

Bearnard21:
@pedrobaeza Hello, could you please take a look? Thank you.

pedrobaeza:
`portal` can work without `website`, and `limit` express a restriction more than a configuration IMO, so I would call it `portal_pager_size`.

halbtonjazz:
> `portal` can work without `website`, and `limit` express a restriction more than a configuration IMO, so I would call it `portal_pager_size`.

@pedrobaeza 
fixed, please check

portal_pager_size/templates/portal_templates.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!-- Copyright (C) 2026 XXP
+     License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl). -->
```
pedrobaeza: Is this file being LGPL an error?

halbtonjazz: Yes, you're right about the license headers. But I removed all copyright lines from the module instead of fixing them manuall, our merge process adds the correct copyright headers automatically when the module is merged into OCA

halbtonjazz:
@pedrobaeza 
Could you please take a look at this PR when you have a chance? 
Thanks

pedrobaeza:
Formally, this should be squashed into one commit. About the module itself, I can't say.

halbtonjazz:
> Formally, this should be squashed into one commit. About the module itself, I can't say.

Done, 
the commits are squashed and rebased, please check

PR #1206, opened 2026-07-22
[18.0][FIX] website_form_require_legal: avoid overriding payment terms widget

pilarvargas-tecnativa:
This widget is registered using the same TermsAndConditionsCheckbox name as the native website_sale widget, so it overrides the original implementation completely.

As a result, the change event of the terms and conditions checkbox in the checkout is no longer handled and the payment button is not enabled until the payment method is selected again.

This widget should use a different registry name so both implementations can coexist.

https://github.com/odoo/odoo/blob/06361cf46237623b8232252da2a3e7763cd28cbc/addons/website_sale/static/src/js/terms_and_conditions_checkbox.js#L6

@Tecnativa TT63742

@carlos-lopez-tecnativa @eduezerouali-tecnativa please review

### DIFF
--- a/website_form_require_legal/static/src/js/terms_and_conditions.esm.js
+++ b/website_form_require_legal/static/src/js/terms_and_conditions.esm.js
@@ -3,26 +3,27 @@
 import {Component} from "@odoo/owl";
 import publicWidget from "@web/legacy/js/public/public_widget";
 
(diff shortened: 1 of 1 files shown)


carlos-lopez-tecnativa (approved):

pedrobaeza (approved):
/ocabot merge patch

sergio-teruel (approved):

PR #1213, opened 2026-08-18
[19.0][ADD] website_blog_schema_org: Inject BlogPosting JSON-LD into blog posts

nachoserra-domatix:
This module injects a schema.org `BlogPosting` JSON-LD block into every
published blog post, so search engines can surface article rich results
(headline, author, dates, publisher and image).

The block is built from the post record (name, author, post date, modification
date, publisher and OpenGraph image) and injected before the main post section.
Only the blog post page is affected.

Tests cover the JSON-LD generation with and without an HTTP request.


website_blog_schema_org/models/blog_post.py
```diff
@@ -0,0 +1,57 @@
...
+class BlogPost(models.Model):
+    _inherit = "blog.post"
+
+    def _get_article_jsonld(self):
+        """Build a schema.org BlogPosting JSON-LD block."""
+        self.ensure_one()
+        website = self.website_id or self.env["website"].get_current_website()
+        base_url = website.get_base_url()
+        slug = self.env["ir.http"]._slug
+        url = f"{base_url}/blog/{slug(self.blog_id)}/{slug(self)}"
```
tarteo, line 20: You could use field `website_url` here to make it simpler.

website_blog_schema_org/models/blog_post.py
```diff
@@ -0,0 +1,57 @@
...
+        elif og_img:
+            image = f"{base_url}/web/image/{og_img}"
+        else:
+            image = None
+        date_published = self.post_date.isoformat() if self.post_date else None
+        date_modified = self.write_date.isoformat() if self.write_date else None
+        data = {
+            "@context": "https://schema.org",
+            "@type": "BlogPosting",
+            "headline": self.name or "",
```
tarteo, line 34: name is required, thus the fallback value is unnecesarry

tarteo (approved):
LGTM only some nitpicking

PR #1216, opened 2026-08-20
[19.0][MIG] portal_invitation_by_website: Migration to 19.0

CarlosRoca13:
cc @Tecnativa TT64114

ping @cristina-hidalgo-tecnativa @pilarvargas-tecnativa


pedrobaeza:
/ocabot migration portal_invitation_by_website

Isn't the new core feature "Share users between websites" (unchecking it) aiding with this showing the possible websites in the invitation wizard, or this is still needed?

CarlosRoca13:
The "Shared Customer Accounts" setting (website.specific_user_account) doesn't cover this. It only acts on the frontend signup: res.users._signup_create_user() stamps website_id = current website when a visitor registers from that website (and only when there's a request).

What core adds to the wizard (https://github.com/odoo/odoo/blob/638f9477ee7a6c301e79e65e33816f700520f535/addons/website/wizard/portal_wizard.py)  is just reading partner_id.website_id to compute if there is any duplicated email, but there is no website selector anywhere in the wizard. So with plain core you have to open each contact form and set its Website field before launching the wizard; once that's set, the signup URL already points to the right domain thanks to get_base_url().

image

And I think that the powerfull feature that this module provides is choosing the website from the wizard itself.

image

pedrobaeza:
Perfect, thanks for the explanation. Then it makes sense to continue having this module.

portal_invitation_by_website/__manifest__.py
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "Portal Invitation by Website",
+    "summary": "Restrict portal users to a specific website from the invitation wizard",
+    "version": "19.0.1.0.0",
+    "author": "Tecnativa, Odoo Community Association (OCA)",
+    "website": "https://github.com/OCA/website",
+    "license": "AGPL-3",
+    "category": "Website",
+    "depends": ["auth_signup", "portal", "website"],
```
pilarvargas-tecnativa:
```suggestion
    "depends": ["portal", "website"],
```

https://github.com/odoo/odoo/blob/57c7c9938725d392a6f2cd6c89a861d2a8385c44/addons/portal/__manifest__.py#L17

CarlosRoca13:
Change done @pilarvargas-tecnativa

pilarvargas-tecnativa (approved):

cristina-hidalgo-tecnativa (approved):
Code reviewed and tested locally and on runbot, LGTM!

PR #1218, opened 2026-09-12
[18.0][ADD] website_mass_mailing_altcha

jans23:
This module adds ALTCHA support to the newsletter subscription snippets.


jans23:
@hbrunn maybe you would be willing to review this?

website_mass_mailing_altcha/static/src/newsletter.esm.js
```diff
@@ -0,0 +1,123 @@
...
+    altcha_insert_widget() {
+        if (this.editableMode || !this.altcha_enabled) {
+            return;
+        }
+        if (!this.el.querySelector("altcha-widget")) {
+            this.$el
+                .find(this.altcha_prepend_to)
+                .append(renderToString("website_altcha.AltchaWidget", {}));
+        }
+    },
```
hbrunn: instead of doing this, an extra commit changing `AltchaBaseFunctionality` such that consumers of this class can either set `altcha_prepend_to` or `altcha_append_to` seems useful

website_mass_mailing_altcha/static/src/newsletter.esm.js
```diff
@@ -0,0 +1,123 @@
...
+
+    async altcha_get_payload() {
```
hbrunn: all of the following is bullshit. the right thing to do here is override just `_onSubscribeClick`, wait for verification, attach to the rpc bus to inject the result (and detach the function afterwards), then call super. ~10loc, not 100

hbrunn (changes requested):

website_mass_mailing_altcha/static/src/newsletter.esm.js
jans23: Changed as requested.

website_mass_mailing_altcha/static/src/newsletter.esm.js
jans23: Changed as requested.

website_mass_mailing_altcha/static/src/newsletter.esm.js
```diff
@@ -0,0 +1,55 @@
...
+        if (this.editableMode || !this.altcha_enabled) {
+            return;
+        }
+        if (!this.el.querySelector("altcha-widget")) {
+            this.$el
+                .find(".js_subscribe_wrap")
+                .append(renderToString("website_altcha.AltchaWidget", {}));
+        }
+    },
+});
```
hbrunn, line 25: this is not what I asked for

hbrunn (changes requested):

website_mass_mailing_altcha/static/src/newsletter.esm.js
jans23: @hbrunn Did I get it now?

